PR #16, opened 2018-05-10
site redesign and new events page

derekfidler:


lexholden (approved):
The random over-indendations in a few CSS files is a bit odd, but is a trivial point. LGTM.

Issue #7, opened 2017-06-19
Logo and donate visuals are broken on mobile


derekfidler:

toChaim:
I looked at this. The text of the top component on the home page is getting cut off at the top and it has no background image. Are these the problems you have?
If so they are set on the style sheet for the component. sheltertech.org/src/routes/Home/components/HomeView.scss
line 15 sets the height to half the screen size. I think that is meant to be minimum height.
min-height: 50vh;
Or you could scale down the text so it fits in only half the screen.
Line 23 removes the background image. Just remove it to bring the image back.

I will put in a pull request with these changes. If I have the problem wrong or you want a different solution reply here or put it as a comment on the pull request. 
Thanks
Chaim

PR #17, opened 2018-06-06
Update HomeView.scss to work better on mobile

toChaim:
Logo and donate visuals are broken on mobile #7


derekfidler:
i think this was an old issue from before the site redesign. the issue has been fixed but the not closed. will review the other issues to see if they are still relevant.

PR #30, opened 2020-04-27
add home page

giftofgrub:
Just added the homepage. Next for me to add is the header and footer.


README.md
```diff
@@ -14,6 +14,15 @@ npm start
...
 Then you can access it at [http://localhost:8000](http://localhost:8000).
 
+## Build this project as a static site
+
+Run the following commands:
+
+``` bash
+npm run build
+```
+Then you can access project at public/
```
richardxia:
```suggestion
Then you can access the static files in the `public/` directory.
```

gatsby-config.js
```diff
@@ -0,0 +1,43 @@
+module.exports = {
+  siteMetadata: {
+    title: `ShelterTech - Technology for the underserved`,
+    description: `Blog example for Gatsby & Prismic`,
```
richardxia: We should replace this with the real description for the site. You can find the current description of the site in the `<meta>` tags: https://github.com/ShelterTechSF/sheltertech.org/blob/0bbe6a3cfc17d9aafbeb4308d9dfccbe56c0c993/src/index.html#L12

gatsby-config.js
```diff
@@ -0,0 +1,43 @@
...
+    author: `Sheltertech`,
```
richardxia:
```suggestion
    author: `ShelterTech`,
```

The "T" in Tech should be capitalized.

gatsby-config.js
```diff
@@ -0,0 +1,43 @@
...
+    {
+      resolve: `gatsby-plugin-manifest`,
+      options: {
+        name: `gatsby-starter-default`,
+        short_name: `starter`,
+        start_url: `/`,
+        background_color: `#663399`,
+        theme_color: `#663399`,
+        display: `minimal-ui`,
+        // icon: `src/images/favicon.ico`, // This path is relative to the root of the site.
```
richardxia: Either we should fill this out with real information, or we should not use this plugin. It wouldn't be good for the word "gatsby-starter-default" or "starter" to show up in the actual app.

package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "sheltertech.org",
+  "version": "1.0.0",
+  "description": "Sheltertech Info Site",
```
richardxia:
```suggestion
  "description": "ShelterTech Info Site",
```

package.json
```diff
@@ -0,0 +1,35 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "gatsby develop",
+    "build": "gatsby build"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/ShelterTechSF/sheltertech.org.git"
+  },
+  "author": "Sheltertech.org",
+  "license": "MIT",
```
richardxia: We actually use the GPLv3 license for all ShelterTech open source projects. See https://github.com/ShelterTechSF/askdarcel-web/blob/aab487c081e87ba6804fefea05f3ed4f0908c71e/package.json#L101 for what the line looks like in the package.json from the askdarcel-web project.

src/pages/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import Home from './pageComponents/Home'
+
+export default () => {
+   return(
```
richardxia: Nit: Most JavaScript style guides recommend putting a space between the `return` keyword and an opening parenthesis.

Also, as a more general comment, we should probably make sure that our linters and code styles are consistent between this project and askdarcel-web, so that people who work on both of them don't have to remember different sets of style rules.

src/pages/index.js
```diff
@@ -0,0 +1,10 @@
...
+    // <Layout>
```
richardxia: Let's not leave in any commented out code and instead just delete them. Some of the issues with committing commented out code include:

- It's difficult for others to understand why the code was commented out in the first place, and when it would be appropriate to uncomment them again
- It adds noise to reading the file, since it's code that isn't relevant to the current functionality of the code. This will especially add up over time.
- It's easy for commented out code to become out of date, since they aren't tested or linted, and it could become so outdated that the code doesn't even work anymore

src/pages/index.js
```diff
@@ -0,0 +1,10 @@
...
+      <Home />
+    // </Layout>
+  )
+}
```
richardxia: Can you configure your text editor to leave trailing newlines at the end of the files? GitHub here is showing that you don't have a final newline with that red icon there.

There are some tools that will not function properly without the newline character at the end, so it's better to always leave it in. See https://github.com/airbnb/javascript#whitespace--newline-at-end. We can enforce this by either adopting the airbnb lint ruleset (which we do in the askdarcel-web app) or at least enabling the `eol-last` ESLint rule.

src/pages/pageComponents/Home.js
```diff
@@ -0,0 +1,154 @@
+import React from 'react'
+import '../../stylesheets/Home.scss'
+
+//import '../../stylesheets/components/Home/assets'
+// import images
```
richardxia: Let's remove the commented out imports.

richardxia (changes requested):
I couldn't leave comments on some of the files, since they are images, but I noticed that you have a lot of images that have the word "copy" in their names. Do these need to be committed to the repository? Here's an example of one:

```
src/images/askdarcel-logo copy.png
```

Also, it looks like you are failing the Travis CI checks. We should fix them so that they are updated for this new code. The Travis CI configuration lives in [.travis.yml](https://github.com/ShelterTechSF/sheltertech.org/blob/0bbe6a3cfc17d9aafbeb4308d9dfccbe56c0c993/.travis.yml), and here are a few of the problem that I see with them:

- They specify Node.JS v6, which is very old at this point, and I'm not sure if the libraries that we're using will be compatible.
- They are trying to use `yarn` instead of `npm`, but you have a package-lock.json file instead of a yarn.lock file
- None of the `npm run` targets specified there exist in the new code.

I've been interested in switching away from [Travis CI](https://travis-ci.org/) to [GitHub Actions](https://github.com/features/actions), so this might be a good opportunity to do that. We don't need to exactly replace the old tests that were running in the Travis CI configuration file, but I think it would be a good idea to at least check that 1) that the site builds with no issues and 2) that all lint checks pass.

src/pages/index.js
giftofgrub: Thanks! I really had to search for what "nit:" is. :)

src/pages/index.js
giftofgrub: Thanks. I added config to adopt the airbnb style

giftofgrub:
Took out the images that have "copy"

.eslintrc
```diff
@@ -1,27 +1,11 @@
...
   },
   "rules": {
-    "key-spacing"          : 0,
-    "jsx-quotes"           : [2, "prefer-single"],
-    "max-len"              : [2, 120, 2],
-    "object-curly-spacing" : [2, "always"]
+    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
   }
-}
+ }
```
richardxia:
```suggestion
}
```

.travis.yml
```diff
@@ -1,19 +1,16 @@
...
+  - "14"
 
 cache:
   directories:
     - node_modules
 
 install:
-  - npm install -g yarn
-  - yarn install
+  - npm install gatsby@2.20.36
```
richardxia:
```suggestion
  - npm install
```

You actually don't need to set the gatsby version like this. If you just run `npm install`, it'll install all of the packages and versions specified in `package.json`.

.travis.yml
```diff
@@ -1,19 +1,16 @@
...
 
 script:
-  - npm run deploy:dev
-  - npm run deploy:prod
+  - npm run build
+
```
richardxia: Let's add another item here for running the linters, since that will make sure that CI is always checking whether code passes the lint check. You'll need to add another item to the `scripts` section of `package.json` to actually run ESLint. If you want to match what we named them in askdarcel-web, they're `lint` and `lint:fix`: https://github.com/ShelterTechSF/askdarcel-web/blob/17bcd0b7c65f7b8cfbe8522e42a6966ed420d3bc/package.json#L88-L89

richardxia (commented):
Almost there, just a couple more pieces of feedback, but I think this is looking pretty good!

.travis.yml
giftofgrub: I added the lint scripts. I also added `public/` to the .eslintignore.

richardxia (approved):
Awesome, this is great! I'll go ahead and merge this in. I'm going to squash merge this so that it gets rid of the large, duplicated image files you had earlier.

PR #31, opened 2020-05-12
Nav layout

giftofgrub:


src/components/Layout/Footer/Footer.js
```diff
@@ -0,0 +1,30 @@
...
+        <li><a href="mailto:info@sheltertech.org">Contact Us</a></li>
+        <li><Link className="donate-button" to="/donate">Donate</Link></li>
+      </ul>
+    </nav>
+    <div className="site--footer--legal">
+      <ul>
+        <li>ShelterTech is a 501 (c)(3)</li>
+        <li>ein: 38-3984099</li>
+        <li>made in san francisco</li>
+        <li>© 2018 All Rights Reserved</li>
```
richardxia:
```suggestion
        <li>© 2020 All Rights Reserved</li>
```

While we're redoing the site, we might as well update the copyright year here.

src/components/Layout/SlideMenu/SlideMenu.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Link } from 'gatsby';
+import { slide as Menu } from 'react-burger-menu';
+import '../../../stylesheets/components/SlideMenu.scss';
+
+class SlideMenu extends React.PureComponent {
```
richardxia: Is there a reason you made this a PureComponent instead of defining it as a function-based component, which are pure by default? If not, I think I'd prefer that we make this consistent with all of our other pure components, which are written as function-based components.

src/components/Layout/index.js
```diff
@@ -0,0 +1,63 @@
...
+  />
+);
+
+
+const Layout = (props) => {
+  // Define the meta title and description
+  const { data } = props;
+  const { title } = data.site.siteMetadata;
+  const { description } = data.site.siteMetadata;
+  const { children } = props;
```
richardxia:
```suggestion
const Layout = ({ data, children }) => {
  // Define the meta title and description
  const { title, description } = data.site.siteMetadata;
```

You can destructure multiple properties at the same time. You can also destructure the arguments directly in the function parameters list.

src/components/Layout/SlideMenu/SlideMenu.js
```diff
@@ -0,0 +1,23 @@
...
+  render() {
+    return (
+      <Menu right>
+        <Link to="/">Programs</Link>
+        <li><Link to="/annual-report-2018">Impact</Link></li>
```
richardxia: Is there a reason that this is the only menu item that is wrapped in a `<li>` tag in this menu? Also, is there a reason that this hamburger menu mostly does not use `<li>` tags even though the non-hamburger header and the footer use `<li>` tags?

src/components/Layout/Footer/index.js
```diff
@@ -0,0 +1,3 @@
+import Footer from './Footer';
+
+export default Footer;
```
richardxia: By the way, this could be more concise as:

```suggestion
export { default } from './Footer';
```

You can see us using this pattern in a number of places in askdarcel-web, such as https://github.com/ShelterTechSF/askdarcel-web/blob/016ba8253cc6d4210f5022dbb774a06e024a6c0e/app/pages/HomePage/components/SearchBar/index.js#L1.

richardxia (commented):
This is looking pretty good!

I do have a question about the organization of files. Given that the three navigation menus are so similar (today, they have the exact same set of links), perhaps it would make sense to define them in a single file, and rather than using the `default` export, we export the three names individually? My thinking is that if someone wants to edit one of the navigation menus in the future, it'll be easier to remember to update the others if they're all in the same file. I think historically we've forgotten to update mobile menus when making changes for desktop screens, since it's often easy to forget to test mobile screen sizes.

giftofgrub:
> This is looking pretty good!
> 
> I do have a question about the organization of files. Given that the three navigation menus are so similar (today, they have the exact same set of links), perhaps it would make sense to define them in a single file, and rather than using the `default` export, we export the three names individually? My thinking is that if someone wants to edit one of the navigation menus in the future, it'll be easier to remember to update the others if they're all in the same file. I think historically we've forgotten to update mobile menus when making changes for desktop screens, since it's often easy to forget to test mobile screen sizes.

I have no preference. Are you suggesting that we consolidate Header, Footer, and the SlideMenu into one Layout component much like the CoreLayout component in the previous site? Or is it that we just use one file for Header, for example, and do away with the Header folder, and the index.js within the folder?

src/components/Layout/SlideMenu/SlideMenu.js
giftofgrub: no reason... this was just my mistake

src/components/Layout/index.js
giftofgrub: Thanks!

richardxia:
> I have no preference. Are you suggesting that we consolidate Header, Footer, and the SlideMenu into one Layout component much like the CoreLayout component in the previous site? Or is it that we just use one file for Header, for example, and do away with the Header folder, and the index.js within the folder?

I was thinking of the latter, keeping all three components in a single file but still as separate components, since that makes it easier to see all the navigation menus at the same time.

On that note, I just noticed that the Header links to `/annual-report`, while the Footer and the SlideMenu both link to `/annual-report-2018`. It's little differences like that which are harder to review when the menus are in different files.

Aside from that, I think this looks pretty good to me, besides the lint failures. Good job on this!

richardxia:
I think I may have been a bit unclear about what I meant regarding having a single file. What I meant is that we could have created a single .js file that includes the definitions of all three components, not just importing the component names into a single file.

In your current code, you still have three separate files for Footer, Header, and SlideMenu, and Header is a little bit odd because it contains both an import of SlideMenu as well as a definition of the main Header itself.

What do you think about creating a single file named Navigation.js which includes definitions of all three components, e.g.?

```jsx
// Navigation.js

const Header = () => { ... }

const Footer = () => { ... }

const SlideMenu = () => { ... }
```

This way, when you have the file open, you can see all three of the menu items at the same time, which will prevent mistakes from editing just one file at a time. I don't think this would make the file too large, since they're only 10-20 lines long, and I think it would make it easier to read to have all three component definitions on screen at the same time.

giftofgrub:
> I think I may have been a bit unclear about what I meant regarding having a single file. What I meant is that we could have created a single .js file that includes the definitions of all three components, not just importing the component names into a single file.
> 
> In your current code, you still have three separate files for Footer, Header, and SlideMenu, and Header is a little bit odd because it contains both an import of SlideMenu as well as a definition of the main Header itself.
> 
> What do you think about creating a single file named Navigation.js which includes definitions of all three components, e.g.?
> 
> ```js-jsx
> // Navigation.js
> 
> const Header = () => { ... }
> 
> const Footer = () => { ... }
> 
> const SlideMenu = () => { ... }
> ```
> 
> This way, when you have the file open, you can see all three of the menu items at the same time, which will prevent mistakes from editing just one file at a time. I don't think this would make the file too large, since they're only 10-20 lines long, and I think it would make it easier to read to have all three component definitions on screen at the same time.

Yes, I agree with this one. I think it would make it easier to work on in the future if they are in one file.

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import { Link } from 'gatsby';
+import { Helmet } from 'react-helmet';
+import PropTypes from 'prop-types';
+import { slide as Menu } from 'react-burger-menu';
+import favicon from '../../images/favicon.ico';
+import '../../stylesheets/components/Navigation.scss';
+
+const Navigation = ({ data, children }) => {
```
richardxia: I'm not sure if this makes sense as a single component, at least not with the name "Navigation". Normally, if I saw a component by the name of Navigation, I would think that it would only contain an independent navigation. However, what you really have here is something that defines the entire application layout.

For example, where you actually use this component, you have:

```jsx
  <Navigation>
    <Home />
  </Navigation>
```

But this looks kind of weird, because from reading it, it looks like you're putting the Home page _inside_ of the navigation. But that's not what's actually happening; what you really have is the Home page rendered inside of the main application layout. The navigation is just a single section of the page that links to other pages; so really, what you have named `Navigation` is not a navigation, but a page layout that includes three separate navigation components within it.

What I would recommend is moving this `Navigation` component into your `src/pages/index.js` file, and just having it _be_ that page's component, without assigning it a name.

I would leave the other three navigation components in this file, but instead of having a default export, export the three components so that you can use them in the `src/pages/index.js` page.

If in the future we want to define multiple page layouts, then we can create a collection of layout components that may each independently reuse some or all of the navigation components.

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+
+const Header = () => (
+  <header className="site--header">
+    <div className="site--header--wrapper">
+      <h1 className="site--header--logo"><Link to="/">ShelterTech</Link></h1>
+      <nav className="site--header--nav">
+        <SlideMenu />
+        <ul>
```
richardxia: It doesn't look like you put any `<li>` tags within the `<Links>` component, so it doesn't make sense to put a `<ul>` tag around it. `<ul>` elements are only really supposed list items within them, but if you don't actually create any, you shouldn't use it.

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,101 @@
...
+const SlideMenu = () => (
+  <Menu right>
+    <Links withBlog />
+  </Menu>
+);
+
+const Footer = () => (
+  <footer className="site--footer">
+    <nav className="site--footer--nav">
+      <ul>
```
richardxia: Same comment about the `<ul>` tag here.

src/components/Navigation/Navigation.js
```diff
@@ -0,0 +1,101 @@
...
+        <li>ein: 38-3984099</li>
+        <li>made in san francisco</li>
+        <li>© 2020 All Rights Reserved</li>
+        <li><a href="https://www.facebook.com/ShelterTechOrg/">facebook</a></li>
+      </ul>
+    </div>
+  </footer>
+);
+
+// eslint-disable-next-line react/prop-types
```
richardxia: Is there a reason you disabled the lint check for the prop types here?

giftofgrub: I opted not to include it in the proptypes since its just a true or false value passed as props to Links

src/components/Navigation/Navigation.js
giftofgrub: Wont this clutter my src/pages/index.js? What if I rename to Navigation to Layout? Would that make more sense?

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "prettier.singleQuote": true,
+  "prettier.jsxSingleQuote": true
+}
```
richardxia: I don't think it's a good idea to add this file to the codebase. This only configures prettier for VSCode, but is not usable by anyone who doesn't use VSCode, or a CI check.

How about we separate the task of adding Prettier to the checks into a separate PR?

richardxia (approved):
LGTM, modulo one minor comment!

PR #37, opened 2020-08-24
Fix stylelint setup

richardxia:
I messed up when I set up stylelint, the CSS linter we are using, so it looks like it wasn't running at all. This fixes up the `npm run lint` script so that it properly picks up all the .scss files, and it corrects the lint violations in the existing files.


davidagustin (approved):

davidagustin:
I ran the linter with the single quotes added in the JSON file. The linter registered. Verified and works.

richardxia:
Thanks for the review!

PR #42, opened 2020-09-06
Port over events page.

richardxia:
This ports over the events page to the new Gatsy-based site.


davidagustin (approved):
Looks good to me, build works.

PR #48, opened 2020-09-13
Get involved

davidagustin:


candywang:
Looks great!

src/pages/get-involved/index.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import EventsView from "../../components/pages/get-involved/GetInvolvedView";
```
richardxia:
```suggestion
import GetInvolvedView from "../../components/pages/get-involved/GetInvolvedView";
```

Could you rename this to `GetInvolvedView` so that it won't be confused with the actual `EventsView` component defined in the `EventsView.js` file? Because this is a default import in ES6-style module imports, the name here doesn't actually have to match the name of the actual component.

You'll also need to update the name that is used below.

candywang (approved):

richardxia (approved):
LGTM!

candywang (approved):

candywang (approved):

PR #49, opened 2020-09-13
Fix CSS for event cards.

richardxia:
This restores the original colors and hover style, which were getting altered by the global `<a>` tag styles. @davidagustin and I noticed that without these fixes, the text in the event cards were getting a blue color and they had an odd on-hover style applied, which were coming from the global `<a>` tag.

This was probably originally caused by me changing the `onClick` event handlers on the cards with an `<a>` tag wrapping the whole element, since the former was causing lint errors related to accessibility (a11y).


davidagustin (approved):
Verified, looks good to me

PR #50, opened 2020-09-14
port-impact-page-to-gatsby-1578

davidagustin:


richardxia (changes requested):
Thanks for working on this! I can confirm that the PDF download appears to be working correctly. However, I noticed that part of the page's styling is a bit off. Here's a screenshot of the part of the page that looks like it has incorrect styling compared to the original:

Screen Shot 2020-09-13 at 9 57 53 PM

Since it's late, I can see if I can make the necessary changes on this branch to get it looking right.

src/components/pages/annual-report-2018/AnnualReportView.scss
```diff
@@ -0,0 +1,451 @@
...
+    position: relative;
+    width: 45%;
+    @media screen and (max-width: 600px) {
+      height: 100px;
+      padding: 10px 0 0 20px;
+    }
+  }
+}
+
+.annual-report-goals li {
```
richardxia:
```suggestion
.annual-report--goals li {
```

Found it. There was a typo in this selector. It should be `--goals`, not `-goals`.

src/components/pages/annual-report-2018/AnnualReportView.scss
```diff
@@ -0,0 +1,451 @@
...
+    font-size: 1.75rem;
+    font-weight: 700;
+    @media screen and (max-width: 600px) {
+      font-size: 1.2rem;
+      line-height: 1.2;
+    }
+  }
+
+  .goal-number {
+    color: #222;
```
richardxia:
```suggestion
    color: #222;
```

Also, this color is wrong. It was the ShelterTech blue on the old site, but here you've turned it into a dark gray.

Is there a reason you replaced all of the variables holder the colors with hardcoded values? I think this strictly makes it worse, since it becomes more difficult to ensure that we are using the consistent colors for the branding everywhere. I'm going to restore these back to what they used to be.

richardxia (changes requested):

src/components/pages/annual-report-2018/AnnualReportView.scss
```diff
@@ -0,0 +1,451 @@
...
+    }
+  }
+}
+
+.annual-report-goals li {
+  &:nth-last-of-type(-n+2) {
+    border-bottom: 1px solid #222;
+  }
+
+  p {
```
richardxia: This didn't get translated properly. It used to be `& > p`, and you've dropped the [direct child selector (`>`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator), which makes a big difference. Without the direct child selector, this will match _any_ `<p>` tag that is nested below the parent selector rather than direct children of the parent selector.

src/components/pages/annual-report-2018/AnnualReportView.js
```diff
@@ -0,0 +1,331 @@
...
+          <div className="annual-report--goals">
+            <h2>ShelterTech exceeded goals across every program</h2>
+            <ul>
+              <li>
+                <span className="goal-number">7</span>
+                <p>ShelterConnect Wifi Installations</p>
+              </li>
+              <li>
+                <span className="goal-number">7</span>
+                <p>AskDarcel Releases</p>
```
richardxia, line 51: @derekfidler, do you have changes that haven't been pushed up to the master branch? I noticed that on the actual live site today, this has been renamed to "SF Service Guide Releases", but the [code on master still shows "AskDarcel Releases"](https://github.com/ShelterTechSF/sheltertech.org/blob/07a64b6ed754cbfb005985dbed3e52987f62b327/src/routes/AnnualReport2018/components/AnnualReportView.js#L32).

Could you please push up any changes that you have locally. It'd be bad if we missed something as we wrap up the code porting changes due to not having the latest changes up on the master branch.

src/components/pages/annual-report-2018/AnnualReportView.js
```diff
@@ -0,0 +1,331 @@
...
+            </ul>
+          </div>
+        </div>
+      </div>
+
+      <div className="annual-report--askdarcel">
+        <div className="wrapper">
+          <img src={ImgAskDarcelLogo} alt="AskDarcel" />
+          <p className="annual-report--text">
+            AskDarcel is a robust and continously updated online directory for
```
richardxia, line 128: Here's also a spot where both this PR and master show "AskDarcel" but the actual live site shows "SF Service Guide".

src/components/pages/annual-report-2018/AnnualReportView.js
```diff
@@ -0,0 +1,331 @@
...
+          <p className="annual-report--text">
+            In 2018, we released an Alpha of the product but decided to hold off
+            on a public launch while we validated the content on the site—trust
+            is the biggest concern for people experiencing homelessness when it
+            comes to referrals for services. The good news is we’ve now
+            validated everything on the site and are ready for our Beta launch.
+          </p>
+          <h3>
+            In 2018, over <span className="white">2,000</span> hours were spent
+            on the development of AskDarcel.
```
richardxia, line 142: Here's also a spot where both this PR and master show "AskDarcel" but the actual live site shows "SF Service Guide".

src/components/pages/annual-report-2018/AnnualReportView.js
```diff
@@ -0,0 +1,331 @@
...
+            needed to keep growing and surpassing our goals. Here’s what we want
+            to accomplish in 2019.
+          </p>
+          <ul className="annual-report--2019-goals">
+            <li>
+              <p>Public launch of AskDarcel</p>
+            </li>
+            <li>
+              <p>2000 AskDarcel users by year end</p>
+            </li>
```
richardxia, line 298: Last two spots where the live site doesn't match the master branch.

src/components/pages/annual-report-2018/AnnualReportView.scss
richardxia: Fixed this in https://github.com/ShelterTechSF/sheltertech.org/pull/50/commits/f45fc1a29497b9476fff6f505fdcdf7ac64c5367.

src/components/pages/annual-report-2018/AnnualReportView.scss
richardxia: Changed all the colors back to variables in https://github.com/ShelterTechSF/sheltertech.org/pull/50/commits/7741bb6e04cd4d3fa47ff52b3beb0ba85eb8ebd8.

src/components/pages/annual-report-2018/AnnualReportView.scss
richardxia: Fixed all the selectors in https://github.com/ShelterTechSF/sheltertech.org/pull/50/commits/2fae51536617359ec8f09ff3868469220ecfdd5b.

richardxia (approved):
This LGTM now, so I think I'll go ahead and merge this in, but I'd like to do something about the code in the master branch not matching what's actually deployed to production. We'll probably need to do a followup PR for the annual report page on this new-master branch once actual master is updated as well.

derekfidler:
Sorry about this. I had a commit I hadn't pushed up. It's now on Master.

richardxia:
Thanks, @derekfidler! I can go try to sync up those changes with our `new-master` branch. I think we're getting pretty close to being able to replace the master branch with our new Gatsby-based codebase.

PR #52, opened 2020-09-15
Port latest changes from master

richardxia:
I went through some of the latest changes to the `master` branch, including the one that @derekfidler just pushed up in https://github.com/ShelterTechSF/sheltertech.org/pull/50#issuecomment-691914763. There were a few other commits that we were missing from this `new-master` branch, so I've gone through and reapplied them here.

The specific commits that I reapplied here were:

- 57105e6efedfff049f2b0ca8f0e479d4bce3558b - Adding Google Analytics tracking
- 56ed23e0b3cac265d927fef7c5fbcc2ce84fff27 - Updating our mailing address
- 9f524a0afe71df853f64a0de4c232e70d84a1773 - Rewording AskDarcel -> SF Service Guide, as well as a few other changes to the Team page

I did notice that we are completely missing the Giving Tuesday page on the `new-master` branch, so I will go port that over in a separate PR, as well as doing a quick pass over all pages that exist on the old site to make sure that we haven't missed anything else.


richardxia:
I pushed up one more commit to this branch, since I noticed some issues with the CSS for the footer navigation compared to the production site.

davidagustin (approved):

PR #54, opened 2020-09-21
Add Storybook

richardxia:
This adds [Storybook](https://storybook.js.org/) to our development environment, and it also adds a basic Button component.

This may not be ready yet to merge, since I have a few more things that I think I want to tweak, but it's most of the way there.

This also ends up setting some base conventions for how code should be structured for this application. Here are some of the choices that I made, which I am happy discuss more about before we bless this as "official":

- Plain, vanilla CSS with CSS modules, instead of a preprocessor like Sass. CSS has advanced a lot since when Sass was originally created, and I think many of the things that we used to rely on from Sass either have replacements in vanilla CSS, JavaScript libraries like React, or PostCSS. See https://cathydutton.co.uk/posts/why-i-stopped-using-sass/ for a blog post on the topic
- [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) in place of Sass variables
- Grouping components into three categories:
  1. inline components, such as buttons, labels, etc.
  2. block components, which act as block elements and which fill the space provided by its parent
  3. grid-aware components, which act as the top level of layout to arrange things based on the screen size and the grid system

  I'd suggest that we organize all files related a component in the following way:
  ```
  src/components/inline/<Name>/
    <Name>.js  # The React component itself
    <Name>.module.css  # The CSS module that styles it
    <Name>.stories.js  # The Storybook stories file
  ```


richardxia:
Also, here's a screenshot of the Storybook so far:

Screen Shot 2020-09-20 at 8 46 59 PM

richardxia:
@davidagustin, I think I have this set up in a way that I'm mostly happy with. Please check it out and let me know if you run into any issues. Also please let me know if there's anything else that I can add to the README.

davidagustin (approved):

PR #55, opened 2020-09-21
Fix minor styling issues on various pages.

richardxia:
I went through all the pages on the site one last time to catch any remaining issues, and I found a few. They're all relatively minor, but I'd like to get these in before swapping out the old site code.


richardxia:
Going to merge this in without a review, since I want to get the new `main` branch created with all of this stuff in it.

PR #61, opened 2020-10-05
1629 create ProjectCard component

candywang:
Implement ProjectCard component with the following changes:

- change component, variable, and file names to `ProjectCard`
- change class selectors to camelCase
- edit stylings: remove background, remove containers, change margins from bottom to top, adjust card width
- make all props required
- add alt prop
- change key name for photo to image
- pass down props from Storybook
- decreased image size and changed png image to a screenshot of the design sample image
- update css to use global css variables
- destructure args in storybook file


src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./Card.module.css";
+
+const Card = ({ photo, title, description }) => (
+  <div className={s["card-container"]}>
```
richardxia: Let's actually use camelCase instead of kebab-case for CSS classes. Since we're using CSS modules, it's easier to use camelCase for CSS class names because then we can use dot notation instead of bracket notation in JavaScript.

I hit an issue yesterday with the CSS lint rules requiring kebab case for class names, so I had to make a change in a local branch to change the linter settings. Let me create a PR for just that change so that this PR can pass lint when you change the class name casing.

src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./Card.module.css";
+
+const Card = ({ photo, title, description }) => (
```
richardxia: Regarding the name of this component and this file itself, could we rename this to `ProgramCard`? I only found out recently that @derekfidler had created a Figma design for this site's design system, and he called this particular card a Program Card. Here's the Figma design for this particular component: https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=665%3A2

Also, on the topic of the file name: sorry for not getting it out sooner, but yesterday I finally finished writing up [my proposal](https://github.com/ShelterTechSF/sheltertech.org/pull/60) for classifying components within one of three categories: inline, block, and grid-aware. I'm open to feedback on the system, including the names, but I was hoping that this system would influence the file directory structure of the repo.

Under the taxonomy in my proposal, I would call this component a "block" component, since I think the default behavior should be for it to fill the space allocated by its parent. This allows the component to be more flexible and reusable in more scenarios. Therefore, I'd suggest that we put this component in a new `src/components/block` directory, which is where I would suggest we place other block components.

A lot of my feedback on the CSS will probably also reflect this.

richardxia: One last thing: I'd suggest we rename `photo` to `image`. "Photo" implies that it's a picture of something captured with a camera, whereas an "image" can include photographs as well as drawings. I know I'm being a little bit pedantic here, but I personally find it valuable be precise with terminology in order to avoid confusion.

src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+  <div className={s["card-container"]}>
+    <div className={s["img-container"]}>
+      <img className={s.img} src={photo} alt="" />
```
richardxia: I think it's important to not set a dummy value for the `alt` property because this hurts users with visual impairments. Since the alternative text is going to be specific to the instance of the card, I think this should probably be a required prop of this component.

src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+  photo: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.string,
+};
+
+Card.defaultProps = {
+  photo: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.string,
+};
```
richardxia: This isn't a correct usage of `defaultProps`. [`defaultProps`](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) is meant to hold the default value that a prop should receive if a value is not passed in by the caller. The values that you have listed here are not things that we want the component to receive when a value is not passed in, since these objects are prop-types-specific values to help the prop-types library enforce type checks.

I would actually suggest that we make all of these props required, which means that you do not and should not provide default values. Making props required allows us to enforce requirements and consistency for the usage of a component. It is always easier to relax a required prop into an optional one, since any user of the component must have provided a value. In contrast, turning an optional prop into a required prop is more difficult, since there may be usages of the component that did not previously pass in a value for the prop but now need to.

src/components/inline/Card/Card.stories.js
```diff
@@ -0,0 +1,28 @@
...
+  title: "Inline/Card",
+  component: Card,
+};
+
+const Template = () => (
+  <div style={{ backgroundColor: "black", width: "370px" }}>
+    <Card
+      photo={img}
+      title="ShelterConnect"
+      description="Every year, our ShelterConnect program provides 3XXX homeless and housing insecure people with free and unlimited internet access and we plan to provide every homeless individual with the digital infrastructure that meets their needs by 2024."
```
richardxia: You'll actually want these to be props of the template. That way, you can actually dynamically control the value of the prop in the Storybook interface.

src/components/inline/Card/Card.stories.js
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import Card from "./Card";
+import img from "./sample.jpg";
```
richardxia: Could you pick a much smaller image to use in the story? Adding a 1.21 MiB image is actually quite large for something that is just a test value for a Storybook story.

src/components/inline/Card/Card.stories.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default {
+  title: "Inline/Card",
+  component: Card,
+};
+
+const Template = () => (
+  <div style={{ backgroundColor: "black", width: "370px" }}>
```
richardxia: We may want both a dark and light variant of the card. Controlling the background color of the containing element like this is fine, but we'll want variants of the Card itself that have light and dark text so that they look fine on dark and light backgrounds, respectively.

src/stylesheets/global.css
```diff
@@ -1,13 +1,15 @@
...
+@import url('https://fonts.googleapis.com/css2?family=Merriweather:wght@300;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,300;0,400;0,700;1,300&display=swap');
 
 :root {
   /* Colors */
   --color-brand: #0025e4;
   --color-white: #fff;
 
   /* Fonts */
-  --font-family-base: 'Lato', 'Arial', sans-serif;
```
richardxia: I think it's still probably a good idea to have a [fallback font](https://css-tricks.com/css-basics-fallback-font-stacks-robust-web-typography/) that is likely to already exist on someone's computer in case if they fail to download the non-standard font that we request.

For Merriweather, I'd suggest Georgia as a fallback font.

src/components/inline/Card/Card.module.css
```diff
@@ -0,0 +1,39 @@
+.card-container {
+  height: auto;
```
richardxia:
```suggestion
```

There's no need to set `height: auto` since that's the default for `<div>`s and almost all block elements.

src/components/inline/Card/Card.module.css
```diff
@@ -0,0 +1,39 @@
...
+  width: 370px;
```
richardxia:
```suggestion
```

Let's not set an explicit width here, since that prevents this element from flexibly resizing according to the occasion. As a concrete example, if you changed the Storybook container div to have a width of 300px, this card is going to look pretty weird.

src/components/inline/Card/Card.module.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.img-container {
+  margin-bottom: 20px;
+}
+
+.img {
+  height: 250px;
+  width: 100%;
```
richardxia: You're going to want to add `object-fit: cover` in order to maintain the aspect ratio of the image. Otherwise, setting the height and the width like this will force the image to stretch or compress, which may distort the image. The `object-fit` property lets you control how an image should be cropped or scaled down, and in our case, I think `cover` is the behavior that we probably want.

See https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit for more information.

src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./Card.module.css";
+
+const Card = ({ photo, title, description }) => (
+  <div className={s["card-container"]}>
+    <div className={s["img-container"]}>
+      <img className={s.img} src={photo} alt="" />
+    </div>
+    <div>
```
richardxia: This `<div>` that wraps the `title-container` and the `description-container` doesn't seem to have a purpose, and the whole card component will render exactly the same even if you didn't have it. Could you remove it?

src/components/inline/Card/Card.module.css
```diff
@@ -0,0 +1,39 @@
...
+  margin-bottom: 10px;
+}
+
+.title {
+  color: var(--color-white);
+  font-family: var(--font-family-title);
+  font-size: 30px;
+  font-style: normal;
+  font-weight: 300;
+  line-height: 130%;
```
richardxia: We should probably capture these as variables so that we can ensure consistent styles across the site. We actually have an entire [Figma page](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=109%3A0) dedicated to typography, and it sets the family, size, weight, and line-height for each typography style.

I haven't tried this kind of thing before, but what if we create a single variable for each typography style that contains everything that goes on the right-hand side of the [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property? For example, `--font-hero` could be set to something like `normal 300 67px/1.3 var(--font-family-title)`, and then you could use it in a component as `font: var(--font-hero)`.

Also, on a separate note: setting a line height of 130% is not the same thing as a line height of 1.3. Although for the current element, it will result in the same line height, the way the value is inherited by child elements is different. You will almost always want to use the decimal version for line height. See https://stackoverflow.com/a/32995898 for a more detailed explanation of why a line height written in percentages will probably not behave the way you want it to.

src/components/inline/Card/Card.module.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.description-container {
+  margin-bottom: 20px;
```
richardxia: I think we should take out this bottom margin. In case if you're not familiar with collapsing margins, [here's an article about them](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing). Without going into the details, what this means for us is that this margin effectively gets applied to the whole card component, not just this `description-container`, which means that this margin exists _below_ the entire card component.

Here's a screenshot of what this looks like with the Chrome inspector, where the orange region is the margin:

Screen Shot 2020-10-05 at 10 09 40 PM


I would advise not doing this because this could make it more difficult to position the card in scenarios where we do not want that bottom margin.

src/components/inline/Card/Card.js
```diff
@@ -0,0 +1,33 @@
...
+      <div className={s["title-container"]}>
```
richardxia: Regarding these `title-container` and `description-container` elements: although I think it's a good thing in general to separate the concerns of styling from positioning, I think this is a case where it's not necessary because the components are simple enough.

Also note that in this case, because you are using `<div>`s for both the container and the inner element, they are both block elements and the nesting of block elements doesn't offer us much benefit. A container element would still be beneficial in the case where you have non-block elements, such as a button (which would normally be inline) or an img (which is basically an inline-block element).

richardxia (changes requested):
Thanks for submitting this PR! I know I left a lot of comments, but a lot of it is due to us not having many examples of the conventions that I'd like to set for this project. Overall, I think this is a great first start, and it does a good job of producing the design, so most of my comments are directed at things that may not have a direct visual impact right now but that I anticipate will help make this component more flexible and reusable in the future.

One other request I'd make of you is to fill out the PR description with a brief description of what you've done. Here's a good article with general advice on writing good pull requests, and I think it covers a lot of what I would like to see from pull requests in this repository: https://medium.com/@hugooodias/the-anatomy-of-a-perfect-pull-request-567382bb6067

src/components/inline/Button/Button.module.css
```diff
@@ -3,7 +3,7 @@
   border: 1px solid var(--color-brand);
   color: var(--color-white);
   cursor: pointer;
-  font-family: var(--font-family-base);
+  font-family: var(--font-family-body);
```
davidagustin: I see the other pull request at https://github.com/ShelterTechSF/sheltertech.org/compare/1648-create-global-variables-colors-and-fonts may give a conflict. So I would change this back and rebase this commit after merging ticket 1648.

davidagustin (changes requested):

davidagustin:
Screen Shot 2020-10-09 at 3 25 37 PM

The update branch button will rebase these commits.

src/components/inline/Card/Card.stories.js
candywang: To match the new Project Card design, I removed the background color and changed the title and description font color to be black.

davidagustin (changes requested):
Demo for light and dark examples needed.

src/components/block/ProgramCard/ProgramCard.js
```diff
@@ -0,0 +1,29 @@
...
+  title: PropTypes.string,
+  description: PropTypes.string,
+};
+
+ProgramCard.defaultProps = {
+  image: PropTypes.string.isRequired,
+  alt: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+};
```
richardxia: I think I was probably a bit unclear with my previous comment about this. `defaultProps` is only used to set the default value of a property when it is not passed in by the caller. In other words, the values of the default props should be of the same type that is require, such as an actual placeholder string.

When you want to mark a prop type as required, it should go in the prop type definitions under the `.propTypes` property. When a prop is required, then you do not need to set a `defaultProp` for it, since it's invalid to instantiate the component without all of its required props set.

You'll want to make a change that looks like this:

```suggestion
ProgramCard.propTypes = {
  image: PropTypes.string.isRequired,
  alt: PropTypes.string.isRequired,
  title: PropTypes.string.isRequired,
  description: PropTypes.string.isRequired,
};
```

src/components/block/ProgramCard/ProgramCard.module.css
```diff
@@ -0,0 +1,23 @@
+.programCardContainer {
+  width: 410px;
```
richardxia: You shouldn't set the width of this component. Under the document I wrote recently, the [Component Design System](https://github.com/ShelterTechSF/sheltertech.org/blob/419f9fa0b1b209408a3f17dac198355355346901/docs/component-design-system.md#space-filling-block-components), this component is what I would categorize as a "space-filling block component". We want the parent component to set a width and for this component to expand to fit it, which will make this component more flexible because it can be used in more situations.

In some cases, it may make sense to set a [`max-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/max-width), which will only set an upper bound on the element width, allowing a parent to make it smaller if it would like to.

src/components/block/ProgramCard/ProgramCard.module.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.img {
+  height: 250px;
+  object-fit: cover;
+  width: 100%;
+}
+
+.title {
+  color: var(--color-black);
```
richardxia: Since we'll want the color of the title and the description to bet set to the same thing at the same time, we should probably set this on the parent element. The `color` property is inherited from parent to child, so you can set it once on the parent element.

src/components/block/ProgramCard/ProgramCard.module.css
```diff
@@ -0,0 +1,23 @@
...
+  font: var(--font-title);
+  line-height: 1.3;
```
richardxia: The `line-height` is actually set as part of the `--font-title` value, so we don't need to override it here with the same value that it's being set to in the first place.

src/components/block/ProgramCard/ProgramCard.module.css
```diff
@@ -0,0 +1,23 @@
...
+  margin-top: 20px;
+}
+
+.description {
+  color: var(--color-black);
+  font: var(--font-body-small);
+  line-height: 1.2;
```
richardxia: This made me realize that we set the line height incorrectly for a few of the styles in the first place when we first defined the `--font-*` variables in https://github.com/ShelterTechSF/sheltertech.org/pull/63.

Looks like Action should have a line height of 1.5, and Body Small should have a line height of 1.2 by default. It's at least the same between the web and mobile typographies, so we can just set them in one place. Do you think you could create a separate PR to fix them? We can merge that separate PR in independently of this one as long as we get rid of the explicit `line-height` properties in this PR.

src/components/block/ProgramCard/ProgramCard.stories.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import ProgramCard from "./ProgramCard";
+import img from "./sample.jpg";
```
richardxia: This is a bit minor, but do you think we could move this image into a subdirectory named `stories/`? This would help avoids us getting confused about which images are actually part of the component vs. which ones are just for demonstration purposes in the story. I think for now, we can just leave that directory in each component directory, but we may want to revisit later to consolidate all Storybook placeholders into one consistent place.

src/components/block/ProgramCard/ProgramCard.stories.js
```diff
@@ -0,0 +1,26 @@
...
+
+export default {
+  title: "Block/ProgramCard",
+  component: ProgramCard,
+};
+
+const Template = ({ image, alt, title, description }) => (
+  <ProgramCard
+    image={image}
+    alt={alt}
```
richardxia: After looking at this here, I'm thinking that we may want to rename `alt` to `imageAlt` to make it clearer that it's the alt text that goes with the image. Since this component currently has props that let you control three different subelements, it's probably a good idea to be clear about which subelement is being affected.

richardxia (commented):
Took another look at this, and I think this is almost there. I think this is probably just one last set of comments before this will be good to go!

richardxia (approved):
LGTM! Thank you!

davidagustin (approved):

PR #63, opened 2020-10-08
1648 create global variables colors and fonts

candywang:
`global.css`
- New global colors and typography added in accordance with the design system.
- Existing global colors and typography variables updated to match design system color and typography names.
- Updated import URL for Google Font Merriweather (300) and Adobe Font Lato (400;700;800).
- Color and typography are separated with comments based on category and subcategory. (Ex: Colors (Interface Colors, Message Colors, Brand Colors), Font Families, Font Sizes)
- Added media query for mobile.

`Button.module.css`
- Implement font property with appropriate global variable to replace font-family, font-weight, and font-size.


src/stylesheets/global.css
```diff
@@ -1,13 +1,71 @@
...
+  --color-midnight-blue: #0f1428;
+
+  /* Font Families */
+  --font-family-hero: 'Merriweather', serif;
+  --font-family-headline: 'Merriweather', serif;
+  --font-family-title: 'Merriweather', serif;
+  --font-family-subtitle: 'Lato', sans-serif;
+  --font-family-caption: 'Lato', sans-serif;
+  --font-family-action: 'Lato', sans-serif;
+  --font-family-body: 'Lato', sans-serif;
```
richardxia: I was thinking that we could create global variables that contained the entirety of the value of the [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property, which lets you set multiple properties at once. This would make it easier to be consistent about all the properties that should be set together.

Here's a short example of what I was thinking:

```css
  --font-hero: 300 67px/1.3 'Merriweather', serif;
  --font-body-medium: 400 20px/1.3 'Lato', sans-serif;
```

Something like that could be used in an actual component with `font: var(--font-body-medium);`

It might be useful to still have a variable to hold the font family strings, since one might want to override a font family somewhere else, and we'll want to make the whole list of fonts and fallback fonts consistent. A similar thing goes for the font sizes, since there are just a few specific sizes that we should be using. The font weights I think we probably care less about, since there's only a few weights that are possible to use, and it's hard to get it wrong.

Putting that all together, I think I'd suggest something that looks like this:

```css
  --font-family-merriweather: 'Merriweather', 'Georgia', serif;
  --font-family-lato: 'Lato', 'Arial', serif;

  --font-size-hero: 67px;
  --font-size-body-medium: 20px;

  --font-hero: 300 var(--font-size-hero)/1.3 var(--font-family-merriweather);
  --font-body-medium: 400 var(--font-size-body)/1.3 var(--font-family-lato);
```

I'd also be fine with directly inlining the `var(--font-size-hero)` within the `--font-hero` property definition, since it is a bit more difficult to read when you have too many variable to cross reference, and since all these variables are within the same file, near each other, I would not be as concerned with them drifting apart from each other. In other words, I think the following is an example of when it is OK to not be [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself):

```css
  --font-family-merriweather: 'Merriweather', 'Georgia', serif;
  --font-family-lato: 'Lato', 'Arial', serif;

  --font-size-hero: 67px;
  --font-size-body-medium: 20px;

  --font-hero: 300 67px/1.3 var(--font-family-merriweather);
  --font-body-medium: 400 20px/1.3 var(--font-family-lato);
```

src/stylesheets/global.css
```diff
@@ -1,13 +1,71 @@
...
+  --font-size-title-w: 30px;
+  --font-size-subtitle-w: 24px;
+  --font-size-caption-w: 18px;
+  --font-size-action-w: 14px;
+  --font-size-body-large-w: 30px;
+  --font-size-body-medium-w: 20px;
+  --font-size-body-small-w: 16px;
+
+  /* Mobile Type Scale */
+  --font-size-hero-m: 40px;
```
richardxia: We could actually do something pretty cool here by making use of media queries for the mobile font sizes, which means that these variables will have the correct values just based on the screen size. See https://css-tricks.com/responsive-designs-and-css-custom-properties-defining-variables-and-breakpoints/#using-css-variables-for-property-values for an article describing this, but it would look like the following:

```css

:root {
  --font-size-hero: 67px;
}

@media screen and (max-width: 600px) {
  :root {
    --font-size-hero: 40px;
  }
}

**Review by richardxia — CHANGES_REQUESTED** [review]
Looking pretty good, but I have a few suggestions here regarding the the font properties.

File: src/stylesheets/global.css (hunk 1 above)
Comment (candywang) [reply src/stylesheets/global.css]: Thanks for the feedback!

In order to make use of media queries for the mobile font sizes, I'll be directly inlining the `var(--font-size-hero)` within the `--font-hero` property definition.

File: src/components/inline/Button/Button.module.css [hunk 3]
```diff
@@ -1,9 +1,9 @@
 .button {
-  background-color: var(--color-brand);
-  border: 1px solid var(--color-brand);
+  background-color: var(--color-sheltertech-blue);
+  border: 1px solid var(--color-sheltertech-blue);
   color: var(--color-white);
   cursor: pointer;
-  font-family: var(--font-family-base);
+  font-family: var(--font-family-body);
```
davidagustin: Should change to  font-family: var(--font-action); this commit should be merged before https://github.com/ShelterTechSF/sheltertech.org/pull/61 so that commit can rebase to these changes.

davidagustin (changes requested):

src/stylesheets/global.css
```diff
@@ -1,13 +1,70 @@
 /* Global settings and variables */
 
 @import '~normalize.css';
-@import url('https://fonts.googleapis.com/css?family=Lato:300i,300,400,700');
+@import url('https://fonts.googleapis.com/css2?family=Merriweather:wght@300&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Lato:wght@400;700&display=swap');
```
davidagustin, line 5: The 800 lato font is available at https://fonts.adobe.com/fonts/lato

src/components/inline/Button/Button.module.css
```diff
@@ -1,9 +1,9 @@
...
   font-size: 14px;
```
davidagustin, line 7: Should be font-size: var(--font-size-action);

davidagustin (changes requested):

src/components/inline/Button/Button.module.css
davidagustin: I talked to @richardxia , these changes are so small that changes in the Button component can be here. In other words, you do not need to change this back. You can do all the appropriate changes for the button here.

src/stylesheets/global.css
richardxia: I don't think we can easily use the fonts from Adobe without a subscription to TypeKit. I think it's fine to use the 700 weight in place of the 800 and just let @derekfidler know.

src/components/inline/Button/Button.module.css
```diff
@@ -1,9 +1,9 @@
...
-  border: 1px solid var(--color-brand);
+  background-color: var(--color-sheltertech-blue);
+  border: 1px solid var(--color-sheltertech-blue);
   color: var(--color-white);
   cursor: pointer;
-  font-family: var(--font-family-base);
+  font-family: var(--font-action);
   font-size: 14px;
   font-weight: 800;
   line-height: 1.5;
```
richardxia: You can't just use `var(--font-action)` on `font-family`, since `--font-action` contains the whole set of values that can be used for the [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property. Instead, you can just replace all of these `font-*` properties with the single `font` shorthand property.

richardxia (commented):
One last request from me for the updated Button component, but this otherwise LGTM!

davidagustin (approved):

richardxia (approved):
Thanks, this LGTM! Great work on getting these all defined in a nice and clean way!

PR #64, opened 2020-10-10
Create donation project card component

ivanleeswe:
1. Created JS, CSS, and storybook files for the donation project card component.

2. In the JS file, title and subtitle are being passed down from the parent component to be used to render on the HTML page.

3. Styled the component in the CSS file.

4. Built the storybook template in order to use its platform.


src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
+.rectangle39 {
```
richardxia: I'm guessing that you copied these names from the Figma designs. I don't think @derekfidler is giving them any specific names, so you should probably give them more sensible ones when you're implementing them. For example, since this `rectangle39` is the top level of the component, you could call it `card` or `container`.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  background-image: url('./blackbackground.jpg');
```
richardxia: Since we'll want to use this component with many different background images, we'll want to make this parameterizable. It's probably best to leave this unset in the CSS and instead set it as a React [inline style](https://reactjs.org/docs/dom-elements.html#style).

Also, I suspect that we'll probably want to set other CSS properties for the background image, but it's not obvious when you have a solid, black background, so you may want to choose a different background image for the Storybook story so that we can guarantee that the image is not distorted.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  height: 240px;
+  position: absolute;
```
richardxia: We shouldn't set `position: absolute` because this will take the item [out of the flow of the document](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout/In_Flow_and_Out_of_Flow). I think almost without exception, the outermost HTML element of a component should never have custom positioning, and `position: absolute` should mostly only ever be used to position a child element relative to its parent.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  width: 240px;
```
richardxia: Let's not set an explicit width for this component, since it may make it less reusable in other contexts. We can assume that the parent component that uses this component will set the bounding box for the component and that this component will fill the space available from its parent. See the document I wrote up recently about this concept: https://github.com/ShelterTechSF/sheltertech.org/blob/main/docs/component-design-system.md#space-filling-block-components

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.frame20 {
+  align-items: flex-start;
```
richardxia: This is the default value of `align-items`, so you don't need to set it here.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  bottom: 0;
+  display: flex;
+  flex-direction: row;
+  height: 72px;
+  position: absolute;
+  width: 240px;
```
richardxia: You shouldn't explicitly set a width here, since you'll want this to inherit the width of its parent. If you redundantly set the width on multiple child components, it means that you have to update multiple places when you want to change the value, and it also makes the component less flexible if you try to override properties in specific situations.

In general, you want to rely on the [cascade](https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade) to have properties automatically flow from parent to child.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.rectangle271 {
+  align-self: flex-end;
+  background: #0025e4;
+  height: 72px;
+  position: static;
+  right: 96.67%;
```
richardxia: `position: static` is the default value of `position`, and it is how all elements in HTML behave by default, so you do not need to set this. Also, the `right`, `top`, `bottom`, and `left` properties have no effect when an element is positioned statically, since elements are positioned based on where they are in the document, so you should delete the `right` property here.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  flex-direction: row;
+  height: 72px;
+  position: absolute;
+  width: 240px;
+}
+
+.rectangle271 {
+  align-self: flex-end;
+  background: #0025e4;
+  height: 72px;
```
richardxia: Instead of setting the height to `72px` here, you should probably set it to `100%` so that it just fills the height of its parent. This follows the general rule of having elements determine their size from their parents as much as possible.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  display: flex;
+  flex-direction: row;
+  height: 72px;
+  position: absolute;
+  width: 240px;
+}
+
+.rectangle271 {
+  align-self: flex-end;
+  background: #0025e4;
```
richardxia: If https://github.com/ShelterTechSF/sheltertech.org/pull/63 merges before this PR, it'd be good to replace all the colors and font properties with the variables that are being defined there.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.space {
+  align-self: flex-end;
+  background: #000;
+  height: 72px;
+  position: static;
+  right: 96.67%;
+  width: 12px;
+}
```
richardxia: Instead of creating a space component, I'd probably just put a margin between the blue bar and the text component.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+
+.hotelname {
```
richardxia: Since this may not always hold the name of a hotel, we should give this a more generic name, such as `title`.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  margin: 0 1px;
+  order: 1;
+  position: static;
+  top: 45.07%;
+  top: 32px;
+  width: 220px;
+}
+
+.rightbox {
+  flex-direction: column;
```
richardxia: This won't do anything without `display: flex`. None of the flexbox parent properties will have any effect unless if `display: flex` is set.

But also, I'm not sure if we need a flexbox here, since I think we'll just want the title text and the other text to just be laid out normally.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  right: 96.67%;
+  width: 12px;
+}
+
+.hotelname {
+  align-items: flex-end;
+  align-self: center;
+  bottom: 56.34%;
+  color: var(--color-white);
+  display: flex;
```
richardxia: There's no use in making this `display: flex` because this element has no children. This also means that `align-items` has no use.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  background: #000;
+  height: 72px;
+  position: static;
+  right: 96.67%;
+  width: 12px;
+}
+
+.hotelname {
+  align-items: flex-end;
+  align-self: center;
```
richardxia: See my comment below about the `rightbox`, but if we do not make that element a flexbox parent, then `align-self` won't have any effect, since it only applies to flexbox children.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  bottom: 56.34%;
```
richardxia: This `bottom` property won't have any effect when you are letting elements by positioned statically.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  color: var(--color-white);
+  display: flex;
+  font-family: Lato;
+  font-size: 24px;
+  font-style: normal;
+  font-weight: bold;
+  line-height: 130%;
```
richardxia: When https://github.com/ShelterTechSF/sheltertech.org/pull/63 merges in you'll be able to use the font variables being defined here.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  margin: 0 1px;
```
richardxia: Since this is such a miniscule amount, I'm guessing that it's probably just a slight error or oversight in the design itself.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+  font-style: normal;
+  font-weight: normal;
+  height: 39px;
+  line-height: 120%;
+  margin: 0 1px;
+  order: 1;
+  position: static;
+  top: 45.07%;
+  top: 32px;
+  width: 220px;
```
richardxia: I think many of my comments related to the `.hotelname` element apply to this, so rather than individually leaving comments on each one of them, I'll just generally say you should make similar kinds of changes here.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import DonationProjectCard from "./DonationProjectCard";
+import background from "./blackbackground.jpg";
+
+export default {
+  title: "Inline/DonationProjectCard",
+  component: DonationProjectCard,
+};
+
+const Template = () => (
```
richardxia: You'll want to actually add props to this so that you can bind them as Storybook controls. Otherwise the `.args` that you defined below are just getting ignored.

richardxia (changes requested):
Thanks for the first PR! I left some comments, but thanks for pretty faithfully building out the component!

As a general comment, I think you may be directly copying all of the CSS that you will find in Figma. You probably shouldn't do that, since that CSS is being automatically generated from the designs, but not necessarily being handwritten and hand-curated. Although looking at the generated CSS is useful, you shouldn't just copy all of them.

The other reason for not just copying the CSS is that the designs may be designed for a static layout or screen size, whereas in code, we want to build the components to be responsive and scale to any size available, so we'll have to think about how it should be expressed in HTML and CSS. It's our job as developers to do this :)

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: Understood. I wasn't sure whether the CSS properties in Figma were automatically generated or written by someone where I have to use them specifically. After reading most of your comments, it makes sense now it is the former.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: I deleted the background-image property from the CSS page and added the following inline in the story.js file.

style={{
      backgroundImage: `url(${background})`,
      backgroundSize: "240px 240px",
      backgroundRepeat: "no-repeat",
    }}

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Width deleted.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Align-items deleted.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Width deleted. The position changed to inherent (would that work?). I read about the cascade and looked at examples but still a little iffy about it.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Thanks for the explanation about the position: static. Right property deleted.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Height set to 100%.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 The background-color property changed to using the global.css variable.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 The space component deleted. Margin added.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Changed hotel name to title and number of people to subtitle.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Thanks for your explanation. Deleted .rightbox property.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Deleted align-self and display.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Deleted align-items and align-self.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Deleted bottom property.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Changed font-family to using global.css variable. Will change others once updated.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Deleted font-style, line-height, and margin.

src/components/inline/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Deleted all except color and font-family.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: When I add props "title" and "subtitle" as parameters, it becomes unused and the lint shows errors as shown below.

  10:21  error  'title' is defined but never used     no-unused-vars
  10:28  error  'subtitle' is defined but never used  no-unused-vars

const Template = ({ title, subtitle })


Need advice.

src/components/inline/DonationProjectCard/DonationProjectCard.js
```diff
@@ -0,0 +1,29 @@
...
+
+DonationProjectCard.propTypes = {
+  title: PropTypes.string,
+  subtitle: PropTypes.string,
+};
+
+DonationProjectCard.defaultProps = {
+  title: PropTypes.string,
+  subtitle: PropTypes.string,
+};
```
davidagustin: Props should be required:

subtitle: PropTypes.string.isRequired,

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
davidagustin: const Template = (args) => (

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,23 @@
...
+  <div style={{ backgroundImage: `url(${background})` }}>
+    <DonationProjectCard title="Star Hotel SRO" subtitle="280 People" />
```
davidagustin: title={args.title} subtitle={args.subtitle}

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,23 @@
...
+  </div>
+);
+
+export const DefaultDonationProjectCard = Template.bind({});
+
+DefaultDonationProjectCard.args = {
+  data: {
+    title: "Star Hotel SRO",
+    subtitle: "280 People",
```
davidagustin: This is just my style but I think subtitle should be camel cased. @richardxia What do you think?

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import DonationProjectCard from "./DonationProjectCard";
+import background from "./blackbackground.jpg";
+
+export default {
+  title: "Inline/DonationProjectCard",
```
davidagustin: I remember in @cw323 card commit that @richardxia  wanted to move the cards into the block directory. So the file path should be src/component/block/WhateverComponentHere. Then this should change to Block/DonationProjectCard.

davidagustin (changes requested):

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import DonationProjectCard from "./DonationProjectCard";
+import background from "./coralreef.jpg";
+
+export default {
+  title: "Inline/DonationProjectCard",
+  component: DonationProjectCard,
+};
+
+const Template = ({ title, subtitle }) => (
```
davidagustin: const Template = (args) => (

src/components/inline/DonationProjectCard/DonationProjectCard.js
ivanleeswe: 👍 Changed to 

subtitle: PropTypes.string.isRequired,

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: 👍 Changed to 

title={args.data.title} subtitle={args.data.subtitle}

since my default is 

DefaultDonationProjectCard.args = {
  data: {
    title: "Star Hotel SRO",
    subtitle: "280 People",
  },
};

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: Like subTitle?

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: 👍 The Block folder created in the component folder separate from the inline folder. Title changed to "title: "Block/DonationProjectCard"".

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: 👍 Changed to (args) param.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
davidagustin: Yup

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
davidagustin: You can take out the data object and put its properties 1 degree upward in the chain.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: I will wait for Richard's response to this.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: 👍 Deleted data object.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,33 @@
...
+  position: relative;
+}
+
+.bottombox {
+  bottom: 0;
+  display: flex;
+  flex-direction: row;
+  height: 72px;
+  position: absolute;
+  width: inherit;
```
richardxia: I'm going to respond to the comment thread from https://github.com/ShelterTechSF/sheltertech.org/pull/64#discussion_r502761928 here, since I think this line is what that ended up turning into in the latest version of this file.

> 👍 Width deleted. The position changed to inherent (would that work?).

I think you meant to say "_width_ changed to _inherit_"? In any case, I would not recommend doing this. Block-level elements by default take up the width of their containing parent, so it's usually unnecessary to ever explicitly set any property to `inherit` unless if you are overriding something back to the original behavior, such as in a media query.

However, now that we've cut down on the number of extraneous properties in this component, I can more easily see another issue with how we're doing this. When you [absolutely position an element](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout/In_Flow_and_Out_of_Flow#Absolute_Positioning), it takes it out of the normal flow of the document, which means it will not determine its width based on the width of its parent. Thinking about this more, since this component should always have a width equal to its parent's width, we probably actually want to make the parent a flex parent and set the `align-items` property so that the this bottom box always ends up aligned to the bottom.

> I read about the cascade and looked at examples but still a little iffy about it.

Without going into the technical minutia, the easiest way of thinking about this is that there are certain properties that will affect all children and descendants of the element that it is being applied to. For example, if I set the `font-family` of an element, it and all of its descendants will receive that font-family. There's no one simple rule for knowing which properties behave this way, so you'll always want to look this up for each property, but a general rule of thumb is that it applies for anything that would "make sense" to apply to all elements within a block.

This [StackOverflow post](https://stackoverflow.com/questions/5612302/which-css-properties-are-inherited) has a few lists of properties that generally behave this way.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,33 @@
...
+  height: 100%;
+  margin-right: 12px;
+  width: 8px;
+}
+
+.title {
+  color: var(--color-white);
+  font-family: var(--font-family-lato);
+  font-size: var(--font-size-subtitle);
+  font-weight: bold;
```
richardxia: One of the things that set up in https://github.com/ShelterTechSF/sheltertech.org/pull/63 is being able to set multiple of these properties at the same time with the [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property.

```suggestion
  font: var(--font-subtitle);
```

src/components/block/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.subtitle {
+  color: var(--color-white);
+  font-family: var(--font-family-lato);
```
richardxia: Similar to above, you should use the `font` shorthand variables that we've defined. This one I believe corresponds to `--font-body-small`.

src/components/block/DonationProjectCard/DonationProjectCard.js
```diff
@@ -0,0 +1,28 @@
...
+
+DonationProjectCard.propTypes = {
+  title: PropTypes.string,
+  subtitle: PropTypes.string,
+};
+
+DonationProjectCard.defaultProps = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
+};
```
richardxia: This isn't valid. `isRequired` goes on the type definitions in `propTypes`. [`defaultProps`](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) is where you provide a value to use when the property is not passed in.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
richardxia: I would leave it as `subtitle`, since that's a full English word. I'd usually only camelcase something if it were two distinct words being combined into a single variable name. I might use camelcase for compound words in certain situations, but "subtitle" is not a compound word ("sub" is not a standalone word here), so this wouldn't apply here.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
richardxia: I think I'd actually prefer the way it was originally with the destructured object in the function arguments, since it makes it clearer which properties are expected to be in the argument and it removes a little bit of visual noise by eliminating the `args` variable in the entire `Template` body. Do you have a specific reason for preferring `(args)`, @davidagustin?

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: Ok, keeping it as "subtitle."

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
ivanleeswe: 👍 Changed it back to a destructured object { title, subtitle }.

davidagustin: It was in an example component online @richardxia

src/components/block/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Makes sense! Deleted width property with inherent value and added flexbox in parent while the bottom box uses align-self to position itself at the bottom it.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Changed.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Done.

src/components/block/DonationProjectCard/DonationProjectCard.js
ivanleeswe: 👍 in that case, I changed the default value to null if there is no title or subtitle.

richardxia: Thanks! I think you may have not pushed those changes up to GitHub yet, since I don't see them here.

I'd also probably suggest that we make the `title` required, which means that you don't need to pass in a default value, since you are required to pass in a value. I don't think this component would make any sense without the title, which is why I think we should probably make it required. I think it's probably also worthwhile making the subtitle required, even if I could kind of see the component working without one.

src/components/block/DonationProjectCard/DonationProjectCard.stories.js
```diff
@@ -0,0 +1,27 @@
...
+  title: "Block/DonationProjectCard",
+  component: DonationProjectCard,
+};
+
+const Template = (args) => (
+  <div
+    style={{
+      backgroundImage: `url(${background})`,
+      backgroundSize: "240px 240px",
+      backgroundRepeat: "no-repeat",
```
richardxia: A few comments about this:

1. I think we should make the background image a prop of the `<DonationProjectCard>` component so that we can pass it into that component. I would say that the image is "part of" the component, not merely something placed below the component.
2. These background properties are probably not what we want. If the card component itself is not 240px by 240px, then the image will not be sized properly. Also, if the image itself is not a square, this will distort the image by stretching or shrinking the height and width in a non-uniform way. Instead, I would suggest setting [`background-size: cover`](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size#cover), which will scale the image to be large enough to cover the diagram and crop it, rather than stretching it. Also, this could probably go in the main CSS for the component, since we'll probably want to apply by default to all instances of the component regardless of the image.

richardxia (commented):
Left a few more comments, but I think you haven't pushed up your latest changes to this branch, @ivanleeswe, since several of the things that you commented on, saying that they were done, aren't showing up in this PR.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,33 @@
+.container {
+  height: 240px;
+  position: relative;
+}
+
+.bottombox {
```
richardxia, line 8: One thing I noticed after you added in a real picture (thanks, btw!) is that when an image is a bit too bright, it becomes more difficult to read white text on top of it. That caused me to take a look at the [design in Figma](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=665%3A2) a bit more closely to see that there's actually a small, partially transparent background gradient on this bottom element, which helps darken the background just below the text and make it more readable. You should also add that background gradient.

src/components/inline/DonationProjectCard/DonationProjectCard.stories.js
richardxia: > It was in an example component online @richardxia

OK, if you don't have a strong preference for the `args` parameter, then I think I'd prefer the destructured form.

davidagustin: I think the destructured form would be a lot better because it does reduce visual noise and also destructuring is a recurring style to react so it is more consistent with the general code.

src/components/block/DonationProjectCard/DonationProjectCard.js
```diff
@@ -0,0 +1,28 @@
...
+
+DonationProjectCard.propTypes = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
+  background: PropTypes.string.isRequired
+};
+
+DonationProjectCard.defaultProps = {
+
+};
```
richardxia: You should delete this if it no longer contains any content.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
```diff
@@ -0,0 +1,30 @@
...
+
+.bar {
+  background: var(--color-sheltertech-blue);
+  height: 100%;
+  margin-right: 12px;
+  width: 8px;
+}
+
+.title {
+  color: var(--color-white);
```
richardxia: Let's put this color property on the `.container` selector, since it will be inherited by the children elements, so you can just set it in one place.

richardxia (commented):
You're currently failing the lint checks, so you'll want to fix up those issues. 

Also, I'm noticing that you're missing one part of the design of the component, which is the semitransparent black background below the text, which helps make the text more visible on top of bright images.

src/components/block/DonationProjectCard/DonationProjectCard.js
ivanleeswe: 👍 Deleted.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Removed color property from title and subtitle children elements and added a color property with the white value to the parent.

src/components/block/DonationProjectCard/DonationProjectCard.module.css
ivanleeswe: 👍 Gradient added. Implement flex-grow to fill the remainder of the flexbox child with the gradient.

src/components/block/DonationProjectCard/DonationProjectCard.js
```diff
@@ -0,0 +1,27 @@
...
+        <div className={s.title}>{title}</div>
+        <div className={s.subtitle}>{subtitle}</div>
+      </div>
+    </div>
+  </div>
+);
+
+DonationProjectCard.propTypes = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
```
cw323, line 23: I think it'd be better to keep the property names that serve the same purpose to be consistent throughout the Card components. Do you mind changing this 'subtitle' key to 'description'?

richardxia (approved):
LGTM! Thanks for working on this!

davidagustin (approved):

candywang (approved):

PR #65, opened 2020-10-10
1649 Create Footer CTA Card Component

davidagustin:
- HTML, CSS, Storybook and index file created.


davidagustin:
Please notice the file directory created with this component is src/components/block

davidagustin:
@richardxia the button behavior when clicked should be dealt with before it gets too ugly. There is no onClick action therefore I don't have props to pass to the button for that behavior.

src/components/block/FooterCTACard/FooterCTACard.stories.js
```diff
@@ -0,0 +1,19 @@
...
+import FooterCTACard from "./FooterCTACard";
+
+export default {
+  title: "Block/FooterCTACard",
+  component: FooterCTACard,
+};
+
+const Template = ({ subtitle, body, buttonText }) => (
+  <div style={{ backgroundColor: "black", width: "480px" }}>
+    <FooterCTACard subTitle={subtitle} body={body} buttonText={buttonText} />
```
richardxia:
```suggestion
    <FooterCTACard subtitle={subtitle} body={body} buttonText={buttonText} />
```

You mistyped the prop name here, which causes the title to not actually show up in the Storybook story.

src/components/block/FooterCTACard/FooterCTACard.js
```diff
@@ -0,0 +1,28 @@
...
+  subtitle: PropTypes.string,
+  body: PropTypes.string,
+  buttonText: PropTypes.string,
+};
+
+FooterCTACard.defaultProps = {
+  subtitle: PropTypes.string.isRequired,
+  body: PropTypes.string.isRequired,
+  buttonText: PropTypes.string.isRequired,
+};
```
richardxia: This is not a correct usage of `defaultProps`. `defaultProps` is meant to hold the value that will be used in place of a prop if it is not passed into the component. See https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values for more details.

The `.isRequired` goes on the prop type definitions under `.propTypes`.

src/components/block/FooterCTACard/FooterCTACard.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from "prop-types";
+import React from "react";
+import Button from "../../inline/Button";
+import s from "./FooterCTACard.module.css";
+
+const FooterCTACard = ({ subtitle, body, buttonText }) => (
```
richardxia: I think it's a little bit confusing to call this prop the "subtitle" when there is no "title". Functionally, this prop acts as the title of this card, so it's probably better to name it "title".

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
+.body {
+  color: var(--color-white);
+  font: var(--font-body-medium);
+  line-height: 1.3;
```
richardxia:
```suggestion
```

`line-height` is already being set with the `--font-body-medium` shortcut that we have, so there's no need to set it to the exact same value again here.

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 15px;
+  max-width: 350px;
+  text-align: center;
```
richardxia: Since we generally want all text within this component to be center-aligned, you should just put this on the `.container` element and allow it to be inherited by child components.

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.button {
+  margin-top: 30px;
+  text-align: center;
+}
+
+.container {
+  height: 400px;
+  width: 480px;
```
richardxia: You should not set the width of this component, since that will prevent any parent component from controlling the size of it. For example, if the parent allocated less space for this component, it would not be able to resize this component and this component would end up being cut off. See this example where I set the width of the parent element in Storybook to have a width of 320px:

Screen Shot 2020-10-11 at 1 33 14 PM

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
...
+  font: var(--font-body-medium);
+  line-height: 1.3;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 15px;
+  max-width: 350px;
+  text-align: center;
+}
+
+.button {
```
richardxia: This class name a little bit misleading because it's not actually used on the button itself, but rather the block-level element that wraps it. How about you rename this to `.buttonWrapper`?

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
+.body {
+  color: var(--color-white);
```
richardxia: Since this component is designed with a dark background, we'll want all the text to be light-colored. Since `color` is an inherited property, you should set this on the top-level element of this component so that all child elements inherit it.

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,28 @@
...
+.container {
+  height: 400px;
+  width: 480px;
+}
+
+.subtitle {
+  color: var(--color-white);
+  font: var(--font-subtitle);
+  line-height: 1.3;
+  padding-top: 109.5px;
```
richardxia: This exact and fixed padding of 109.5px was probably not Derek's original design intent, since 1) most browsers/devices can't even draw half pixels and 2) this is just a strange number. I can see that you got this number directly from the Design System Figma for this component, but if you look at [the designs for the whole pages of the site](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1602), you can see that not all the cards have the same amount of text in them, specifically that some have two lines of body text and others have three.

The offset from the top of the component is not the same between the different components, but what is consistent about all three instances of this card is that they are all vertically centered. I think what you probably want to do is make the container a flex parent and vertically center all of these items in between.

src/components/block/FooterCTACard/FooterCTACard.stories.js
```diff
@@ -0,0 +1,19 @@
...
+  </div>
+);
+export const DefaultFooterCTACard = Template.bind({});
+DefaultFooterCTACard.args = {
+  subtitle: "Get Involved",
+  body: "Volunteer with us and put your professional skills to good use.",
+  buttonText: "JOIN US",
```
richardxia: This text should not be in all caps because the button component itself is the one that is setting the [`text-transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) property to always uppercase the text. In other words, the all-caps-ness is part of the style, not the content.

richardxia (changes requested):
> @richardxia the button behavior when clicked should be dealt with before it gets too ugly. There is no onClick action therefore I don't have props to pass to the button for that behavior.

Agreed, I think that's something we should probably start working on right after this current set of components. It's going to be a bit more complicated than just having an `onClick` handler, since we'll want to support at least three types of buttons: 1) links to internal pages that Gatsby knows about (e.g. with the [GatsbyLink](https://www.gatsbyjs.com/docs/gatsby-link/) component or API), 2) links to external pages, and 3) buttons that trigger some JavaScript event rather than opening a page, such as opening a popup modal. I can scope that out as a ticket.

src/components/block/FooterCTACard/FooterCTACard.module.css
```diff
@@ -0,0 +1,24 @@
+.body {
+  font: var(--font-body-medium);
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 15px;
```
candywang: Would it be better to use the shorthand margin property? For example, `margin: [top] [left-and-right] [bottom];` or in this case, `margin: 15px auto 0px;`.

src/components/block/FooterCTACard/FooterCTACard.stories.js
```diff
@@ -0,0 +1,19 @@
...
+  title: "Block/FooterCTACard",
+  component: FooterCTACard,
+};
+
+const Template = ({ title, body, buttonText }) => (
+  <div style={{ backgroundColor: "black", width: "480px" }}>
+    <FooterCTACard title={title} body={body} buttonText={buttonText} />
+  </div>
+);
+export const DefaultFooterCTACard = Template.bind({});
```
candywang: Can we add a new line here?

src/components/block/FooterCTACard/FooterCTACard.stories.js
```diff
@@ -0,0 +1,19 @@
...
+DefaultFooterCTACard.args = {
+  title: "Get Involved",
+  body: "Volunteer with us and put your professional skills to good use.",
```
candywang: This Footer CTA Card `body` property serves the same purpose as the Program Card `description` property. I think we should keep the keys consistent throughout all the Card components. Would you prefer 'body' or 'description'?

src/components/block/FooterCTACard/FooterCTACard.module.css
richardxia: 👍, except instead of `0px`, just write `0` because the unit is not necessary and also meaningless for a length of 0. I believe we probably already have this enforced by a [stylelint rule](https://stylelint.io/user-guide/rules/length-zero-no-unit), but just in case if we don't, I just wanted to put that out there.

src/components/block/FooterCTACard/FooterCTACard.stories.js
richardxia: 👍 for consistency across components. I have a slight preference for "description", since I feel that "body" has the connotation of a large body of text, but these are pretty small components and the text here is limited to one or two sentence. But I could see this making sense either way.

richardxia (approved):
LGTM!

candywang (approved):

PR #66, opened 2020-10-11
Add ThreeParagraphBlock component.

richardxia:
This is still a WIP, but it's probably good enough to at least start reviewing or having a discussion about.

The main component it builds is the [Three Paragraph Block](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14) from the Design System Figma, but more importantly it tries to set up a common grid system that we can use on other components as well.

I left a lengthy comment in the CSS describing the techniques I used, so let me copy it here for convenience:

---

This component makes use of a few advanced techniques to achieve a
magazine-style layout that involves sections that break out of grid as well
as backgrounds that can bleed out of the grid to either side.

It consists of two, unaligned layers of grids, where the outer grid is solely
used for the background bleed and the inner grid is aligned to the main,
12-column grid of the site.

The background bleed is accomplished by manually positioning a background div
and the inner grid content such that they are overlapping, but where the
backgrid div spans the left margin area and the center area while the inner
grid content only spans the center area.

The inner grid is set up with our standard 12-column grid, where elements are
placed along the grid, but the image elements are relatively positioned to
break out of the grid.

https://en.wikipedia.org/wiki/Bleed_(printing)
https://mastery.games/post/overlapping-grid-items/

---

Here's a screenshot of what it currently looks like for me:

Screen Shot 2020-10-11 at 9 20 26 AM

---

There are a couple of things I'd still like to do with this before merging:

1. Factor out the grid stuff so that it's easier to reuse or at least consistently apply this style to the rest of the components we bring in.
2. Implement the mobile version of this component, which has a single-column layout.
3. Get the responsive behavior to look right. @derekfidler's designs are for a 1440px-wide window, which is actually quite large, and I haven't figured out what to do at smaller resolutions and exactly where to set the breakpoints. I know that there are suggested breakpoints in the Design System Figma, but it's a bit hard for me to figure out exactly how things should look above 600px and below 1440px.


richardxia:
Actually, the following may be a better screenshot that shows the grid and how items are aligned to it:

Screen Shot 2020-10-11 at 9 34 21 AM

Notice that all the main content is aligned to one of the 12 grid columns, that the gray background bleeds to the left to some maximum content width, and that the images are shifted out of the main layout.

davidagustin:
I realized this is a WIP. Should we cherrypick this in order for us to use the grid aware variables for our task related to configuring the grid?

richardxia:
> I realized this is a WIP. Should we cherrypick this in order for us to use the grid aware variables for our task related to configuring the grid?

I think you can either cherry-pick or just copy the bits and pieces that are relevant, since there are only a few lines of variables I defined in `globals.css`. One thing that I want to do before I consider this PR to be "ready" is to get the [PostCSS custom media queries](https://github.com/postcss/postcss-custom-media) plugin installed so that our media queries can be consistent. I'm working on that right now, but probably won't be able to make more progress until after the work day is over today.

richardxia:
I've removed the WIP here, since I think this is in as good enough shape, and I don't want to further block us from making use of the stuff and conventions introduced in this PR. Please take a look and let me know what you think, and if you've already started trying to apply this pattern for other components, then let me know if there was anything that could be improved or anything that you felt like you had to do a lot of code duplication in your own component.

davidagustin (dismissed):

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.stories.js
```diff
@@ -0,0 +1,71 @@
...
+  paragraph1Description:
+    "Our strength is rooted in our growing community of volunteers who have been mobilized to help the most vulnerable among us-the elderly, the children, domestic abuse survivors, single mothers and others struggling with homelessness who are not getting their basic digital needs met.",
+  paragraph2Title: "Empower your organization",
+  paragraph2Description:
+    "We work with government agencies and nonprofits to empower their programs and build better connections between agencies. We build alongside your teams, so everyone is trained and prepared for handoff. We build alongside your teams, so everyone is trained and prepared for handoff.",
+  paragraph3Title: "Donate to end the digital divide",
+  paragraph3Description:
+    "A monthly donation helps change the lives of those in need. Whether it’s providing free internet access to someone for the night or connecting a brand new shelter to the internet, your support will have a sustaining impact.",
+  ctaTitle: "Volunteer, donate, or reach out to our partnerships team",
+  ctaButtons: [{ text: "Volunteer", link: "/foo" }],
```
davidagustin: Screen Shot 2020-10-16 at 8 49 26 PM

The Figma Wireframe shows 3 buttons: Volunteer, Donate and Work With Us

Should you add those buttons also?

davidagustin (changes requested):
Buttons need to be added

davidagustin:
When reducing the resolution in the storybook file. What constitutes the selection of the images that stays vs removes?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.stories.js
richardxia: The original mockup didn't have them, but the usage of this component on the home page does. After starting to build out the actual home page component, however, I realized that my current component has a bug where if you have multiple buttons, there's no margin between them, so let me add another button and ensure that they have appropriate spacing.

richardxia:
> When reducing the resolution in the storybook file. What constitutes the selection of the images that stays vs removes?

I just went with what the design mockups had, where only the third image remained in the mobile view: https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14.

richardxia:
> Buttons need to be added

Just added an additional button to the Story, and I had to fix up some of the styling to ensure proper spacing between them. Could you take a look at the latest diff and approve if it LGTY?

davidagustin (approved):

richardxia:
Thanks! Let me merge this in so that we can start getting other PRs rebased on top of the `main` branch.

PR #69, opened 2020-10-13
1648 create global variables colors and fonts

candywang:
Changed the line height for action and body small typography.


davidagustin (approved):
Yes, we discussed this in our meeting and Richard explained this to me also.

richardxia:
The final PR diff here looks fine, but I just wanted to comment on something that I think you're probably doing in your local development workflow, based on what I'm seeing in the PR.

I see that there are a lot of unrelated commits in the commit history displayed in this PR. Since we're exclusively doing squash merges when merging PRs into the main branch in this repo, that means that a simple "merge-based" development workflow may not always do what you are expecting. For example, if you continued to do work along the same commit from a previous branch, those commits will not have been registered as having merged into `main` because we squashed them into a new, solitary commit.

Instead, what you'll want to do is start work on new features from the most up-to-date `main` branch. E.g. `git checkout main`, `git pull`, and `git checkout -b <new-branch-name>`. When you want to update your branch with what's in `main`, you may want to opt for a `git rebase` workflow, although if you are not as familiar with Git operations that change history, you may want to create a backup of your code before running it. See https://git-scm.com/book/en/v2/Git-Branching-Rebasing for the Git Book's chapter on a rebase-oriented workflow.

I'm happy to explain this topic more some other time, if you have any questions.

PR #70, opened 2020-10-14
Add PostCSS plugin for custom media queries

richardxia:
I added the [postcss-custom-media](https://github.com/postcss/postcss-custom-media) plugin to our site so that we can start using the future, not-yet-standardized [CSS custom media query syntax](https://www.w3.org/TR/mediaqueries-5/#custom-mq).

[CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) (also known as CSS variables) can't be used in media queries because they are defined inside of selectors and are scoped to those selectors and the inheritance hierarchy. Instead, you have to use a separate syntax for defining custom values for media queries.

No browsers support this yet, as the specification is in very early stages, but with PostCSS, we can use it in our code and have PostCSS process it so that the variables are replaced with their values.

I defined variables following the up/down style that many breakpoint libraries follow. `--tablet-and-down` means for tablet-sized screens and down, and `--laptop-and-up` means for laptop-sized screes and up. I was debating whether to name these variables after device types or whether to use the Bootstrap-style xs, sm, md, lg, xl names. I find the device naming scheme easier to remember, but you have to keep in mind that devices can display things in different screen sizes, such as holding a mobile device sideways gives you a wider screen and resizing a window on a laptop/desktop can give you a smaller screen.

I'm open for comments and suggestions here, since now's the best time to rename these, before we start using them in many places.


davidagustin:
I know we are discouraged in using SCSS or SASS. But wouldn't it be alot easier to use a technology that already exists instead of installing a new library?

https://stackoverflow.com/questions/9122195/using-sass-variables-with-css3-media-queries

A mixin can be used:

    @mixin styling($base-width) {
        // your SCSS here, e.g.
        #Contents {
            width: $base-width;
        }
    }

    @media screen and (max-width: 1170px) {
        @include styling($base-width: 960px);
    }
    @media screen and (min-width: 1171px) {
        @include styling($base-width: 1160px);
    }




Once there is support for CSS custom properties we can easily just change SCSS back to CSS rather than working with the changes of a new library.

candywang (approved):

richardxia:
I will admit that this is partially an experiment of mine to see what it's like to not use Sass. That said, my rationale for avoiding Sass is because many of the features of Sass that made it attractive as a tool can be replaced with things that are part of the CSS standard now and supported natively by browsers.

`calc()` is natively supported by all modern browsers; variables can be implemented with CSS custom properties, which can follow the cascade; and imports can be handled with Webpack and loaders, which many web stacks use now. The benefit of using native features of CSS is that they bring you closer to the actual CSS language itself, rather than introducing another layer, and over time, the native support will be the one that survives in the end, since that is what the browsers have chosen to adopt.

Regarding "But wouldn't it be alot easier to use a technology that already exists instead of installing a new library?": "newness" is relative. This PostCSS plugin has existed since 2014, so I would hardly call it "new", and this syntax, although not finalized, is present in a [draft specification for a new revision of CSS media queries](https://www.w3.org/TR/mediaqueries-5/#custom-mq).

Also, from our app's perspective, adding Sass is adding a "new library", since it is something that this app does not use yet. Sass is a much "larger" dependency than this plugin, and it comes with a lot of other functionality that we may not need. I am perhaps biased by my past experience with [CoffeeScript](https://coffeescript.org/), since I learned both that language and Sass at the same time, but most of CoffeeScript's useful features fed into ES6, which became natively supported by browsers, and CoffeeScript was left by the wayside.

You've asked a valid question, but I think I'd like to try this out longer without introducing Sass into the codebase, since I feel that the web has finally evolved enough to render Sass unnecessary, which is something I've been waiting for for almost 10 years :)

davidagustin:
Ahh I see now. You were using a specific plugin within the POSTCSS repo at https://github.com/postcss/postcss-custom-media ; Sorry the website and documentation was misleading and gave me the impression that it installed all the modules associated with POSTCSS. This specific plugin takes less space, therefore better space-wise compared to SCSS/SASS ;

I would assume that since POSTCSS is in the global variables file, it wouldn't be too much of a change if the CSS media queries gets revised.

-Never knew there was a consortium for the world wide web, very interesting. CoffeeScript looks like Ruby to me too.

davidagustin (approved):

richardxia:
> This specific plugin takes less space, therefore better space-wise compared to SCSS/SASS

Yes, although my reasons are more about features/syntax that we rely on, rather than just the size in bytes of the dependency itself. The fewer things we use, the simpler our program is, and the less someone needs to learn in order to work in our codebase.

> I would assume that since POSTCSS is in the global variables file, it wouldn't be too much of a change if the CSS media queries gets revised.

Just to be clear, PostCSS is the name of a generic tool that can parse and transform CSS. The actual transformations are written as plugins for PostCSS, and many of the transformations are not introducing their own syntax, but rather taking future CSS syntax or current syntax that does not have 100% browser support yet and transpiling it to older syntax. In this case, there's no "PostCSS in the global variables file" since PostCSS is not a language.

But yes to your point that it should be easy for us to globally adjust these breakpoints in the future.

> CoffeeScript looks like Ruby to me too.

Yup, it was hugely inspired by Ruby, and in fact, the initial compiler was written in Ruby!

PR #71, opened 2020-10-15
1659 create homepage large paragraph section

ivanleeswe:
1. Implement HomePageParagraph component to render the text passed into the parameter.
2. Styled the component.
3. Implement the storybook file to display component.


src/components/block/HomePageLargeParagraph/HomePageParagraph.stories.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+import HomePageParagraph from "./HomePageParagraph";
+
+export default {
+  title: "Block/HomePageParagraph",
```
richardxia: Apologies for not being clearer about this in the Clubhouse ticket, but this component should be a Grid-Level component and make use of the grid system stuff in https://github.com/ShelterTechSF/sheltertech.org/pull/66. Also please move these files into the `src/components/grid-aware` directory that I started created in #66.

Pretty much all of the remaining tickets that we have now will be grid-level components, since they should each create a `display: grid` component, define the grid lines, and place the child items onto the grid. In this component, even if you are basically creating one element that is as wide as the whole inner grid, this will ensure that the width of the element is in alignment with all other elements of the page which are aligned to the grid.

richardxia (changes requested):

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
...
+ * backgrid div spans the left margin area and the center area while the inner
+ * grid content only spans the center area.
+ *
+ * The inner grid is set up with our standard 12-column grid, where elements are
+ * placed along the grid, but the image elements are relatively positioned to
+ * break out of the grid.
+ *
+ * https://en.wikipedia.org/wiki/Bleed_(printing)
+ * https://mastery.games/post/overlapping-grid-items/
+ */
```
davidagustin: This is incongruent with the file name. This comment appears to be inspired from Richard's commit so maybe it can be reworded.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
+.text {
+  font-size: 50px;
+  font-style: var(--font-family-merriweather);
+  font-weight: 300;
+  line-height: 130%;
+  text-align: center;
+  width: 1060px;
+}
```
davidagustin: Should only use font and its respective variable

davidagustin (changes requested):

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,42 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./HomePageParagraph.module.css";
+
+const HomePageParagraphPropType = PropTypes.shape({
```
davidagustin: Don't know if shape should be declared if there is one property and I wouldn't think that the shape would change anytime soon

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,42 @@
...
+        <section className={s.gridParent}>
+          <GridAreaMiddle />
+        </section>
+      </div>
+    </div>
+  );
+};
+
+HomePageParagraph.propTypes = {
+  text: HomePageParagraphPropType.isRequired,
```
richardxia: It looks like `HomePageParagraphPropType` only has a single prop defined on it, which is itself just a property named `text`. You can cut out the extra layers here by just having this component directly take in the string `text` rather than having to go through one extra layer of indirection.

I'm guessing that you probably got this from my Three Paragraph Block component in https://github.com/ShelterTechSF/sheltertech.org/pull/66. I'd say that the difference between your component and that component is that yours is much simpler, which means that the extra indirection is less helpful.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,42 @@
...
+import s from "./HomePageParagraph.module.css";
+
+const HomePageParagraphPropType = PropTypes.shape({
+  text: PropTypes.string.isRequired,
+});
+
+const HomePageParagraphBlock = ({ text }) => (
+  <div>
+    <p className={s.text}>{text}</p>
+  </div>
```
richardxia: The extra `<div>` wrapper here is not really serving any value because `<p>` tags are already block-level components, and the `<div>` is not getting styled.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: I'll add on top of @davidagustin comment here: You should also not be setting the width here, especially not to a hardcoded value. You should let this component expand to fit the space being allocated to it by its parent, since this would be an example of a [space-filling block component](https://github.com/ShelterTechSF/sheltertech.org/blob/7ed7d00bb64d8039d72760cb5ac16ed6405242e7/docs/component-design-system.md#space-filling-block-components).

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
...
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  background: var(--color-gray-300);
```
richardxia: I believe the actual design calls for a white background, and you can probably just leave this as a transparent background since the site itself has a white background.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
...
+
+.gridAreaRight {
+  grid-column: 9 / span 3;
+  grid-row: 1;
+}
+
+.gridAreaBottom {
+  grid-column: 5 / span 8;
+  grid-row: 2;
+}
```
davidagustin: There should be one grid area for the component but since the sub-grid is should be taken out, this should be taken out also.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
...
+
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: repeat(2, auto);
```
richardxia: Your component here only has one row, so you don't need to create two rows. The Three Paragraph Block that I created actually had two rows, which is why I created two rows here, but in your case, you only have one (and therefore, don't even need to define this property, since it defaults to 1).

davidagustin (changes requested):

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,220 @@
...
+/* Override grid item default of auto so that item contents cannot stretch the grid itself */
+.gridParent > * {
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .gridParent {
+    display: block;
+    padding-bottom: 80px;
+    padding-top: 40px;
```
richardxia: I know we don't have any mobile designs for this component, but I feel that having symmetric top and bottom padding would probably look better here.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: Given that you aren't using most of these, you should remove them from this component. Although it's fine to copy and paste code from elsewhere to get started, I think it's important to distill things down to the minimum set of things that you actually need for your feature.

richardxia (changes requested):

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
ivanleeswe: 👍 Makes sense. Deleted.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
ivanleeswe: 👍 Changed the proptype to string.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
ivanleeswe: 👍 You're correct. Deleted the <div> wrapper.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: 👍 I just deleted all the comments. Thanks!

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: So, if I don't use .bleedBackground at all, will the background be transparent as default?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: I just changed padding-top to 80px, but where can I see how this component looks on the mobile?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: So I kept .gridAreaMiddle and removed the rest since the text is centered. Is that ok?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: Ok, I made some changes to the CSS file. Will you please take a look?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: I think this comment wasn't properly addressed. It looks like the current revision of this class still sets a hardcode font size as well as an invalid usage of the [`font-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style) property. `font-style`, by the way, is meant to select between normal, italic, or oblique (a special type of italics) styles. Here you have put a `font-family` value in the `font-style`.

Instead of all of this, can you use the [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) shorthand property, which combines many of the font properties into a single property, and use one of the [variables we have defined](https://github.com/ShelterTechSF/sheltertech.org/blob/23d20a80c384151d4ee92601b37496f41e51085b/src/stylesheets/global.css#L54-L62) which set all these values to match one of the typographies from the [style guide](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=47%3A75)? This helps us keep the site typography consistent, as there are only about 9 combinations of font family, size, and weight in use.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,32 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./HomePageParagraph.module.css";
+
+const HomePageParagraphBlock = ({ text }) => <p className={s.text}>{text}</p>;
```
richardxia: Since this is so short and only used in one place, I think it'd be cleaner to just inline this into the main component rather than defining another component for it.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,32 @@
...
+HomePageParagraphBlock.propTypes = {
+  text: PropTypes.string.isRequired,
+};
+
+const HomePageParagraph = ({ text }) => {
+  const GridAreaMiddle = () => (
+    <div className={s.gridAreaMiddle}>
+      <HomePageParagraphBlock text={text} />
+    </div>
+  );
```
richardxia: I think this can also be inlined into the main return expression because this is a very small and simple component, and the extra layers are not necessarily useful. Also, the "middle" designation here may not really make sense here because there are no other grid areas to distinguish from.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,43 @@
+.text {
+  font-size: 50px;
+  font-style: var(--font-family-merriweather);
+  text-align: center;
+}
```
richardxia: Do you think you could define this lower in this file? As a general organizational principle, I think it is useful to group this in some meaningful and consistent way. For example, ordering it from outermost element to innermost element, or vice versa, would make it easier to find an element.

Here, you are doing a mix of the two, which makes it more difficult to locate a style. The `.text` component is the innermost element, and it appears first in the stylesheet. Then the `.bleedWrapper` appears, which is the outermost element. Lastly, you have the `.gridParent`, which is in the middle of the hierarchy.

If you don't have a strong preference, then I think I would like for all our CSS files to generally go from outermost to innermost, since as one tries to read the CSS, you often need to know about the outer elements' properties before you can understand the inner ones, such as `display: flex` being set on a parent and `align-items: center` being set in a child.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,43 @@
...
+
+.bleedWrapper {
+  display: grid;
+  grid-template-rows: auto 0;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
```
richardxia: You're missing a `grid-template-columns` definition on the non-mobile version of this element, and I think this is causing the component to render very strangely in Storybook. This is what I see in a desktop view (it's off-center):

Screen Shot 2020-10-18 at 7 35 16 PM

And this is what I see on mobile:

Screen Shot 2020-10-18 at 7 36 55 PM

`grid-template-columns` is what actually determines the number of columns here, and I think you will still want to set it for this component both in mobile and desktop views.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: Yes, if you don't set it, the background will be transparent by default.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: Storybook has a little button that lets you see what it looks like in different viewport sizes. It's the rightmost button on the top middle-left area.

Also, you can use your browser's developer toolbar to try out different sizes.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,34 @@
+.text {
+  font: var(--font-headline);
+  text-align: center;
+}
```
richardxia: Let me repeat a previous comment I had made, which I think probably got lost in the shuffle:

Do you think you could define this lower in this file? As a general organizational principle, I think it is useful to organize CSS selectors in a way that relates to the organization of the corresponding HTML. For example, ordering selectors from outermost element to innermost element, or vice versa, would make it easier to find an element in the stylesheet.

Here, you are doing a mixture of the two orderings, which makes it more difficult to locate a style. The `.text` component is the innermost element, and it appears first in the stylesheet. Then the `.bleedWrapper` appears, which is the outermost element. Lastly, you have the `.gridAreaMiddle`, which is in the middle of the hierarchy.

If you don't have a strong preference, then I think I would like for all our CSS files to generally go from outermost to innermost, since as one tries to read the CSS, you often need to know about the outer elements' properties before you can understand the inner ones, such as `display: flex` being set on a parent and `align-items: center` being set in a child.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,34 @@
...
+
+.bleedWrapper {
+  display: grid;
+  align-items: center;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
```
richardxia: Since you only have one grid row, I don't think it's necessary to define two grid rows, and you can even omit this property altogether because the default behavior is to have one auto-sized row.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,34 @@
...
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr 40px;
+  }
+}
+
+/* Main grid areas */
+
+.gridAreaMiddle {
+  grid-column: 2;
```
richardxia: I don't think property is doing anything, since this element's parent does not have `display: grid` set. This probably also means that you can just delete this selector and the component associated with it.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,27 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./HomePageParagraph.module.css";
+
+
+const HomePageParagraph = ({ text }) => {
+  const GridAreaMiddle = () => (
+    <div className={s.gridAreaMiddle}>
```
richardxia: See my other comment in the CSS file about removing this selector, but if you get rid if this `<div>` element, then you can get rid of this `GridAreaMiddle` subcomponent and just inline the `.text` element into the main body of this whole component.

richardxia (commented):
This is looking pretty good, and you've managed to slim this down considerably! In addition to my last few comments, you're currently failing lint, so you'll want to address that as well.

davidagustin (approved):

richardxia:
By the way, since https://github.com/ShelterTechSF/sheltertech.org/pull/76 merged in, I think it would be a good idea for you to rebase this branch on top of that, since that will guarantee that we run the basic CI checks on all new stories that are added.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: 👍 Thanks for the organization tip, Richard. Makes a lot of sense! Moved .text to the bottom of CSS file because it is an innermost element.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: Deleted  "grid-template-rows: auto 0." That is the property you are talking about above correct?

Lines 7-9 are grid tracks defining the column which builds the bleed columns, so I keep those right?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
ivanleeswe: 👍 Deleted gridAreaMiddle and <div>.

richardxia:
Sorry for the last-minute change, but I'm seeing that in the [latest designs](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1602), this component now has a separate title text from body text. Do you think you could update your component to match? Hopefully it shouldn't be that much more work, but I think you'll probably want to have two separate props for it now, probably a `title` and a `description`.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
```diff
@@ -0,0 +1,27 @@
...
+import React from "react";
+
+import s from "./HomePageParagraph.module.css";
+
+const HomePageParagraph = ({ title, description }) => {
+  const GridAreaMiddle = () => 
+    <div>
+      <p className={s.title}>{title}</p>
+      <p className={s.description}>{description}</p>
+    </div>
```
richardxia: Can you inline this into the `return` statement below? This is 1) simple enough that I think we can just inline it and 2) the name `GridAreaMiddle` doesn't even accurately describe this anymore, since there's no inner grid.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,35 @@
...
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr 40px;
+  }
+}
+
+.title {
```
richardxia: This should also be center-aligned to match the design. It's hard to notice in your example, since the example is wide enough to almost exactly fit the width, but if you delete some of the text in the example, you can see that it is left-aligned.

Also, you can set the `text-align` on a higher up element, such as the `<div>` that you have that currently lacks a class name, since it'll apply to all descendants.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
```diff
@@ -0,0 +1,35 @@
...
+  font: var(--font-headline);
+  margin-bottom: 40px;
+}
+
+
+.description {
+  font: var(--font-title);
+  text-align: center;
+  width: 1020px;
```
richardxia: Please remove this line, since this component should be receiving its width automatically from its parent. This is an example of a space-filling block component, per my taxonomy here: https://github.com/ShelterTechSF/sheltertech.org/blob/559e1cf695ed06c1d3d6f38e14b23d4cd6e5f102/docs/component-design-system.md#space-filling-block-components

An example of a problem that this causes is that in a mobile design, it causes the width to stretch beyond the width of the screen:

Screen Shot 2020-10-25 at 6 27 14 PM

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.js
ivanleeswe: 👍 Makes sense. Deleted GridAreaMiddle const and moved the two "p" elements down to the return statement below.

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: Instead, can I delete that empty div and set text-align to the bleedMainContent property?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
ivanleeswe: Sorry, I didn't take the mobile format into consideration. If I took the width line out, what is another way to format it so it looks like how it does in Figma?

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: Yeah, I think you can do that, which I see you've already done.

richardxia (approved):

src/components/grid-aware/HomePageLargeParagraph/HomePageParagraph.module.css
richardxia: Ah, I see. Maybe then you should set the grid columns for this component such that it spans columns 2 through 11, inclusive?

richardxia (commented):
Cool, the width looks right to me.

I'm noticing that in the mobile view, however, you have access padding/margins. You have three separate elements that add spacing to the top of the component:

### `.bleedWrapper`
Screen Shot 2020-10-26 at 10 32 27 AM

### `.gridParent`
Screen Shot 2020-10-26 at 10 32 37 AM

### The native `<p>` elements that make up the title and the description
Screen Shot 2020-10-26 at 10 32 43 AM

---

In fact, even in the desktop view, you have too much spacing on top. According to the design mockup, the entire component should have exactly 120px of spacing between the title text and the top edge of the component. Can you fix up the spacing to match the design?

Although the background is white, this extra space will be very apparent when we actually place this component next to others. In Storybook, the top edge of your component is flush with the edge of the Storybook controls (where the "Canvas", "Docs", Zoom In, etc. buttons are).

You may need to override some of the default styles on tag elements, such as the `p` element's native margin, in order to have this look correct.

ivanleeswe:
image
image
image

davidagustin:
I pair programmed with @ivanleeswe and told him to rebase his component to main so that he could add his component to the new homepage.

davidagustin (commented):
Screen Shot 2020-11-02 at 8 25 28 PM

The spacing of the two text items needs spacing in the new layout page

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.js
```diff
@@ -0,0 +1,26 @@
...
+
+import s from "./HomePageLargeParagraph.module.css";
+
+const HomePageLargeParagraph = ({ title, description }) => {
+  return (
+    <div className={s.bleedWrapper}>
+      <div className={s.bleedMainContent}>
+        <div className={s.gridParent}>
+          <div className={s.textWrapper}>
+            <p className={s.title}>{title}</p>
```
davidagustin: p tags should turn into div elements. There may be inherent properties in p that may need to be resetted. If this is implemented for semantic reasons, a p element should not have a title-like behavior.

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
```diff
@@ -0,0 +1,65 @@
...
+.description {
+  font: var(--font-title);
+  margin-top: 40px;
+
+}
+
+div p {
+  margin-top: 0;
+  padding: 0;
+}
```
davidagustin: If p tags were deleted above, this shouldn't be here.

davidagustin (changes requested):

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.js
ivanleeswe: 👍 Changed p tag to div.

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
ivanleeswe: 👍 Deleted.

davidagustin (approved):

davidagustin:
@richardxia to hit two birds with one stone I helped Ivan get his component in the new page. I'm requesting another review, thankyou.

richardxia:
> @richardxia to hit two birds with one stone I helped Ivan get his component in the new page. I'm requesting another review, thankyou.

I think the commit history got messed up, since I see 35,000 lines of changes, and the diff view on GitHub shows a lot of completely unrelated changes that are already in `main`. Could you clean it up before I take another look, since it's going to be very challenging for me to even find the files that are relevant to this PR?

davidagustin:
@ivanleeswe I fixed the branch so that only the files associated with the ticket shows. @richardxia , commit ready for review

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
```diff
@@ -0,0 +1,59 @@
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  justify-items: center;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+  padding-bottom: 120px;
+  padding-top: 120px;
```
richardxia, line 8: In the mobile design, these should both be 80px. Could you override these in the media query for this selector to match the design?

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
```diff
@@ -0,0 +1,59 @@
...
+  grid-column: 2;
+  grid-row: 1;
+  /* Override grid item default of auto so that item contents cannot stretch the grid itself */
+  min-width: 0;
+  text-align: center;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr 40px;
```
richardxia: In the designs, it looks like it actually has 20px of left and right spacing. Could you update this?

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+@media (--tablet-and-down) {
+  .gridParent {
+    display: block;
+  }
+
+  .description {
+    margin-top: 40px;
+  }
```
richardxia: It's a bit weird to have the media-query version of this selector defined higher than the non-media-query version. Everywhere else in our codebase, we always define the default styles first and then the media query overrides later.

Actually, now that I look at it more closely, this is redundant because there should be a 40px top margin in both the mobile and desktop views, so there's no need to override this in a media query.

richardxia (commented):
Just a few small comments about the mobile view, but this otherwise almost looks ready to merge in.

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
ivanleeswe: 👍 Good to know that the media-query version should be lower than the non-media query version. 

Deleted the media query for description because as you said, the 40px top margin applies to both mobile and desktop views without the need to override with a media query.

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
ivanleeswe: 👍 Added padding-bottom and padding-top with 80px to bleedWrapper media query.

src/components/grid-aware/HomePageLargeParagraph/HomePageLargeParagraph.module.css
ivanleeswe: 👍 You're right. Changed from 40px to 20px.

davidagustin (approved):
Reviewed to check if final changes were made

richardxia (approved):
LGTM! Thanks for working on this!

ivanleeswe:
Thank you, everyone! Let's keep it going!

PR #72, opened 2020-10-17
1658 create home page hero component

davidagustin:
Desktop Storybook Image:
Screen Shot 2020-10-17 at 12 33 43 AM

From: https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=154%3A572

More discussion is needed for the visual hierarchy of the mobile view. How should we order the items for mobile view? Should the Video be first or the Text? @derekfidler needs to give input and design for the mobile view.

Should the "X" In the title be dynamic?


src/components/grid-aware/Hero/Hero.module.css
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding-bottom: 150px;
+  padding-top: 150px;
```
candywang: If left and right padding are 0 by default, would it be better to use the short-hand `padding` property such as `padding: [top] [right-left] [bottom];` or in this case `padding: 150px 0 150px;`?

src/components/grid-aware/Hero/Hero.module.css
```diff
@@ -0,0 +1,76 @@
...
+  grid-column: 9 / span 1;
+  text-align: center;
+}
+
+.bleedMainContent {
+  grid-column: 2;
+  grid-row: 1;
+  /* Override grid item default of auto so that item contents cannot stretch the grid itself */
+  min-width: 0;
+}
```
richardxia: With regards to the organization of this CSS file, could you move this selector above all the main grid stuff to just below the `.bleedWrapper`? It's easier to follow when things related to each other are grouped together, and this is more related to the "Bleed Wrapper" than the main grid.

src/components/grid-aware/Hero/Hero.module.css
richardxia: I don't have a strong opinion about using the shorthand properties or not when it's just two sides being set, but if you do go with the shorthand here, please use the two-item shorthand, `padding: [top-bottom] [right-left]`, since the top and bottom padding use the same values here.

src/components/grid-aware/Hero/Hero.module.css
```diff
@@ -0,0 +1,76 @@
...
+
+.gridAreaLeft {
+  align-content: center;
+  grid-column: 1 / span 6;
+}
+
+.gridAreaRight {
+  align-self: center;
+  grid-column: 9 / span 1;
```
richardxia: In the design mockup, it looks like this play button is centered between the 9th and 10th columns, so you may want this to span two columns.

src/components/grid-aware/Hero/Hero.js
```diff
@@ -0,0 +1,93 @@
...
+      </div>
+    </div>
+  );
+};
+
+Hero.propTypes = {
+  ctaButtons: PropTypes.arrayOf(CTAButtonPropType).isRequired,
+  heroImage: PropTypes.string.isRequired,
+  heroLink: PropTypes.string.isRequired,
+  heroText: PropTypes.arrayOf(HeroTextPropType).isRequired,
```
richardxia: It's a bit redundant to prefix each of these properties with "hero" when the component name itself is "Hero", so I'd probably just drop the "hero" part of the name in each of these.

Why is `heroText` declared as an array? You don't seem to be treating it like an array in the code.

Lastly, I would suggest flattening the structure here a bit and making the sub-properties of `HeroTextPropType` direct properties of this top-level component. The title and the description are the most prominent parts of this component, and there will always be exactly one instance of them, so there's not much benefit to creating a separate, nested structure like this.

src/components/grid-aware/Hero/Hero.js
```diff
@@ -0,0 +1,93 @@
...
+);
+
+HeroText.propTypes = {
+  text: PropTypes.arrayOf(HeroTextPropType).isRequired,
+};
+
+const PlayButton = ({ link }) => (
+  <div className={s.playButtonContainer}>
+    <a rel="noreferrer" href={link} target="_blank">
+      <img src={Shape} alt="Play Button" className={s.playButton} />
```
richardxia: The alt text here should probably describe the actual action initiated by clicking this link, not the image itself. The alt text is important for visually impaired users, who may rely on screen readers in order to browse the web, and a description like "Play Button," describing the image itself, is much less useful than something like "Play video," which describes what will happen when you click on this link.

We may also want to make this alt text a prop, since it is functionally similar to how a link comprises both a URL and the hyperlinked text.

src/components/grid-aware/Hero/Hero.js
```diff
@@ -0,0 +1,93 @@
...
+  const GridAreaRight = () => (
+    <div className={s.gridAreaRight}>
+      <PlayButton link={heroLink} />
+    </div>
+  );
+
+  return (
+    <div
+      className={s.bleedWrapper}
+      style={{ backgroundImage: `url(${heroImage})` }}
```
richardxia: In the actual design, there is a semitransparent gradient that darkens the image below the text so that text is more visible over the background. Could you figure out a way to add it in?

src/components/grid-aware/Hero/Hero.js
richardxia: Also, the design calls for a white play button icon, while you have a black one. You can use CSS to change the color of a monochrome SVG image, although you may not be able to do it with the `` tag.

src/components/grid-aware/Hero/Hero.js
```diff
@@ -0,0 +1,93 @@
...
+
+/* Main component */
+
+const Hero = ({ ctaButtons, heroImage, heroLink, heroText }) => {
+  const GridAreaLeft = () => (
+    <div className={s.gridAreaLeft}>
+      <HeroText text={heroText} />
+      <CTAButtons buttons={ctaButtons} />
+    </div>
+  );
```
richardxia: I think this is probably short enough that you could just inline this directly into the main `return` statement below. In https://github.com/ShelterTechSF/sheltertech.org/pull/66, I had split it out because that component was large and complex enough that I felt that splitting it up made it easier to follow. This component has a lot less stuff, so I think it suffices to keep it simple.

src/components/grid-aware/Hero/Hero.module.css
```diff
@@ -0,0 +1,76 @@
...
+.title {
+  font: var(--font-title);
+  margin: 0;
+}
+
+/* Description block */
+
+.description {
+  font: var(--font-body-medium);
+  margin-top: 25px;
```
richardxia: In the design mockup, this looks like it's 20px, not 25px.

src/components/grid-aware/Hero/Hero.module.css
```diff
@@ -0,0 +1,76 @@
...
+.ctaButtonBlock {
+  margin-top: 40px;
+}
+
+/* Play Button block */
+
+.playButton {
+  height: 45px;
+  width: 35px;
+}
```
richardxia: These values seem awfully coupled to the image itself, and I wonder if it would be better to set the image as a background image here, rather than as an `` tag.

src/components/grid-aware/Hero/Shape.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="20" viewBox="0 0 16 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
richardxia: Can you rename this button to something like `play-button.svg`. `Shape.svg` is not very descriptive.

richardxia (changes requested):
I have some feedback, but overall this looks pretty good to me!

davidagustin:
The figma version is wrong. This should look at MVP.

richardxia (commented):
Oh, one more thing I just noticed: Could you rename this component to HomePageHero? There's a different-looking Hero component on the Volunteer page, so we may want to keep these separate until it's clearer how well these hero banners will look on each page.

richardxia:
Ah, it looks like @derekfidler has given this component a name, which is now the ["Video Header"](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14), and it's now in the Design System Figma. Do you think you can rename this component to that?

davidagustin:
> Ah, it looks like @derekfidler has given this component a name, which is now the ["Video Header"](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14), and it's now in the Design System Figma. Do you think you can rename this component to that?

Got it

src/components/grid-aware/Hero/Hero.module.css
davidagustin: I changed the img tag to an svg to fill the color from black to white. I manually set inside the SVG  width="35" height="45"

src/components/grid-aware/Hero/Hero.js
davidagustin: I changed to svg so I will also remove the alt property of the SVG

src/components/grid-aware/Hero/Shape.svg
davidagustin: got it

src/components/grid-aware/Hero/Hero.js
davidagustin: img converted to svg

src/components/grid-aware/Hero/Hero.js
davidagustin: Okay

src/components/grid-aware/Hero/Hero.js
davidagustin: > Also, the design calls for a white play button icon, while you have a black one. You can use CSS to change the color of a monochrome SVG image, although you may not be able to do it with the `` tag.

I manually changed the SVG code to have a white color.

davidagustin:
Screen Shot 2020-11-01 at 11 35 51 PM
Screen Shot 2020-11-01 at 11 35 40 PM

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,96 @@
...
+import Button from "../../inline/Button";
+import playIcon from "./PlayIcon.svg";
+import s from "./VideoHeader.module.css";
+
+/* PropType shapes */
+
+const VideoHeaderPropType = PropTypes.shape({
+  title: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+});
```
richardxia: This only seems to be used in one place. Is this really worth extracting into a separate, named constant?

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/* Main grid */
+
+.bleedMainContent {
+  grid-column: 2;
+  grid-row: 1;
+  /* Override grid item default of auto so that item contents cannot stretch the grid itself */
+  min-width: 0;
+}
```
richardxia: Should this be moved higher up to where the other class names that start with `bleed*` are? Not only are they named similarly, but they are functionally cohesive, since they are all directly related to the outer grid. If you moved this (and its associated media query) higher up, then everything at the top of this file is related to the outer grid and everything related to the outer grid is at the top.

On a lesser, but related, note, having this right below the comment heading "Main grid" is misleading because the `grid-*` properties of this selector are not related to the main grid, but instead the outer grid.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+
+  .ctaButtonRowItem {
+    margin-top: 20px;
+  }
+
+  .playIcon {
+    height: 33px;
+    width: 25.67px;
+  }
+}
```
richardxia: Grouping together the CTA buttons and the play button is a little bit confusing, since aside from the fact that they both act as "buttons", the CTA buttons have a totally different function, location, and style than the play button. The media query versions of the CTA buttons is much more related to the non-media-query version of the CTA buttons, yet you've separated them in the code with the `.playIcon` selector.

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,96 @@
...
+      </div>
+    ))}
+  </div>
+);
+
+CTAButtons.propTypes = {
+  buttons: PropTypes.arrayOf(Button.propTypes).isRequired,
+};
+
+const VideoHeaderText = ({ text }) => (
```
richardxia: I think it's not very intuitive to have a solo prop named `text` here that is actually an object with two props in it. Unless if there's some benefit to having an object here, I think it's much simpler and clearer to just flatten this into the two props that are actually there. That is, just have this component directly take a `title` prop and a `description` prop.

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,96 @@
...
+  </div>
+);
+
+VideoHeaderText.propTypes = {
+  text: VideoHeaderPropType.isRequired,
+};
+
+const PlayButton = ({ link }) => (
+  <a rel="noreferrer" href={link} target="_blank">
+    <img src={playIcon} alt="Play Video" className={s.playIcon} />
```
richardxia: Can you put the `className` prop first to much all of our other code?

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,96 @@
...
+};
+
+/* Main component */
+
+const VideoHeader = ({ ctaButtons, image, playButtonLink, text }) => {
+  const GridPlayButton = () => (
+    <div className={s.gridPlayButton}>
+      <PlayButton link={playButtonLink} />
+    </div>
+  );
```
richardxia: I don't think having this defined separately is buying us much. There's only two elements/components here, and it's just three lines long. It also makes it asymmetric with respect to `gridTextArea`, which this is siblings with. Why is `gridPlayButton` special enough to warrant defining separately by `gridTextArea` is not?

I think you should be consistent by either defining both of them outside of the main `return` statement or just inlining both of them, but what you have is halfway in between.

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,96 @@
...
+};
+
+VideoHeader.propTypes = {
+  ctaButtons: PropTypes.arrayOf(Button.propTypes).isRequired,
+  image: PropTypes.string.isRequired,
+  playButtonLink: PropTypes.string.isRequired,
+  text: PropTypes.shape({
+    title: PropTypes.string,
+    description: PropTypes.string,
+  }).isRequired,
```
richardxia: Is this really worth adding an extra level of indirection? The most prominent part of whole `VideoHeader` component is the title text, yet it is being relegated to a second layer of keys here. I'd suggest just flattening the props so that the `VideoHeader` just has `title` and `description` keys.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  background-image: linear-gradient(90deg, #000 0%, rgba(0, 0, 0, 0) 100%), var(--background-image);
+  background-size: cover;
+  grid-column: 1/ span 3;
```
richardxia: Nit: Please put a space before the slash.

```suggestion
  grid-column: 1 / span 3;
```

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+  padding: 150px 0;
+}
+
+.gridTextArea {
+  align-content: center;
+  grid-column: 1 / span 6;
+  grid-row: 1;
+}
+
+.gridPlayButton {
```
richardxia: Nit: Could you rename this to `gridPlayButtonArea` for consistency with its sibling? Also because `gridPlayButton` makes it sound like it's the actual button, when this is in fact the grid area that contains the actual button.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+/* Title and Description container */
+
+.textContainer {
+  align-content: center;
+  color: var(--color-white);
+  display: flex;
+  flex-direction: column;
```
richardxia: These properties don't do anything. Flex-box in the column direction is only interesting when you constrain the vertical height of the flex parent. You can delete these.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+
+  .gridPlayButton {
+    grid-column: 6 / span 2;
+  }
+}
+
+/* Title and Description container */
+
+.textContainer {
+  align-content: center;
```
richardxia: This doesn't do anything because this flexbox doesn't wrap, so the children are naturally aligned already.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+  color: var(--color-white);
+  display: flex;
+  flex-direction: column;
+}
+
+/* Title block */
+
+.title {
+  font: var(--font-title);
+  margin: 0;
```
richardxia: What is this explicit 0 margin for?

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+.description {
+  font: var(--font-body-medium);
+  margin-top: 20px;
+}
+
+/* Call To Action Buttons block */
+
+.ctaButtonBlock {
+  margin-top: 40px;
+}
```
richardxia: This is an unused class name (duplicate of `ctaButtonRow`?).

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+  grid-column: 1/ span 3;
+  grid-row: 1;
+  /* Override grid item default of auto so that item contents cannot stretch the grid itself */
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    background-color: var(--color-black);
+    display: grid;
```
richardxia: This is already `display: grid` in the non-media-query version of the selector, so this override doesn't do anything.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+    grid-template-columns: 40px 1fr 40px;
+    grid-template-rows: 214px auto;
+    max-width: var(--grid-bleed-max-width);
```
richardxia: This is identical to the non-media-query version of this selector, so this override doesn't do anything.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  background-image: linear-gradient(90deg, #000 0%, rgba(0, 0, 0, 0) 100%), var(--background-image);
+  background-size: cover;
```
richardxia: Actually, something I just thought of is that you may want to center the image.

You may want to replace this with:

```suggestion
  background:
    linear-gradient(90deg, #000 0%, rgba(0, 0, 0, 0) 100%),
    center / cover var(--background-image);
```

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,145 @@
...
+.gridPlayButton {
+  align-self: center;
+  grid-column: 9 / span 2;
+  grid-row: 1;
+  text-align: center;
+}
+
+@media (--tablet-and-down) {
+  .bleedMainContent {
+    grid-row: 1 /span 2;
```
richardxia: Nit: Please put a space after the `/`.

```suggestion
    grid-row: 1 / span 2;
```

src/pages/new/index.js
```diff
@@ -13,10 +13,26 @@ import ThreeParagraphBlock from "../../components/grid-aware/ThreeParagraphBlock
...
+      text={{
+        title:
+          "X in 10 people experiencing homelessness lack access to the internet and essential digital services",
+        description:
+          "ShelterTech is a nonprofit organization dedicated to supporting people who are experiencing homelessness or housing insecurity by leveraging technology and connectivity.",
+      }}
+      image={videoHeaderImage}
+      ctaButtons={[
+        { text: "Donate", internalLink: "/foo" },
+        { text: "Volunteer", internalLink: "/bar" },
```
richardxia: Could you set these to the actual URLs of these pages? I created them in https://github.com/ShelterTechSF/sheltertech.org/pull/80, but it should be fine to reference those URLs even before that PR merges.

src/pages/new/index.js
```diff
@@ -13,10 +13,26 @@ import ThreeParagraphBlock from "../../components/grid-aware/ThreeParagraphBlock
...
 import Layout from "../../components/layout";
 
 export default () => (
   <Layout>
+    <VideoHeader
+      text={{
+        title:
+          "X in 10 people experiencing homelessness lack access to the internet and essential digital services",
+        description:
+          "ShelterTech is a nonprofit organization dedicated to supporting people who are experiencing homelessness or housing insecurity by leveraging technology and connectivity.",
```
richardxia: Can you have this text match the latest version in the mockups?

richardxia (commented):
The actual rendered component in the browser looks good to me, so most of my comments are about code organization, naming, and some redundant CSS properties. This is getting pretty close!

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,138 @@
...
+/* Call To Action Buttons block */
+
+.ctaButtonRow {
+  display: flex;
+  flex-flow: row wrap;
+  margin-top: 40px;
+}
+
+.ctaButtonRowItem:not(:last-child) {
+  margin-right: 24px;
```
richardxia: I think this is a typo, where you flipped the digits? According to the designs, this should be 42px, not 24px.

src/components/grid-aware/VideoHeader/VideoHeader.js
```diff
@@ -0,0 +1,90 @@
...
+    ))}
+  </div>
+);
+
+CTAButtons.propTypes = {
+  buttons: PropTypes.arrayOf(Button.propTypes).isRequired,
+};
+
+const VideoHeaderText = ({ title, description }) => (
+  <div className={s.textContainer}>
```
richardxia: Could you make the CSS class names more consistent with or even exactly match the React component names? I think this would make the code easier to follow, since you'd only have one set of names to remember.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,138 @@
...
+    padding: 0;
+  }
+
+  .gridTextArea {
+    grid-column: 1 / span 12;
+    grid-row: 2;
+    margin-top: -14.5px;
+  }
+
+  .gridPlayButton {
```
richardxia: Looks like you misspelled this, which is causing the play button to not be centered in mobile.

```suggestion
  .gridPlayButtonArea {
```

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,138 @@
...
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  background:
+    linear-gradient(90deg, #000 0%, rgba(0, 0, 0, 0) 100%),
+    center / cover var(--background-image);
```
richardxia, line 14: Since this is a bit esoteric, I think it'd be a good idea to write a comment describing how this `--background-image` works. That is, we should describe that it's expected that the user of the component should set an inline style that sets this CSS custom property.

src/components/grid-aware/VideoHeader/VideoHeader.module.css
```diff
@@ -0,0 +1,138 @@
...
+/* Main grid */
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding: 150px 0;
+}
+
+.gridTextArea {
+  align-content: center;
```
richardxia: I don't think that this does anything, since this property only applies to grid/flex parents, not grid/flex children.

```suggestion
```

If you really want to use this, then move it to `.gridParent` (and also get rid of the `align-self` on `.gridPlayButtonArea`, since it would then be redundant).

richardxia (commented):
Just have a few last remaining comments left, which are just small bugs or suggestions to naming. I think this is something we'll be able to merge in pretty soon!

src/components/grid-aware/VideoHeader/VideoHeader.module.css
davidagustin: Okay, will fix this

src/components/grid-aware/VideoHeader/VideoHeader.js
davidagustin: Okay

src/components/grid-aware/VideoHeader/VideoHeader.module.css
davidagustin: Okay

src/components/grid-aware/VideoHeader/VideoHeader.module.css
davidagustin: Okay

src/components/grid-aware/VideoHeader/VideoHeader.module.css
davidagustin: Okay

richardxia (approved):
Thanks! This LGTM! I'm pretty excited to see this land on the staging site!

PR #73, opened 2020-10-17
1656 enhance button component

davidagustin:
I thought of a standard for different button behaviors:

Screen Shot 2020-10-17 at 3 08 12 AM

You must pass an object with a text property and declare one of the properties:
1. externalLink
2. internalLink
3. scriptAction

All of these properties are self explanatory but for script action, that may use code for modals, dialogues or some customizable behavior.

Please let me know and lets discuss different approaches to solving different button click behaviors.

Screen Shot 2020-10-17 at 3 08 28 AM


src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+  let buttonView;
 
-Button.propTypes = {
-  children: PropTypes.node,
+  if (button.externalLink) {
+    buttonView = (
+      <a rel="noreferrer" target="_blank" href={button.externalLink}>
+        <button type="button" className={s.button}>
+          {button.text}
+        </button>
```
richardxia: For external links, I would suggest not using the `<button>` element at all, and instead just style the `<a>` tag directly [to look like a button](https://css-tricks.com/css-basics-styling-links-like-boss/#links-as-buttons).

src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+      </a>
+    );
+  } else if (button.internalLink) {
+    buttonView = (
+      <a href={button.internalLink}>
```
richardxia: Internal links should use Gatsby's `<Link>` component, which is integrated with Gatsby's routing system to navigate to a separate page without triggering a page refresh. You'll need to figure out how to apply custom styles to a Gatsby `<Link>`.

src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+    internalLink: PropTypes.string,
+  }),
+  PropTypes.exact({
+    text: PropTypes.string,
+    scriptAction: PropTypes.func,
+  }),
+]);
+
+Button.propTypes = {
+  button: ButtonPropType.isRequired,
```
richardxia: This extra layer seems unnecessary. Why not just set `Button.propTypes = PropTypes.oneOfType(...)`? Otherwise, you have this extra layer of thing that you have to step through in order to access values.

src/components/inline/Button/Button.stories.js
```diff
@@ -7,9 +7,41 @@ export default {
...
+      display: "flex",
+      flexDirection: "row",
+      justifyContent: "space-evenly",
+      alignItems: "center",
+      marginTop: "50vh",
+    }}
+  >
+    <Button button={externalLinkButton} />
+    <Button button={internalLinkButton} />
+    <Button button={scriptActionButton} />
```
richardxia: I think you probably want to make these separate Storybook stories, since they are each an example of a different usage of the component.

src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+        onClick={() => {
+          button.scriptAction();
+        }}
+      >
+        {button.text}
+      </button>
+    );
+  }
+
+  return <div>{buttonView}</div>;
```
richardxia: You should remove this extra `<div>`, since it causes the component to become a block element, not an inline element.

Also, you may find it easier to just `return` from each of the inner conditional cases, since there's no need to assign them to a variable only to just directly return it in the end.

src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+    text: PropTypes.string,
+    externalLink: PropTypes.string,
+  }),
+  PropTypes.exact({
+    text: PropTypes.string,
+    internalLink: PropTypes.string,
+  }),
+  PropTypes.exact({
+    text: PropTypes.string,
+    scriptAction: PropTypes.func,
```
richardxia: I think it'd better to name this `onClick`, since that's the name of the equivalent HTML property, and it's a fairly common name for this type of prop in React components.

src/components/inline/Button/Button.js
```diff
@@ -3,18 +3,59 @@ import React from "react";
...
+        </button>
+      </a>
+    );
+  } else {
+    buttonView = (
+      <button
+        type="button"
+        className={s.button}
+        onClick={() => {
+          button.scriptAction();
```
richardxia: For flexibility, you may want to actually capture the JavaScript event object and pass it through to the prop, which would allow someone to handle the event in their own custom action.

richardxia (changes requested):
In addition to the feedback I left here, I'm pretty sure that this change will break some of the existing components that try using the Button component.

We should probably add a CI check to make sure that all Storybook components render and have no prop type errors, which is something I've been wanting to do for a while.

src/components/inline/Button/Button.js
davidagustin: Me and Richard Agreed on the a11y conventions and kept button.

src/components/inline/Button/Button.js
```diff
@@ -1,20 +1,58 @@
...
 
-const Button = ({ children }) => (
-  <button className={s.button} type="button">
-    {children}
-  </button>
-);
-
-Button.propTypes = {
-  children: PropTypes.node,
+const Button = ({ button }) => {
```
richardxia: Is it possible to set the props of this component such that you can pass them in as a flat list rather than in this `button` prop? For example, as a user of this component, I think I would prefer to write

```jsx
<Button text="Click Me" externalLink="https://google.com" />
```

instead of

```jsx
<Button button={{ title: "Click Me", externalLink: "https://google.com"}} />
```

richardxia (commented):
This looks almost right to me, but I have one last request for the props of the component. This might mean that what you're trying to do with the prop types might not work, but I think it's important to make this component more intuitive to people who using it.

davidagustin:
This is the updated standard for the buttons. After this commit is done, Buttons component on previous commits  need to be modified.

You must pass an object with a text property and declare one of the properties:

1. externalLink
XOR
2. internalLink
XOR
3. onClick

All of these properties are self explanatory but for script action, that may use code for modals, dialogues or some customizable behavior.

richardxia (commented):
Cool, this almost LGTM. I think the last thing you need to do is update the existing components which used the `<Button>` component to use the new API. I think the only two are the Three Paragraph Block and the Footer CTA Card, which you can easily inspect in Storybook.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -42,8 +43,8 @@ const CTABlock = ({ title, buttons }) => (
     </div>
     <div className={s.ctaButtonRow}>
       {buttons.map((button) => (
-        <div className={s.ctaButtonRowItem}>
-          <Button key={button.text}>{button.text}</Button>
+        <div className={s.ctaButtonRowItem} key={button.text}>
```
davidagustin, line 46: Screen Shot 2020-10-19 at 7 51 58 PM

I refactored the parent element of the body in the map function to make this warning go away.

PR #74, opened 2020-10-18
Scope legacy global styles

richardxia:
You'll probably want to review the diff with "ignore whitespace differences" turned on, since I had to indent almost every legacy stylesheet.

While trying to add a new page on the Gatsby site for the new site design, I was hitting issues with global styles messing with the view of the new components that we've been developing in Storybook. This PR just changes the old stylesheets so that anything that was changing global styles (i.e. selectors that were only based on tag names and not classes) is now scoped under an `.app` class. There happens to be a `<div class="app">` in the old site that was the parent of every page, so it's safe to nest every selector under a `.app` class without changing the appearance of the page.

I hit one other issue with trying to get rid of global styles, which is that the old `_typography.scss` file was setting the font size of the root element. This has another, subtle global effect, which is that it changes the value of the [`rem`](https://developer.mozilla.org/en-US/docs/Web/CSS/length#rem) unit. `rem` is like `em` except that it's always based on the font size of the root element (i.e., the `html` element) rather than the parent of the current element. Changing the font size of the parent element therefore changes the value of the `rem` unit everywhere, so I had to readjust all of the `rem` units back to what they were supposed to look like.

I think this should result in the site not changing at all, and I did manual, visual comparisons of each page in order to test it.

I'll submit a separate PR to actually introduce the new home page with our new components.

### DIFF
--- a/src/legacy/components/pages/annual-report-2018/AnnualReportView.scss
+++ b/src/legacy/components/pages/annual-report-2018/AnnualReportView.scss
@@ -1,453 +1,461 @@
 @import '../../../stylesheets/utils/all';
 
-.annual-report {
-  align-items: center;
-  display: flex;
-  flex: 0 0 auto;
-  flex-direction: column;
-  justify-content: space-between;
-}
+// Disable this because we need to temporarily add extra specificity to all
+// legacy styles in the middle of the site transition. All the legacy styles and
+// code will go away soon.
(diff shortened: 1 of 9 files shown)


davidagustin (approved):
Looks good to me. Checked side by side with the production website.

richardxia:
Thanks!

PR #75, opened 2020-10-19
Create new home page component

richardxia:
https://app.clubhouse.io/sheltertech/story/1665/create-actual-new-home-page-component

I created a new page component for the new home page under `/new`. This will allow us to easily preview the new site without taking down the old one. As we develop the components on this page, we can gradually add them to this page component. Note that although this will look very similar to the Storybook stories, the data we are setting here for each of props is what will actually go on the site. After we integrate the CMS, these page components are the ones that should be querying the CMS, but for the initial site launch, we'll stick with the content being hardcoded here.

Eventually, when we do the site launch, we can make one final PR to delete all the old components and move everything from `/new` back to the parent level.

One other thing I did here was to create a `<Layout>` component. This is a convention that [Gatsby recommends](https://www.gatsbyjs.com/docs/layout-components/) for extracting common layout across multiple pages. When we actually build out the header and footer components, they probably would go here, but for now, I'm using it as a way to always ensure that the `global.css` file gets imported.

---

Here's a screenshot of the actual thing, where you can see that this is the real deal because of the little Intercom chat bubble icon on the bottom right:

Screen Shot 2020-10-18 at 5 24 16 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,13 @@ For more specific documentation, visit one of the pages below:
 [Command-Line Quick Reference]: ./docs/command-line-quick-reference.md
 [Component Design System]: ./docs/component-design-system.md
 [How To Deploy]: ./docs/how-to-deploy.md
+
+## Staging website
+
+We're currently in the process of developing a redesign of the sheltertech.org
+website! Until we are done and ready to flip the switch to the new design, we
+will be placing the new site pages under the `/new` URL path so that they can be
+easily previewed before we roll them out to the main site page routing.
+
+To view the current state of the new design on the staging website, please visit
+https://staging.sheltertech.org/new.
--- a/src/pages/new/index.js
+++ b/src/pages/new/index.js
@@ -1,0 +1,48 @@
+import React from "react";
+
+import ThreeParagraphBlock from "../../components/grid-aware/ThreeParagraphBlock";
+import image1 from "../../components/grid-aware/ThreeParagraphBlock/stories/image1.png";
+import image2 from "../../components/grid-aware/ThreeParagraphBlock/stories/image2.png";
+import image3 from "../../components/grid-aware/ThreeParagraphBlock/stories/image3.png";
+import Layout from "../../components/layout";
+
+export default () => (
+  <Layout>
+    <ThreeParagraphBlock
+      title="Get involved"
+      paragraph1={{
+        title: "Volunteer",
+        description:
+          "Volunteers make our work possible. There are several ways to support our mission. Learn more and get involved.",
+      }}
+      paragraph2={{
+        title: "Partnerships",
+        description:
+          "We work with companies, nonprofits, and local governments to empower the community. Reach out to us.",
+      }}
+      paragraph3={{
+        title: "Donate",
+        description:
+          "Our programs are largely funded by donations from people who care about bridging the digital divide. Support ShelterTech today.",
+      }}
+      image1={{
+        url: image1,
+        alt: "Two volunteers working on a laptop together at a datathon.",
+      }}
+      image2={{
+        url: image2,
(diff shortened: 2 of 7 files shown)


src/pages/new/index.js
```diff
@@ -0,0 +1,48 @@
+import React from "react";
+
+import ThreeParagraphBlock from "../../components/grid-aware/ThreeParagraphBlock";
+import image1 from "../../components/grid-aware/ThreeParagraphBlock/stories/image1.png";
+import image2 from "../../components/grid-aware/ThreeParagraphBlock/stories/image2.png";
+import image3 from "../../components/grid-aware/ThreeParagraphBlock/stories/image3.png";
```
richardxia, line 6: I just wanted to explicitly note that I know that I am using the Storybook images here, which is not a good idea, but until we get the final images, I thought it'd be OK to reuse them here.

README.md
```diff
@@ -15,3 +15,13 @@ For more specific documentation, visit one of the pages below:
...
+
+## Stanging website
+
+We're currently in the process of developing a redesign of the sheltertech.org
+website! Until we are done and ready to flip the switch to the new design, we
+will be placing the new site pages under the `/new` URL path so that they can be
+easily previewed before we roll them out to the main site page routing.
+
+To view the current state of the new design on the staging website, please visit
+https://staging.sheltertech.org/new.
```
richardxia, line 27: Note that this URL won't work until this PR is merged into `main`, but once it does, it'll automatically be deployed to staging.

README.md
```diff
@@ -15,3 +15,13 @@ For more specific documentation, visit one of the pages below:
 [Command-Line Quick Reference]: ./docs/command-line-quick-reference.md
 [Component Design System]: ./docs/component-design-system.md
 [How To Deploy]: ./docs/how-to-deploy.md
+
+## Stanging website
```
davidagustin: Typo. Should be Staging website.

davidagustin (changes requested):

README.md
richardxia: Haha, yes, thanks for catching that.

richardxia: This should now be fixed.

davidagustin (approved):

src/pages/new/index.js
```diff
@@ -0,0 +1,48 @@
...
+      image3={{
+        url: image3,
+        alt: "Multiple volunteers working at a datathon.",
+      }}
+      ctaTitle="Volunteer, donate, or reach out to our partnerships team"
+      ctaButtons={[
+        { text: "Volunteer" },
+        { text: "Donate" },
+        { text: "Work with us" },
+      ]}
```
davidagustin: Button behaviors needed when clicked.

davidagustin (changes requested):

src/pages/new/index.js
richardxia: Yeah, I was waiting for https://github.com/ShelterTechSF/sheltertech.org/pull/73 to merge in first, since that was going to change the APIs, but I can now swap these out. I'll still have to dummy these out to some extent, since the new pages/interactions don't exist yet, but perhaps I can create placeholders for the new pages as well.

davidagustin (approved):

src/pages/new/index.js
richardxia: This is updated now. I created placeholder pages for the volunteer and donate pages in https://github.com/ShelterTechSF/sheltertech.org/pull/75/commits/ab685a04999584fd8432de58fefb7d23e94fe2de, and the first two links now link there. The third one I have just dummied out with an `onClick` that does nothing.

davidagustin (approved):
LGTM

PR #76, opened 2020-10-20
Add Storybook CI tests

richardxia:
https://app.clubhouse.io/sheltertech/story/1666/add-ci-check-for-storybook

This adds a bunch of dependencies and infrastructure to allow us to run our Storybook stories as test cases. These check that the stories render without errors, including prop-type or other React warnings.

## Detailed explanation of what I did and what I referenced while working on this
I had to cobble together several things together to make it all work. I ultimately followed the directions from https://storybook.js.org/docs/react/workflows/snapshot-testing and https://www.gatsbyjs.com/docs/unit-testing/ to install the [Jest](https://jestjs.io/) testing framework and [Storybook's StoryShots](https://www.npmjs.com/package/@storybook/addon-storyshots) add-on.

StoryShots enables snapshot testing for Storybook stories, but I didn't actually enable snapshot testing here. I don't think it's worth the overhead for tis project, given that there is basically no "logic" in our components and they're all just presentational components. Instead, I'm using the [render-only mode](https://github.com/storybookjs/storybook/blob/v6.0.26/addons/storyshots/storyshots-core/README.md#renderonly) option to just render and not generate/compare with [snapshots](https://jestjs.io/docs/en/snapshot-testing), and the main benefit we get from using this add-on is that it automatically finds and runs our Storybook stories as tests.

I had to do a bit of special work to make any `console.error()` calls fail tests, inspired by the comments in https://github.com/facebook/jest/issues/6121. The code changes for this are all in https://github.com/ShelterTechSF/sheltertech.org/commit/3e00f8a3265c8a0846e8beac5cdd685a0f173176. One thing that I don't like about this current implementation is that the stack traces are a bit confusing, since they end up pointing to the test infrastructure rather than the actual line of code that triggered the console errors, but I think this is the best that I can do for now until I have time to research how to customize the stack trace in Jest/Node.js.

## Fixing an actual error
Lastly, we had an actual `console.error()` failure in the Three Paragraph Block, which I had accidentally introduced at the last minute when trying to fix the spacing between the button components. In order to actually have these new CI tests pass, I had to fix this, which I did so in https://github.com/ShelterTechSF/sheltertech.org/commit/5efeb42804fd44a0fcb033496e53cc3172965637.

## How to use this

I added some documentation to the command line quick reference, but to repeat it here, you can now run these tests with:

```sh
npm run test
```


davidagustin (approved):
LGTM

PR #77, opened 2020-10-21
1662 create partners and sponsors block

candywang:
`PartnersAndSponsorsBlock.stories.js`
- Created stories Template component
  - Originally passed each logo down as individual props then decided to refactor `DefaultPartnersAndSponsorsBlock.args` to add a `partnersAndSponsors` property that contains an array of elements representing individual logos and passed `partnersAndSponsors` down as a prop

`PartnersAndSponsorsBlock.js`
- Used PropType shape to define individual logos
- Separated logos into a subcomponent and used JS map function
- Separated left (`title` prop) and right (`LogosBlock` subcomponent) grid area

`PartnersAndSponsorsBlock.module.css`
- Used `Main grid` and `Main grid areas` from **ThreeParagraphBlock** CSS modules

`index.js`
- Exported component

**Concerns:**
- The `key` value needs to be changed as it will cause issues down the line. I'm thinking of changing the value to a string of the organization name, any suggestions? - value of key is currently the alt text for the respective logo

**WIP:**
- Need to add --tablet-and-down media queries based on the final design


src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.js
```diff
@@ -0,0 +1,46 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./PartnersAndSponsorsBlock.module.css";
+
+const LogoPropType = PropTypes.shape({
+  key: PropTypes.number.isRequired,
```
richardxia: `key` is actually a special React prop [that is only meant to be used on array items](https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys). It's not a prop that is used by the item itself, but rather a prop used by the implicit array parent in order to track which child elements to re-render if the array contents change.

To quote the documentation that I linked to above:

> For example, if you extract a `ListItem` component, you should keep the key on the `<ListItem />` elements in the array rather than on the `<li>` element in the `ListItem` itself.

Also, the key isn't something that you necessarily need the person instantiating the top-level component to specify. It just needs to be some unique identifier amongst the array items. In think case, I think either the URL or the alt text should be unique, so you can use either.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.js
```diff
@@ -0,0 +1,46 @@
...
+          <div className={s.logoContainer} key={logo.key}>
+            <img className={s.logo} src={logo.url} alt={logo.alt} />
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+
+  return (
+    <div>
```
richardxia: This extra `<div>` isn't doing anything, so you can remove it. [`<section>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) is a semantic HTML tag that is visually equivalent to a `<div>` (i.e. is a block element and doesn't really have any other visible features), but it implies a certain type of semantic on the page.

richardxia (commented):
I'm out of time this morning to leave a more thorough review, but one other thing that I wanted to note is that you will still want to add a "bleed parent" to this component, or else the content won't be centered and with a maximum width like other "grid-aware" elements on the page.

Since my home page component PR (https://github.com/ShelterTechSF/sheltertech.org/pull/75) merged in, you can actually see this for yourself, where if you instantiated your component on the page, it won't be horizontally aligned with the Three Paragraph Block.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,63 @@
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: repeat(2, auto);
```
richardxia: It looks like you only have one row, so you don't need to declare this, which declares that there are two rows. Omitting this property altogether will just create a single, auto-sized row.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+
+.logosWrapper {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.logoContainer {
+  height: 96.08px;
+  width: 215.14px;
```
richardxia: This probably isn't what we want. If you find yourself setting very oddly specific values (especially ones with fractional pixels), it's probably not the right approach.

I wonder if what we actually want here is another grid, as this would allow you to evenly allocate space in each row. Could you try constructing a grid with automatically generated rows, like [this example](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout#The_implicit_and_explicit_grid)?

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
```diff
@@ -0,0 +1,66 @@
...
+import sfFamiliesLogo from "./stories/sf-families-logo.png";
+import uberLogo from "./stories/uber-logo.png";
+
+export default {
+  title: "Grid-Aware/PartnersAndSponsorsBlock",
+  component: PartnersAndSponsorsBlock,
+};
+
+const Template = ({ title, partnersAndSponsors }) => (
+  <div>
```
richardxia: This `<div>` isn't doing anything, so I think we can remove it.

richardxia (commented):
I got a bit more time to take a look at this, and it's looking pretty good so far. I have a suggestion to try using a CSS grid for the actual sequence of logos, since that may make the sizing more flexible and consistent.

Also, I noticed that the mobile view looks a bit off, since the images are center-aligned, but it seems to be offset. I think this might be fixed by adding the "bleed wrapper" back in, so try that and see if it also helps the mobile view.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
candywang: Yes! I agree a grid layout is a much better implementation.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,67 @@
...
+  max-width: var(--grid-bleed-max-width);
+}
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-column-start: 2;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding-bottom: 100px;
+  padding-top: 100px;
```
davidagustin: Can use shorthand 
padding: 100px 0;

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,67 @@
...
+.gridAreaRight {
+  grid-column: 5 / span 8;
+}
+
+
+/* Logos block */
+
+.logosWrapper {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
```
davidagustin: can shorthand to:
grid-template-columns: repeat(4, 1fr);

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
```diff
@@ -0,0 +1,64 @@
+import React from "react";
+
+import PartnersAndSponsorsBlock from "./PartnersAndSponsorsBlock";
+import mohcdLogo from "./stories/MOHCD-logo.png";
```
davidagustin: Optional: Lowercase MOHCD for consistency since sf is lowercase.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
```diff
@@ -0,0 +1,64 @@
...
+      url: benetechLogo,
+      alt: "Logo of Benetech nonprofit organization.",
+    },
+    {
+      url: larkinStreetLogo,
+      alt: "Logo of Larkin Street Youth Services nonprofit organization.",
+    },
+    {
+      url: sfFamiliesLogo,
+      alt: "Logo of SF Families online public service directory organization.",
```
davidagustin, line 49: I don't know what is the criteria for words that should be capitalized. Can you explain to me what it is?

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
```diff
@@ -0,0 +1,64 @@
...
+      url: ciscoLogo,
+      alt: "Logo of Cisco Systems networking hardware company.",
+    },
+    {
+      url: pagerdutyLogo,
+      alt: "Logo of PagerDuty computer software company.",
+    },
+    {
+      url: uberLogo,
+      alt: "Logo of Uber Technologies Inc company.",
```
davidagustin: Should Inc have a period in the end? Inc.

davidagustin (changes requested):

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.js
```diff
@@ -0,0 +1,45 @@
...
+const LogoPropType = PropTypes.shape({
+  url: PropTypes.string.isRequired,
+  alt: PropTypes.string.isRequired,
+});
+
+const PartnersAndSponsorsBlock = ({ title, partnersAndSponsors }) => {
+  const GridAreaLeft = () => (
+    <div className={s.gridAreaLeft}>
+      <h1 className={s.title}>{title}</h1>
+    </div>
```
davidagustin, line 15: This area appears simple enough to move into the return. Isn't that right @richardxia ?

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,67 @@
+/* Main grid */
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
```
richardxia: You can remove this line, since there is only one row in this outer grid. My ThreeParagraphBlock had to play some tricks to get some of the bleed effect working, but because you don't have that here, you can just remove this and use the default value of one, auto-sized row.

```suggestion
```

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
richardxia: In addition to making that change, I think you'll also want to change this to be a single-column grid on mobile devices, so you'll want to write a media query that overrides this and just sets a single column.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -0,0 +1,67 @@
+/* Main grid */
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
```
richardxia, line 4: You'll want to override this in a media query for smaller screen sizes, since this isn't appropriate for mobile. See https://github.com/ShelterTechSF/sheltertech.org/blob/559e1cf695ed06c1d3d6f38e14b23d4cd6e5f102/src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css#L48-L51.

richardxia (commented):
This looks pretty good to me. My last remaining comments are all about the mobile view, which I think we can touch up with only a few changes to it.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.js
richardxia: Although I would agree with that in isolation, because in this case `<GridAreaRight>` is more complex, I think it'd be better to keep this as a separate component for symmetry with `<GridAreaRight>`. I think that either we inline both of them or neither of them, and in this case, `<GridAreaRight>` looks complex enough to warrant being defined separately.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
candywang: `SF Families` is capitalized since it's the organization name, `online public service directory` was added to describe the organization's service as "SF Families" may sound a bit generic, and `organization` is the type of entity.

To answer your question, the name of any business entity, organization, etc. should be capitalized.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
candywang: I think either way works in this context but you're right, I can add the period. Also, these are just placeholders as we'll be using the data we receive from the CMS.

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.stories.js
candywang: Great observation! I've made the changes.

richardxia (commented):
One final request: I think we'll want to center-align the logo column on mobile, since the logos themselves are centered, but the column is off-center:

Screen Shot 2020-10-25 at 7 07 16 PM

I think if you fix that up, this will be good to merge!

richardxia (approved):
LGTM! :shipit:

davidagustin (approved):

PR #79, opened 2020-10-21
1661 create blockquote block component

ivanleeswe:
1. Render a block of text on the grid.

2. Style the grid.

3. Add the carousel to display multiple pages of text.


package.json
```diff
@@ -32,7 +32,9 @@
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
     "react-burger-menu": "^2.6.13",
-    "react-helmet": "^6.1.0"
+    "react-helmet": "^6.1.0",
+    "react-slick": "^0.27.12",
+    "slick-carousel": "^1.8.1"
```
richardxia: Sorry, could you find a different carousel library? It looks like slick-carousel depends on jQuery, which I don't want to pull in as a dependency into this project, since it's older technology and most of its benefits are no longer relevant due to browsers having evolved a lot over the last decade and incorporating many of jQuery's features natively.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./BlockQuoteBlock.module.css";
+
+const BlockQuoteBlock = ({ quote, author }) => {
+  const GridAreaMiddle = () => (
+    <div className={s.gridAreaMiddle}>
+      <div className={s.quote}>{quote}</div>
+      <br />
```
richardxia: You shouldn't use `<br />` tags for spacing purposes. Instead, you should just use CSS.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,28 @@
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./BlockQuoteBlock.module.css";
+
+const BlockQuoteBlock = ({ quote, author }) => {
+  const GridAreaMiddle = () => (
```
richardxia: This subcomponent is probably unnecessary to have in such a simple component, so I would just inline this directly into the main body below.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,28 @@
...
+    <div className={s.gridAreaMiddle}>
+      <div className={s.quote}>{quote}</div>
+      <br />
+      <div className={s.author}>{author}</div>
+    </div>
```
richardxia: I was looking up how to semantically write this, since this is a rare opportunity to make use of the [`<blockquote>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) tag, and I found [this article](https://css-tricks.com/quoting-in-html-quotations-citations-and-blockquotes/#hey-what-about-the-figure-element), which has a suggestion that I like. Basically, it suggests putting both the quote and the attribution in a `<figure>` element, the quote itself in a `<blockquote>` element, and the attribution in a `<figcaption>` element.

By the way, can we rename `author` to `attribution` here, since it can be more than the person's name, and also "author" to me implies someone who _wrote_ something, but "attribution" is a bit more general.

richardxia: Also, I think it'd probably be nicer to make the em dash part of the React component itself so that you don't have to place it in the `author` prop for every quote.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
```diff
@@ -0,0 +1,37 @@
...
+const settings = {
+  dots: true,
+  infinite: true,
+  speed: 500,
+  slidesToShow: 1,
+  slidesToScroll: 1,
+};
+
+const Template = ({ quote, author }) => (
+  <Slider {...settings}>
```
richardxia: This really ought to be part of our `<BlockQuoteBlock>` component itself. Otherwise, you have to instantiate your own carousel every time you want to use it. I think you should instead make the component take in an array of quote/attribution objects.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
```diff
@@ -0,0 +1,37 @@
...
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
+  </Slider>
+);
+
+export const DefaultBlockQuoteBlock = Template.bind({});
+
+DefaultBlockQuoteBlock.args = {
+  quote: `"We've heard from San Franciscans, including as part of our recent strategic planning community outreach that it's hard to know where to go when you're looking for supportive services. It's hard to know where to start, and hard to navigate different resources to find accurate information. We're glad to be supporting ShelterTech and the broader community to build SF Service Guide - a one-stop, reliable place to help connect residents to the services they need"`,
```
richardxia: Perhaps we should just make the opening quote and the closing quotes part of the React component itself so that people don't have to put the quotes in props everywhere?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
```diff
@@ -0,0 +1,25 @@
...
+import React from "react";
+import BlockQuoteBlock from "./BlockQuoteBlock";
+
+export default {
+  title: "Grid-Aware/BlockQuoteBlock",
+  component: BlockQuoteBlock,
+};
+
+const Template = ({ quote, author }) => (
+  <Carousel renderCenterLeftControls={false} renderCenterRightControls={false}>
```
richardxia: Looks like `false` is not a valid value to put here, since it's expecting a [render prop](https://reactjs.org/docs/render-props.html), that is, a function that returns a JSX component in the end. Maybe you could just try passing in an empty function?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,28 @@
...
+        <GridAreaMiddle />
+      </div>
+    </div>
+  );
+};
+
+BlockQuoteBlock.propTypes = {
+  quote: PropTypes.string.isRequired,
+  author: PropTypes.string.isRequired,
+};
```
richardxia: I think we should make this component include the actual carousel, so that this component takes in an array of quotes+attributions. That way, when you want to use it with multiple quotes, you don't have to create the carousel yourself.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
```diff
@@ -0,0 +1,25 @@
...
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
+    <BlockQuoteBlock quote={quote} author={author} />
```
richardxia: Could you make the story display a couple different quotes? Or at least alternate between two quotes so that it feels a bit more realistic?

richardxia (commented):
Left some more comments, but nuka-carousel looks like a good replacement library!

package.json
ivanleeswe: 👍 Changed to using nuka-carousel instead.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Deleted.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: But if I use the figure, blockquote, figcaption elements, do I use the CSS too? If I do, it does not really look like the design on Figma. And if I don't include the CSS, what is the point of using these three elements?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 I used margin-bottom in CSS for line separation.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: 👍 Makes sense! Added quotes to React component instead.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: 👍 Passed in empty functions and worked. Will you please explain why inserting empty functions work?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: 👍 You got it. Pretty sure the code needs some cleaning up though. Please advise.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: I added an array of objects with key quote and value attribution. Not sure if that's how you envisioned it. Please let me know.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: Ok, I took the carousel and made it part of the blockquoteblock component instead. Will you please take a look at the array I built? Is that how you want it?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
richardxia: The nuka-carousel [documentation](https://github.com/FormidableLabs/nuka-carousel#rendercontrols) said that these parameters take in render props. If you look at the link that I posted in my previous message, you'll see that a render prop is a prop that takes in a function and return a React component. I assumed that you could pass in a render prop argument that returned nothing in order to not actually render anything.

However, now that I read the nuka-carousel documentation more closely, I see that you're actually just supposed to pass in `null` if you don't want these to render:

> To change the position or remove "Paging dots", the default positions need to be disabled by setting them to null.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
richardxia: > But if I use the figure, blockquote, figcaption elements, do I use the CSS too?

Yes, you should still style the elements, since we don't want them to look like their default styles.

> If I do, it does not really look like the design on Figma.

Then that's your job to write the CSS such that it matches the designs.

> And if I don't include the CSS, what is the point of using these three elements?

You should write custom CSS for these elements, but the point of using them is to have more semantic HTML. [Here's a decent article](https://html.com/semantic-markup/) explaining what semantic HTML is and why we care about using it. To provide a brief summary here, using semantic tags allows you to more precisely describe the content of the page in HTML alone, regardless of how it is styled or rendered in a browser. These are important for navigating the HTML in non-visual ways, such as for search engine crawlers or screen readers, which are devices that allow blind or otherwise visually impaired users navigate the web.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
richardxia: Thanks for moving the carousel into the component itself, but I have a few other comments about the component now, which I'll leave as new inline comments.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,81 @@
...
+import PropTypes from "prop-types";
+import React from "react";
+import s from "./BlockQuoteBlock.module.css";
+
+const BlockQuoteBlock = ({ quote, attribution }) => {
+  const quotes = [
+    { quote: "ShelterTech, Welcomes You!", attribution: "Richard Xia" },
+    { quote: "Please Donate!", attribution: "David Agustin" },
+    { quote: "Happy Holidays", attribution: "Candy Wang" },
+  ];
```
richardxia: These quotes shouldn't live in this component. The way you have this here, it means that we can't actually set different quotes. These actual quotes should go into the Storybook story, and you should change this component to accept an array of quote + attribution pairs as a prop.

I also want to remark that you have both a prop for a single quote as well as this array of quotes, which means that you're throwing away what the user of this component has specified and rendering something totally different.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,81 @@
...
+      <div className={s.bleedWrapper}>
+        <div className={s.bleedMainContent}>
+          <div className={s.gridAreaMiddle}>
+            <figure>
+              <blockquote className={s.quote}>
+                &quot;{quotes[0].quote}&quot;
+              </blockquote>
+              <figcaption className={s.attribution}>
+                &mdash;{quotes[0].attribution}
+              </figcaption>
```
richardxia: Instead of hardcoding items for each of the three elements, you should `.map()` over the quotes. The way you have it written here, this component won't work unless if you have exactly three quotes. If you have fewer than three quotes, it'll probably throw an exception; if you have greater than three quotes, it'll just ignore the later quotes.

You should write this component that it can accept any number of quotes and properly render itself for all of them.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,81 @@
...
+            </figure>
+          </div>
+        </div>
+      </div>
+      <div className={s.bleedWrapper}>
+        <div className={s.bleedMainContent}>
+          <div className={s.gridAreaMiddle}>
+            <figure>
+              <blockquote className={s.quote}>
+                &quot;{quotes[0].quote}&quot;
```
richardxia: Could you use curly quotes instead of straight quotes? It's `&ldquo;` for &ldquo; and `&rdquo;` for &rdquo;.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,42 @@
+.bleedWrapper {
+  background: var(--color-gray-300);
```
richardxia: You can't apply this background color to this component, since it ends up in the bleed areas. In the designs, the gray background is only supposed to be as wide as the main 12 columns.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,42 @@
...
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+  padding-bottom: 120px;
+  padding-top: 120px;
```
richardxia: The designs show 60px of padding above and below, not 120px. Also, there's supposed to be 50px of spacing on the left and right sides.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,42 @@
...
+  margin: 0;
+  margin-bottom: 22px;
+}
+
+.attribution {
+  font: var(--font-caption);
+}
+  
+
+
```
richardxia: Could you avoid having empty blank lines at the end of the file?

davidagustin:
Fixed merge conflict and correctly rebased

richardxia:
> Fixed merge conflict and correctly rebased

I'm not sure what you did, but you somehow included a commit of mine, although the diff of the commit is totally different. The commit of mine is titled "Add Storybook CI tests (#76)".

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Used semantic elements figure, blockquote, and figcaption for readability and clarity.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: 👍 We talked about this and changed to pure-react-carousel instead of using nuka-carousel.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Created the blockQuoteblock component to have propTypes of an array that consists of keys "quote" and "attribution", which consists of values with propTypes of strings.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Quotes moved to default args in the storybook file.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Used map function to render quotes and attributions into each slide of the carousel.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Nice to know. Changed quotes.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Understood. The gray background is displayed behind the content with the padding now.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Changed padding top and bottom to 60px and padding left and right to 50px.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,49 @@
...
+          </Slide>
+        ))}
+      </Slider>
+      <DotGroup className={s.dots} />
+    </CarouselProvider>
+  );
+
+  return (
+    <div className={s.bleedWrapper}>
+      <div className={s.bleedBackground} />
```
richardxia: The design for this component doesn't actually have a bleed background, so you can just get rid of this element altogether as well as its associated CSS selector.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,49 @@
...
+      <div className={s.gridParent}>
+        <div className={s.gridAreaMiddle}>
```
richardxia: Given that this component isn't internally aligned to the 12-column layout, you could probably collapse these two into one element and get rid of the internal grid altogether. The outer grid is still necessary for setting up the proper maximum width, but the inner grid is unnecessary, since you're not using it.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,49 @@
...
+              <figcaption className={s.attribution}>
+                &mdash;{dataObj.attribution}
+              </figcaption>
+            </figure>
+          </Slide>
+        ))}
+      </Slider>
+      <DotGroup className={s.dots} />
+    </CarouselProvider>
+  );
```
richardxia: I think you may want to split up this component in a different way to better logically organize it. What you have here isn't bad, but I think it feels a bit odd to define an inner variable component, `<Carousel>`, that is significantly larger than the rest of the component.

There are a few different ways you could do this that I think would be improvements, but how about you try the following?

- Just inline the body of the `Carousel` component into the normal `return` statement of this `BlockQuoteBlock`
- Also just get rid of the `return` keyword, since you'd just have a single-statement component at that point. In order to do this, you also need to turn the curly braces of the function body into parentheses.
- Split out the `<figure>` and its children into its own component, which we could call `<SlideQuote>`. This component should just be defined at the top level, and it should take in two props: `quote` and `attribution`. This is a logical boundary for splitting into its own component, since each `<figure>` corresponds to a single quote, and then this would simplify the `map()` body.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,49 @@
...
+import React from "react";
+
+import "pure-react-carousel/dist/react-carousel.es.css";
+import s from "./BlockQuoteBlock.module.css";
+
+const BlockQuoteBlock = ({ quotes }) => {
+  const Carousel = () => (
+    <CarouselProvider totalSlides={quotes.length} isIntrinsicHeight>
+      <Slider>
+        {quotes.map((dataObj, index) => (
```
richardxia: Nit: But I usually like destructuring the `map()` argument into the individual properties you want to use:

```suggestion
        {quotes.map(({ quote, attribute }, index) => (
```

When you do that, you no longer have to access the properties on the `dataObject` variable, since you have individual variables for each property.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,49 @@
...
+      <div className={s.gridParent}>
+        <div className={s.gridAreaMiddle}>
+          <Carousel />
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const carouselData = PropTypes.shape({
```
richardxia: I think this name may not be the most accurate, since it describes just the data for one slide of the carousel, but not the whole carousel.

If you go with my suggestion above of breaking out the `<figure>` into its own `<SlideQuote>` component, you can just define the prop types directly on that component and then reference that prop type in the BlockQuoteBlock prop types below, similar to how we reference the `<Button>` prop types in some of our other components.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-column-start: 2;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding: 100px 0;
```
richardxia: You should get rid of this padding, since it's not actually present in the design. Adding padding like this will make it harder to reuse the component, since this particular element surrounds the element that actually has the gray background, so this is portion is "invisible". By putting padding here, it functions more like "margin", at least in the sense of putting invisible spacing between the gray box and other components on the page. This means that you could never place this component any closer than 100px away from another component on the page.

In order to ensure that we can reuse this component in as many contexts as possible, we generally should not put any margin in the component definition itself. Here, even though it is padding and not margin, because it is padding in the transparent area it basically functions as margin, since it's the gray box that a person will perceive to be the component itself.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  row-gap: 50px;
```
richardxia: This `row-gap` doesn't do anything because there's only one row in this grid, so you should delete it.

But you should also delete it as well as pretty much every other grid-related property on this element if you follow my earlier suggestion of combining the `gridParent` and the `gridAreaMiddle` `<div>` elements into a single one.

To be a bit more concrete, could you combine the `gridParent` and `gridAreaMiddle` classes, name it `carouselWrapper`, and delete every property except for `grid-column: 2`? In the HTML, you should also combine the `<div>` elements into one and also name its class to match.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+    padding: 0;
+  }
+
+  .gridAreaMiddle {
+    grid-column: 1 / span 13;
+  }
+}
+
+
+.innerWrapper {
```
richardxia: If you go with my suggestion to pull out the `<figure>` element into its own `<SlideQuote>` component, then could you rename this class to `.slideQuote`?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  background: var(--color-gray-300);
+  height: 100%;
```
richardxia, line 22: Since this is a bit tricky, I think it'd be useful to add a comment right after this that mentions that this makes all the slides the same height.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  margin: 0;
+  padding: 60px 50px;
+}
+ 
+@media (--tablet-and-down) {
+  .innerWrapper {
+    grid-column: 1 /span 13;
+    grid-row: 1;
```
richardxia: These properties don't do anything because the `.innerWrapper` is not the direct child of a `display: grid` element.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  font: var(--font-caption);
+  margin-top: 22px;
+}
+
+.dots {
+  padding-top: 19px;
+  text-align: center;
+}
+
+.dots button {
```
richardxia: It's probably not a good idea to use a tag name selector like this, since you're making certain assumptions about the internals of the carousel library's implementation.

One way that we may be able to select the individual buttons without using a tag selector would be to mix both local and global class names in this selector. This is a bit of an advanced usage of CSS modules, but you can [mark some classes as `:global`](https://github.com/css-modules/css-modules#exceptions), which tells CSS modules to not automatically generate a unique name anymore.

Since the built-in class name for the react-pure-carousel's `<Dot/>` component is named `carousel__dot`, I think we could replace this selector with:

```suggestion
.dots :global(.carousel__dot) {
```

I haven't tried this out before, so I'm not sure if this is exactly right, but I think it'll work out something like that.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  border-radius: 50%;
+  border-width: 0;
+  height: 6px;
+  margin: 10px;
+  padding: 6px;
+}
+
+button:disabled {
```
richardxia: Going along with my suggestion above to mix in `:global` CSS class names into these selectors, I think you actually want to use the CSS class name for the selected button, not a disabled button. Although the carousel library happens to disable the buttons when they are selected, they actually set a special CSS class name for the selected buttons, named `carousel__dot--selected`. Therefore, I think you can replace this selector with:

```suggestion
.dots :global(carousel__dot--selected) {
```

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  background-color: var(--color-sheltertech-blue);
+}
+
+button:not(:disabled) {
+  background-color: var(--color-gray-400);
```
richardxia: If you follow my advice above, you can actually move this background color property into the base `.dots :global(.carousel__dot)` selector, since you shouldn't have specificity issues anymore.

richardxia (commented):
Left some more feedback here. The component's styling mostly looks right, so most of my feedback has been focused on organization, naming, and specific methods for dealing with the carousel library's styles.

I also noticed that some of the mobile styles are a bit off, but I think I'd like to see you make these changes before I comment on those, since I think these will end up changing your HTML and CSS quite a bit.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Deleted bleedBackground.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: On Figma, isn't it aligned from column 1 to column 12 or is that the outer grid's maximum width?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Destructured.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Created SlideQuote component.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted padding.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Defined prop types to SlideQuote component and referenced prop type in BlockQuoteBlock.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted row-gap and combined gridParent and gridAreaMiddle into carouselWrapper.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Changed innerWrapper to slideQuote

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Added comment to address the reason we have 100% height there.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Changed.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Changed.

ivanleeswe: .dots :global(.carousel__dot--selected) <--- with the dot before carousel

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted.

davidagustin:
I asked design on figma on what to put for the other slide cards at and they replied with this link:
https://docs.google.com/document/d/1G6-5B8An-V-o7Dy6eWpuWcdXPgtXG1I_MOAxItOMRjA/edit

davidagustin:
I can pair program with you also if you need help on putting this in the new index

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
```diff
@@ -0,0 +1,40 @@
...
+        "We've heard from San Franciscans, including as part of our recent strategic planning community outreach that it's hard to know where to go when you're looking for supportive services. It's hard to know where to start, and hard to navigate different resources to find accurate information. We're glad to be supporting ShelterTech and the broader community to build SF Service Guide - a one-stop, reliable place to help connect residents to the services they need.",
+      attribution:
+        "Barry Roeder, Mayor's Office of Housing and Community Development",
+    },
+    {
+      quote:
+        "We’ve heard from San Franciscans, including as part of our recent strategic planning community outreach that It’s hard to know where to go when you’re looking for supportive services. It’s hard to know where to start, and hard to navigate different resources to find accurate information. We’re glad to be supporting ShelterTech and the broader community to build SF Service Guide – a one-stop, reliable place to help connect residents to the services they need.",
+      attribution:
+        "Barry Roeder, Mayor’s Office of Housing and Community Development",
+    },
```
richardxia, line 26: Looks like this quote is duplicated. Can you delete the duplicate?

```suggestion
```

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
richardxia: It is aligned from column 1–12, but because it spans the whole width and doesn't make use of the internal columns, you don't actually need the column definitions.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
richardxia: Awesome, this looks really clean now!

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,46 @@
...
+          ))}
+        </Slider>
+        <DotGroup className={s.dots} />
+      </CarouselProvider>
+    </div>
+  </div>
+);
+
+BlockQuoteBlock.propTypes = {
+  quotes: PropTypes.arrayOf(CarouselPropType).isRequired,
```
richardxia: You can actually reuse the props defined on the `SlideQuote` so that you don't need to maintain a completely separate `CarouselPropType` that duplicates the props on `SlideQuote`.

```suggestion
  quotes: PropTypes.arrayOf(PropTypes.shape(SlideQuote.propTypes)).isRequired,
```

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
richardxia: Thanks! Small nit: Could you put a space between the `/*` and `Height`?

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+    column-gap: 10px;
+    grid-column: 1/ span 13;
+    grid-column: 1 / span 13;
+    grid-template-columns: repeat(13, minmax(0, 1fr));
+    padding: 0;
+  }
+}
+
+.slideQuote {
+  background: var(--color-gray-300);
```
richardxia, line 19: Could you add [`box-sizing: border-box;`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing) to this style? Without it, the `height: 100%` ends up applying to just the content, but not the content + padding, which causes it to look like this:

Screen Shot 2020-11-09 at 10 49 14 PM


instead of:

Screen Shot 2020-11-09 at 10 49 22 PM


```suggestion
  background: var(--color-gray-300);
  box-sizing: border-box;
```

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+  margin: 0;
+}
+
+.attribution {
+  font: var(--font-caption);
+  margin-top: 22px;
+}
+
+.dots {
+  padding-top: 19px;
```
richardxia: Could you make this margin, not padding? Generally, when you want to space two different elements apart, you use margin. Padding is for putting spacing on the inside of an element, between its content and its border.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.stories.js
ivanleeswe: 👍 Sorry, deleted.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+  text-align: center;
+}
+
+.dots :global(.carousel__dot) {
+  background-color: var(--color-gray-400);
+  border-radius: 50%;
+  border-width: 0;
+  height: 6px;
```
richardxia: Does this do anything? I think the `padding` makes up for the height.

Also, it looks like these dots are twice as big as the ones in the design, but at the same time I think if we make them any smaller, they become too difficult to click, so maybe it's fine.

src/pages/new/index.js
```diff
@@ -106,6 +107,62 @@ export default () => (
...
+            attribution:
+              "Veronica Pastore, Director of Communications, Larkin Street Youth Services",
+          },
+          {
+            quote:
+              "There’s an overwhelming number of resources for the vulnerable populations that live here and it’s important to have a place where they can go where they know all the information is accurate, up-to-date, and consistent with what their understanding has been of that particular resource. And if there are any changes that need to be made those will be reflected.",
+            attribution:
+              "Julie Rosenthal, Director of Social Services, Homeless Advocacy Project",
+          },
+        ])
```
richardxia: I'm a bit confused by what's going on here. Why does the list of quotes appear twice here? I'm actually a bit surprised that this is syntactically valid.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 30px 1fr 30px;
+    grid-template-rows: auto;
```
richardxia: I think this property doesn't do anything, since you only have one element anyway.

```suggestion
```

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 30px 1fr 30px;
```
richardxia: Is it worth setting up a grid here? You don't actually use the 30px margins from here, since you separately set that up in the padding elsewhere.

You might just want to set `display: block` to get rid of the `grid`. If you do that, then you don't need to set `grid-template-columns` in the media query, since it will no longer behave like a grid.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+.carouselWrapper {
+  grid-column: 2;
+}
+
+@media (--tablet-and-down) {
+  .carouselWrapper {
+    column-gap: 10px;
+    grid-column: 1/ span 13;
+    grid-column: 1 / span 13;
+    grid-template-columns: repeat(13, minmax(0, 1fr));
```
richardxia: A few things here:

1. `grid-template-columns` and `column-gap` do nothing, since this is not a grid parent.
2. `grid-column` is being set twice.
3. `grid-column` is being set to span 13 columns, even though there are only three. Keep in mind that `grid-column` is set on the grid child, so the column spec refers to the parent element's columns, which in this case is the `.bleedWrapper`, which only has three columns.
4. If you go with my advice above of changing the `.bleedWrapper` to a `display: block` element in the mobile view, then you don't need to set any of this.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
```diff
@@ -0,0 +1,63 @@
...
+    padding: 0;
+  }
+}
+
+.slideQuote {
+  background: var(--color-gray-300);
+  /*Height 100% is to make all slides the same height */
+  height: 100%;
+  margin: 0;
+  padding: 60px 50px;
```
richardxia, line 24: You'll need to override this in a media query for mobile, since the padding is on 30px on mobile.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
```diff
@@ -0,0 +1,46 @@
...
+
+SlideQuote.propTypes = {
+  quote: PropTypes.string.isRequired,
+  attribution: PropTypes.string.isRequired,
+};
+
+const BlockQuoteBlock = ({ quotes }) => (
+  <div className={s.bleedWrapper}>
+    <div className={s.carouselWrapper}>
+      <CarouselProvider totalSlides={quotes.length} isIntrinsicHeight>
```
richardxia: Do you think you could configure this to autoplay? I think it'd look nice like that, and it'd help reinforce to the user that this is a carousel.

It looks like setting `isPlaying` will enable that. I played with it locally, and it also looks like the default `interval` is too short (5000 milliseconds), so maybe we could also set that to 10000 milliseconds?

richardxia (commented):
This is looking pretty good! I've left what I think will be the last major set of feedback, and I think we'll be very close to merging after you address these comments!

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
richardxia: Ah, yeah, thanks for catching that!

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Changed.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Whoa! What is this magic? Thanks for introducing box-sizing property.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Thanks!

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 You're right, it doesn't do anything. Deleted.

src/pages/new/index.js
ivanleeswe: 👍 Deleted. Sorry, I have a broken/sticky v-key.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Oh, changed grid-template-columns to display.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Ah, the display block makes it much easier. Thought the mobile view needs to be in a grid.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Added media query for mobile override padding with 30px.

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.js
ivanleeswe: 👍 Cool addition! Added!

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
ivanleeswe: 👍 Deleted

src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.module.css
richardxia: Yeah, it's a bit magical. It's a property that changes the definition of "width" and "height". Some people turn it on for all elements, since they find it easier to style.

richardxia (approved):
Awesome, this LGTM! Thanks for adding in the autoplay, I feel like that really makes the component come to life!

davidagustin (approved):
Looks good to me. Pair programmed with developer and helped with fixes and changes also.

ivanleeswe:
Woo, thanks team! Another one down. More, more, more!

PR #80, opened 2020-10-25
Create header navigation component

richardxia:
https://app.clubhouse.io/sheltertech/story/1618/build-new-top-navigation-component

## What changed

This adds the main navigation component described in https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=2%3A37. Note that I only implemented the first-level navigation, since we don't have a second-level navigation in the initial launch of the site, and it would have been more work to do implement that.

This uses https://github.com/negomi/react-burger-menu for the mobile, hamburger menu, which happened to be what our old site used. I updated it to the latest version of the library because I wanted to make use of some newer functionality for this implementation.

I created two sets of files: `Navigation.js/css` and `BurgerMenu.js/css`. The former contains the main styles, while the latter just contains styles that customize the BurgerMenu itself. Note that I created a `<NavList>` internal component that is used in both the main navigation and the BurgerMenu, and I was able to just restyle the components in both cases, so at least the actual navigation items themselves are structured identically.

Lastly, I added this to the main site's `<Layout>` component, which we use to wrap every page on the site. I also ended up having to make some changes to the legacy site's CSS, since they were globally styling the burger menu and interfering with the new styles. You can basically ignore most of the changes to `src/legacy/components/Navigation.scss`.

### Desktop view
Screen Shot 2020-10-24 at 10 28 28 PM

### Mobile view, closed
Screen Shot 2020-10-24 at 10 41 02 PM

### Mobile view, open
Screen Shot 2020-10-24 at 10 41 06 PM


davidagustin (changes requested):
What I'm requesting to change:
The resizing of the image sheltertech logo in the mobile view. It should be smaller.

What my question is:
Should the sheltertech logo be aligned with the left colored area of the Get Involved Grid Component?

Screen Shot 2020-10-25 at 3 37 58 PM

Screen Shot 2020-10-25 at 3 29 05 PM

Screen Shot 2020-10-25 at 3 28 55 PM

richardxia:
> The resizing of the image sheltertech logo in the mobile view. It should be smaller.

Thanks for catching that, I'll go fix that now.

> Should the sheltertech logo be aligned with the left colored area of the Get Involved Grid Component?

I don't think so, since that would be inside the bleed area. Although the Figma design for the top navigation is not aligned to the 12-grid column (I think it was designed back when there was a 15-column design), it looks like it intentionally should only have content within the main grid and not in the bleed area.

Here's a screenshot of the Figma design with the column overlay turned on:

Screen Shot 2020-10-25 at 3 43 30 PM

Do you agree with my reasoning, given the grid lines?

richardxia:
> Thanks for catching that, I'll go fix that now.

Actually, upon closer inspection, this looks correct to me. I already have a media query that shrinks the logo on smaller displays.

davidagustin:
> > Thanks for catching that, I'll go fix that now.
> 
> Actually, upon closer inspection, this looks correct to me. I already have a media query that shrinks the logo on smaller displays.

I see, the screenshot images on the PR gave me the impression that the logo did not changed. Should've verified it manually.

davidagustin:
Figma Example Below:
Screen Shot 2020-10-25 at 4 15 46 PM

Storybook Example Below:
Screen Shot 2020-10-25 at 4 17 12 PM

How should the items on the right be aligned then? I see that they're different in relation to positioning.

davidagustin:
> > The resizing of the image sheltertech logo in the mobile view. It should be smaller.
> 
> Thanks for catching that, I'll go fix that now.
> 
> > Should the sheltertech logo be aligned with the left colored area of the Get Involved Grid Component?
> 
> I don't think so, since that would be inside the bleed area. Although the Figma design for the top navigation is not aligned to the 12-grid column (I think it was designed back when there was a 15-column design), it looks like it intentionally should only have content within the main grid and not in the bleed area.
> 
> Here's a screenshot of the Figma design with the column overlay turned on:
> 
> Screen Shot 2020-10-25 at 3 43 30 PM
> 
> Do you agree with my reasoning, given the grid lines?

I agree with your reasoning on the left side but the right side threw me off.

richardxia:
> How should the items on the right be aligned then? I see that they're different in relation to positioning.

I assumed that the items should be aligned to the grid, since the design for the navigation was created before we had settled on the 12-column layout, and I assumed that we had not updated the designs to match the new 12-column layout. I could double check with @derekfidler on whether these should have that extra padding or whether they should be aligned to the columns.

The "Get involved" section's design calls for a left bleed, but not a right bleed, so the gray background extends all the way to the left, but only up until the right-most column on the right side.

richardxia:
I pushed up some updates with some significant changes, based on the updated mobile designs.

The biggest change is that the mobile menu is expected to push the main site to the left, and instead of having an "X" button, the hamburger button can be used to toggle the menu. This required me to significantly restructure the mobile menu, because although the library we are using supports this, it requires you to organize your HTML in a different way.

Basically, I had to pull out the hamburger menu as a totally separate component, place it at the top level of the page, create a new wrapper div that wraps the whole page, including the top navigation bar, and then wire up the nav bar's hamburger button and the element IDs so that the library knows which elements to slide in and out.

As such, I ended up splitting the code into multiple separate files, and having the `Navigation/index.js` file export two separate, non-default objects (`Navigation` and `BurgerMenu`). There's also a separate `NavList.js` file, since both the BurgerMenu and Navigation use it. However, because the styles are slightly different, I made the `<NavList>` component take in a `theme` prop, which allows me to pass in a different imported CSS module to control the styling.

I'm a bit worried that this has creeped up in complexity, but I hope that the code is still not too difficult to understand.

Also, I spoke with Derek about the padding, and he mentioned that the logo and Donate button are intentionally padded inward, so I've added that spacing in. I think this is pretty much Derek-approved at this point (including my hover effect for the nav links), so is ready for a final review. Please let me know if there's anything I can do to better document or explain how this component works.

davidagustin:
@derekfidler @richardxia Good to review or changes in design need to be made?

richardxia:
> @derekfidler @richardxia Good to review or changes in design need to be made?

I think this is good to review, since I don't think there are any more changes to the design coming.

davidagustin (approved):
Looks good to me. Good use of React Hooks

richardxia:
> Looks good to me. Good use of React Hooks

Thanks! Yeah, it's one of the few components we have on this site that is actually stateful, but thankfully it's really simple here.

PR #83, opened 2020-10-26
1660 create home page programs block

candywang:
Created the `ProgramsBlock` component for the home page using the block `ProgramCard` component


richardxia (commented):
Nice, this code looks pretty clean, so I don't think I have any comments about that!

Regarding the actual behavior of the components across different screen sizes, I have the following comments:

## Non-mobile, but smaller screen
On a smaller screen, but still "laptop-sized", the cards end up wrapping in an awkward way:

Screen Shot 2020-10-26 at 10 46 31 AM

I think this is being caused by your `max-width: 410px` on the actual card component itself. What you need to do is have the parent element here constrain the size of the card. The parent's job is to define how much space the children are allowed to have, and the children are forced to adjust to that space.

Your choice of flexbox here means that you should be setting the [`flex-basis`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis) property (or the third component of the [`flex`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex) shorthand based on some number that is proportional to the actual width of our block here. This math actually gets tricky because flexbox doesn't have the concept of the gap (or "gutter") between flex items, so you'd have to subtrack the space for the gap before dividing by 3.

Alternatively, you could switch to using grid, where you can force three columns per row and have the columns shrink accordingly.

## Mobile view

The margins are a bit too small between the title and the first card as well as between the other cards:

Screen Shot 2020-10-26 at 10 52 23 AM

I know we don't have any mobile designs here, but the general rule of thumb of whitespace in designs is that there should be larger spacing separating different components than the spacing within the component. For example, there should be more space separating two cards than space separating the image from the text within a single card. In your current code, you have a 20px margin separating the image from its title within the same card, but there's 0 margin between the bottom of a description of one card and the image of the next card, which is a bit confusing because it makes it look like the title and description are tied to the card below rather than the card above.

You may want to take a look at some of the mobile designs for other components to get some ideas for how much spacing we should have.

candywang:
I made the following changes per @richardxia suggestions and based on the new Figma mobile design:
- Changed program cards' theme to be in "dark mode" where the background is black and the font color is white
- Switched flexbox to grid display
- In the media query for the program card block I added `grid-row-gap: 30px;` for spacing between each program card and `margin-top: 30px;` property for spacing above the component that holds all the program cards so that there is visual space between the title and program cards block


Tablet view:
Screen Shot 2020-11-01 at 6 45 53 PM

Large mobile view:
Screen Shot 2020-11-01 at 6 46 09 PM

Small mobile view:
Screen Shot 2020-11-01 at 6 46 22 PM

davidagustin (dismissed):
No inconsistencies found, LGTM

richardxia:
On a laptop/desktop resolution screen, it looks like your cards are not aligned to the grid:

Screen Shot 2020-11-01 at 11 17 41 PM

This is most noticeable on slightly narrower windows, where you can eventually get it to the point where all the cards are right next to each other with no margin in between. I think you're probably missing some grid column gap in that innermost grid, which would provide a fixed, minimum spacing between the elements.

src/components/grid-aware/ProgramsBlock/ProgramsBlock.module.css
```diff
@@ -0,0 +1,76 @@
...
+  background: var(--color-black);
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr 40px;
```
richardxia: It looks like in the mobile designs for this component, it's actually a 20px margin on the sides, not 40px.

src/components/grid-aware/ProgramsBlock/ProgramsBlock.module.css
```diff
@@ -0,0 +1,76 @@
...
+/* Program card block */
+
+.programsWrapper {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+}
+
+@media (--tablet-and-down) {
+  .programsWrapper {
+    grid-row-gap: 30px;
```
richardxia: According to the designs, it looks like this gap should actually be closer to 40px.

src/components/grid-aware/ProgramsBlock/ProgramsBlock.module.css
```diff
@@ -0,0 +1,76 @@
...
+  font: var(--font-headline);
+  margin: 0;
+}
+
+
+/* Program card block */
+
+.programsWrapper {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
```
richardxia, line 67: Regarding my comment from last night, if you add a `grid-column-gap` property here, it should allow all the program cards to be aligned to the parent grid.

richardxia (commented):
This is looking pretty close! I think if you address the last few things I just commented on, this will be good to merge in!

src/components/grid-aware/ProgramsBlock/ProgramsBlock.module.css
davidagustin: The component as a whole has 30px gap

Screen Shot 2020-11-03 at 4 06 05 PM

Screen Shot 2020-11-03 at 4 06 17 PM

Screen Shot 2020-11-03 at 4 05 58 PM

The textbox in relation to the picture of another card is around 40px although
Screen Shot 2020-11-03 at 4 06 54 PM

Screen Shot 2020-11-03 at 4 06 49 PM

candywang: I noticed that the spacings between the three cards are inconsistent. The first gap between the first and second card has a spacing of 41px and the second gap between the second and third card has a spacing of 46px. I made a comment on the design to ask @derekfidler for his thoughts on this. In the meantime, I'll change this gap to 40px.

davidagustin: No the spacings are good @cw323 , you're looking at the textbox and not the component as a whole

candywang:
I fixed the issue for the missing `grid-column-gap: 30px` property in the `.programsWrapper` selector. The cards are now aligned to the grid regardless of laptop/desktop resolution screen size. I also changed the gap on the mobile view to have a 40px spacing.

**Narrow laptop view**:
Screen Shot 2020-11-03 at 5 08 55 PM

**Large mobile view**:
Screen Shot 2020-11-03 at 5 09 57 PM

davidagustin (approved):
LGTM after discussion with Richard

richardxia (approved):
Cool, this LGTM! Thanks for all the hard work on this!

candywang:
Thank you! 🤓

PR #84, opened 2020-10-26
Create Article Spotlight component

richardxia:
https://app.clubhouse.io/sheltertech/story/1664/create-article-spotlight-card-component

## Summary of changes

This PR adds the Article Spotlight component. I think this is a mostly straightforward implementation, but there are two things that I think are a bit more tricky, which I'll explain in more detail here.

In the desktop view, the black card is offset by 50px downward, so that it goes below the bottom edge of the background image. To achieve this effect, I had to create several wrapper layers of `<div>`s, each with a different purpose. I have a `.gridParent` element which sets up the usual 12-column grid. Then I have a `.gridAreaCard` child whose main purpose is to set the width of the card (here, spanning 6 columns). It also sets a `position: relative` for the next layer. Next we have a `.textCardWrapper` element, which actually sets the `50px` offset using a `position: absolute`. Finally, we have a `.textCard`, whose job is to act as a "space-filling, block-level component", which sets the actual black background of the card.

In the mobile view, I got rid of both the inner and outer grids entirely, instead laying out both the background image section and the text card below as normal block elements. I disabled all the grids and the relative/absolute positioning, which basically renders almost all the classes I described in the previous class useless. The `.textCard` element ends up acting as a normal block-level element and fills the full width of the page, setting a slightly smaller padding to add enough cushion between the text and the outer page.

The last thing I did was add this component to the home page, so you can see it in action.

### Desktop
Screen Shot 2020-10-26 at 1 33 58 AM

### Mobile
Screen Shot 2020-10-26 at 1 34 11 AM


src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.js
```diff
@@ -0,0 +1,70 @@
...
+      </div>
+    </div>
+  </div>
+);
+
+ArticleSpotlightCard.propTypes = {
+  eyebrowText: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  button: PropTypes.shape(Button.propTypes).isRequired,
```
davidagustin, line 66: I see that shape is used instead of arrayOf. Should arrayOf be used to give a developer the idea that they can use multiple buttons if that developer wanted to reuse this component??

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.module.css
```diff
@@ -0,0 +1,96 @@
...
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: 600px;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  grid-column: 2 / span 2;
+  grid-row: 1 / -1;
```
davidagustin: Is there a reason for the -1? grid-row: 1 works perfectly fine.

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.js
richardxia: I didn't use `arrayOf()` because I don't know if that actually makes sense for this component. It's called the "Article Spotlight Card," and presumably it's meant to highlight one specific call to action. Until we actually have an example of having two buttons here, I don't think it makes sense to add another one, since it may allow a developer to abuse the component in a way that breaks the design aesthetic.

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.module.css
richardxia: Good catch! I think this is probably a vestige from an earlier iteration of this where the grid had multiple rows, and where I wanted the background to span all rows, which is what the `-1` would have allowed it to do. I can remove the `/ -1` here because there's only a single row.

richardxia: Just fixed this, squashed it into the original commit, and did a force-push back out to this branch. LMK if this all looks good to you now.

davidagustin (approved):
Looks good to me after changes made

PR #85, opened 2020-10-27
Correct black color to pure black.

richardxia:
There was apparently a mistake in the [Design System Figma](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=116%3A441), where the hex code for the black we're using was incorrectly stated as `#202020`, when in fact it should have been a pure `#000000`.

This corrects the value of `--color-black` in our global CSS variables file to match.

### DIFF
--- a/src/stylesheets/global.css
+++ b/src/stylesheets/global.css
@@ -22,7 +22,7 @@
   --color-gray-700: #666;
   --color-gray-800: #484848;
   --color-gray-900: #212121;
-  --color-black: #202020;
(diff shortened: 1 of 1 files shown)


davidagustin (approved):
LGTM

candywang (approved):
Got the Figma notification. Thanks for correcting it!

richardxia:
Thanks for the quick review!

PR #87, opened 2020-11-02
1684 create two paragraph block component

candywang:
Created the TwoParagraphBlock component based on the new Design System in Figma and attached screenshots below.

---
## Storybook
**Web view**:
Screen Shot 2020-11-02 at 2 50 48 AM

**Tablet view**:
Screen Shot 2020-11-02 at 3 38 55 AM
Screen Shot 2020-11-02 at 3 38 51 AM

**Large mobile view**:
Screen Shot 2020-11-02 at 3 39 25 AM
Screen Shot 2020-11-02 at 3 39 31 AM

**Small mobile view**:
Screen Shot 2020-11-02 at 3 39 59 AM
Screen Shot 2020-11-02 at 3 40 17 AM
Screen Shot 2020-11-02 at 3 40 27 AM
Screen Shot 2020-11-02 at 3 40 36 AM

---

## Integrated with the main page
screencapture-localhost-8000-new-2020-11-02-03_47_24


src/pages/new/index.js
```diff
@@ -53,6 +55,46 @@ export default () => (
...
+        </>
+      }
+      image={{
+        url: darcelHeadshot,
+        alt: "Darcel Jackson head shot wearing a heather gray hoodie.",
+      }}
+      ctaButton={{
+        text: "More About us",
+        internalLink: "/foo",
+      }}
```
davidagustin: More buttons may be implemented or removed. The Buttons class is expected to have a map behavior and therefore buttons should be an array.

```suggestion
      ctaButton={[
        {
          text: "More About us",
          internalLink: "/foo",
        },
      ]}
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.js
```diff
@@ -0,0 +1,79 @@
...
+    </div>
+  );
+};
+
+TwoParagraphBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  paragraph1: PropTypes.string.isRequired,
+  paragraph2: PropTypes.node.isRequired,
+  image: ImagePropType.isRequired,
+  ctaButton: Button.propTypes.isRequired,
```
davidagustin: Since an array of buttons is expected, arrayOf should be used 
```suggestion
  ctaButton: PropTypes.arrayOf(Button.propTypes).isRequired,
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.js
```diff
@@ -0,0 +1,79 @@
...
+  const GridAreaBottomLeft = () => (
+    <div className={s.gridAreaBottomLeft}>
+      <div className={s.paragraph2Wrapper}>{paragraph2}</div>
+      <div className={s.ctaButton}>
+        <Button
+          text={ctaButton.text}
+          internalLink={ctaButton.internalLink}
+          externalLink={ctaButton.externalLink}
+          onClick={ctaButton.onClick}
+        />
```
davidagustin: A mapping behavior should be used since the Button class is expected a propType array

```suggestion
 {ctaButton.map((button) => (
          <div key={button.text}>
            <Button
              text={button.text}
              internalLink={button.internalLink}
              externalLink={button.externalLink}
              onClick={button.onClick}
            />
          </div>
        ))}
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedBackground {
+  background: var(--color-gray-300);
+  grid-column: 1 / span 2;
+  grid-row: 1 / -1;
```
davidagustin: I don't see any difference on Desktop view and mobile view with just grid-row: 1
Is there only one row? @richardxia I've been seeing this in a couple of commits also.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+    padding-top: 50px;
+  }
+}
+
+
+/* Main grid areas */
+
+.gridAreaTopLeft {
+  grid-column: 2 / span 4;
+  grid-row: 1;
```
davidagustin, line 59: I don't see any changes when property is deleted @richardxia should this be here?
```suggestion

```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: auto 590px;
+  padding-top: 122px;
+  row-gap: 84px;
+}
+
+.gridParent > * {
+  min-width: 0;
```
davidagustin, line 44: Are there negative widths for children in gridParent? I see no effect here @richardxia

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+/* Main grid areas */
+
+.gridAreaTopLeft {
+  grid-column: 2 / span 4;
+  grid-row: 1;
+}
+
+.gridAreaTopRight {
+  grid-column: 6 / span 6;
+  grid-row: 1;
```
davidagustin, line 64: Row 1 is auto but I see no activity when this is deleted @richardxia 
```suggestion

```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.stories.js
```diff
@@ -0,0 +1,64 @@
...
+    </>
+  ),
+  image1: {
+    url: image,
+    alt: "Darcel Jackson head shot wearing a heather gray hoodie.",
+  },
+  ctaButton: {
+    text: "More About us",
+    internalLink: "/foo",
+  },
```
davidagustin: See reasoning below for ctaButton in index.js
```suggestion
ctaButton: [
    {
      text: "More About us",
      internalLink: "/foo",
    },
  ],
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+}
+
+.gridAreaBottomLeft {
+  grid-column: 2 / span 5;
+  grid-row: 2;
```
davidagustin, line 69: gridAreaBottomLeft influences the row and taking out the other grid-row properties except this will have no changes. This is the only grid-row that should be implement. Am I right @richardxia ?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+}
+
+.gridAreaBottomRight {
+  grid-column: 8 / span 4;
+  grid-row: 2;
```
davidagustin, line 75: Also here @richardxia 
```suggestion

```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr;
+  }
+
+  .bleedBackground {
+    grid-column: 1 / span 3;
```
davidagustin, line 28: There are two columns specified above at line 25. Spanning by 3 in column one is too much.

```suggestion
    grid-column: 1 / span 2;
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,143 @@
...
+
+  .paragraph2Wrapper {
+    margin-right: 40px;
+    margin-top: 40px;
+  }
+
+  .image1Wrapper {
+    margin-left: 40px;
+    margin-top: 40px;
+    position: static;
```
davidagustin: Although position static does the behavior wanted. position: inherit is should be more appropriate right @richardxia ? We don't want to introduce new "un-needed" positions to elements right?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.stories.js
```diff
@@ -0,0 +1,64 @@
...
+);
+
+export const Gray = Template.bind({});
+Gray.args = {
+  title: "Rooted in the Community",
+  paragraph1:
+    "ShelterTech was founded in 2016 by Darcel Jackson, who experienced homelessness first hand. Darcel envisioned a Bay Area where technology serves everyone, and those who were most vulnerable had the digital infrastructure and tools they need to get back on their feet.",
+  paragraph2: (
+    <>
+      <p style={{ marginTop: "0" }}>
```
davidagustin: I see that the inline style prevents a downward shift in the desktop view for this paragraph. However if another developer with no knowledge of how the view should look like may be unaware of this behavior when passing a prop. Could you fix this in the Component itself?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.stories.js
```diff
@@ -0,0 +1,64 @@
...
+      <p>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
+        tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
+        veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
+        commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
+        velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+        occaecat cupidatat non proident, sunt in culpa qui officia deserunt
+        mollit anim id est laborum.
+      </p>
+      <p style={{ marginBottom: "0" }}>
```
davidagustin: Nothing is happening with here marginBottom 0 here but dont think an inline style is appropriate while passing props.

src/pages/new/index.js
```diff
@@ -53,6 +55,46 @@ export default () => (
         { text: "Work with us", onClick: () => {} },
       ]}
     />
+    <TwoParagraphBlock
+      title="Rooted in the Community"
+      paragraph1="ShelterTech was founded in 2016 by Darcel Jackson, who experienced homelessness first hand. Darcel envisioned a Bay Area where technology serves everyone, and those who were most vulnerable had the digital infrastructure and tools they need to get back on their feet."
+      paragraph2={
+        <>
+          <p style={{ marginTop: "0" }}>
```
davidagustin: See suggestion above in the stories file.

src/pages/new/index.js
```diff
@@ -53,6 +55,46 @@ export default () => (
...
+          <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+            eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim
+            ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut
+            aliquip ex ea commodo consequat. Duis aute irure dolor in
+            reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+            pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
+            culpa qui officia deserunt mollit anim id est laborum.
+          </p>
+          <p style={{ marginBottom: "0" }}>
```
davidagustin: See suggestion above in the stories file.

davidagustin (changes requested):

davidagustin:
I see Richard's commit at https://github.com/ShelterTechSF/sheltertech.org/pull/84 for the Article Spotlight Component using

button: PropTypes.shape(Button.propTypes).isRequired

@richardxia is it appropriate to use shape or array for the Button proptypes on this component?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: By default, it's 1 but I wanted `.gridAreaTopLeft` and `.gridAreaTopRight` to explicitly start at 1 since I'm using 2 rows for this component's grid layout.

src/pages/new/index.js
```diff
@@ -53,6 +55,46 @@ export default () => (
         { text: "Work with us", onClick: () => {} },
       ]}
     />
+    <TwoParagraphBlock
```
richardxia: Actually, you shouldn't add this to the home page because this isn't actually on the home page. It should be on the Volunteer page, but my PR that adds the Volunteer page hasn't merged yet: https://github.com/ShelterTechSF/sheltertech.org/pull/80

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.module.css
```diff
@@ -1,4 +1,4 @@
-/* Outter Grid */
+/* Outer Grid */
```
richardxia, line 1: giphy

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: I'm guessing people have been copying and pasting from the ThreeParagraphBlock, which actually has multiple rows.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.js
richardxia: Expected by whom? I don't think it's clear cut from the Figma designs, which only ever place one button there, so I think it's a bit of a stretch to call it "expected". That said, I agree that it would be reasonable to make this an array.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: Although I think you're correct that omitting the `grid-row` property from the other elements will not affect the final result, I think it's probably a good idea to just explicitly list them anyway because the implicit behavior is not very obvious.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: I don't think you ever want to use `position: inherit`. `position` is not a property that is normally inherited, so it's really strange to force it to be inherited. `inherit` means "set this to whatever the parent's value for this is". If the parent has `position: absolute`, then this would also get `position: absolute`; if the parent has `position: fixed`, then this would also get `position: fixed`. I can't think of a single situation where you would ever want this kind of a behavior.

`position: static` will reset this back to the normal behavior for all elements (_all_ elements start with `position: static`), so I think this is fine.

You could also invert the direction of the media query and wrap the other selector with `@media (--laptop-and-up)`, and that way, you wouldn't have to reset this property.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
davidagustin: With that said @cw323 change to grid-row: 1

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.stories.js
candywang: Thanks for pointing this out! I fixed it by removing the inline styling and added children selectors.

src/pages/new/index.js
candywang: Sorry, I just realized that. I'll remove this component from the home page.

davidagustin (approved):
LGTM

candywang:
I added the component onto the new Volunteer page.

**Gatsby view**:
screencapture-localhost-8000-new-volunteer-2020-11-03-19_58_25

screencapture-localhost-8000-new-volunteer-2020-11-03-19_58_14

src/pages/new/volunteer/index.js
```diff
@@ -1,7 +1,52 @@
...
 
+import TwoParagraphBlock from "../../../components/grid-aware/TwoParagraphBlock";
+import darcelHeadshot from "../../../components/grid-aware/TwoParagraphBlock/stories/darcel-jackson-headshot.png";
 import Layout from "../../../components/layout";
 
 export default () => (
-  <Layout>This is a placeholder for the volunteer page.</Layout>
+  <Layout>
+    <TwoParagraphBlock
+      title="Rooted in the Community"
```
richardxia: Since this is going to be the actual volunteer page, could you take the exact text and image from the [Figma design](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1764) and put them here?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.js
```diff
@@ -0,0 +1,85 @@
...
+    </div>
+  );
+};
+
+TwoParagraphBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  paragraph1: PropTypes.string.isRequired,
+  paragraph2: PropTypes.node.isRequired,
+  image: ImagePropType.isRequired,
+  ctaButton: PropTypes.arrayOf(Button.propTypes).isRequired,
```
richardxia: Since this is an array, could you pluralize the name of the prop as `ctaButtons`?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,151 @@
+/* Background bleed */
+
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
```
richardxia: I think this could probably be deleted, since you only have one row here.

I think I originally added this to my Three Paragraph Block to work around some weird issue with not being able to get the bleed background to span the whole row, but it doesn't seem like we even need it in that component now, either. Let's remove this line, and then I can also take a pass through other components where we added it and remove them as well.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,151 @@
...
+
+.bleedMainContent {
+  grid-column: 2;
+  grid-row: 1;
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    grid-template-columns: 40px 1fr;
```
richardxia: I see what you're doing here, but I think it may result in some unintuitive layout, since you're forced to "push back" the text at the lower levels. Eventually, we'll be able to use [CSS subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid) to solve things like this, but Firefox is currently the only browser that supports it.

I also had some trouble with this in the Three Paragraph Block, and I ultimately ended up not implementing the mobile view correctly, mostly just so that I could get the grid example out sooner for you all to start using as an example for your own components.

After having more time to think about this and seeing all the other types of components that we've been implementing for the site, I think a potential solution to this that can be applied consistently to most of our components is to basically "abandon" the grid at the bleedWrapper level and instead do most of the grid manipulations at the inner grid level. This is because our mobile designs either have the backgrounds fully bleed to both sides or have no background at all. It's really only the images that tend to hug the sides of the mobile views.

In more specific terms, we could have the `.bleedWrapper` only have a single column which spans the entire width of the page (e.g. `grid-template-columns: 1fr`), and then we could have the `.gridParent` define a real grid which sets the fixed side margins. Then we can align all text grid areas to the middle area while having more freedom to allow images to occupy the margin areas.

If this sounds good to you, could you give this a try and see if this makes the code and the styling cleaner?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,151 @@
...
+
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: auto 590px;
+  padding-top: 122px;
```
richardxia: I see that this matches the Design System Figma, but the version of this component on the actual Volunteer page appears to have a 140px padding. I'm a sucker for nice, round numbers, and it also matches the 1-paragraph block, so do you think we could change this to 140px?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: No, the reason this is here is because the default min-width is `auto` (and `auto` means however wide it needs to be to fit its content), which means that a child can stretch out its parent. This is most noticeable on images, since a large image can then take up more space than allocated to it unless if we have this property set.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,151 @@
...
+.title {
+  font: var(--font-headline);
+  margin: 0;
+}
+
+
+/* Image block */
+
+.image1 {
+  height: auto;
```
richardxia: Is there a reason you needed to set this? This is what heights are set to by default.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,151 @@
...
+@media (--tablet-and-down) {
+  .paragraph1Wrapper {
+    margin-right: 40px;
+    margin-top: 40px;
+  }
+
+  .paragraph2Wrapper {
+    margin-right: 40px;
+    margin-top: 40px;
+  }
```
richardxia: You actually need to override the font sizes on mobile. This is especially true of `.paragraph2Wrapper`, where the font size is way too small for significant body text on a mobile device.

richardxia (commented):
Left some feedback, mostly about the mobile styles, but in general, this is pretty awesome, and I feel like you're getting a pretty good handle on CSS grid and our kind of complicated multi-layer grid system!

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.js
candywang: Yes! I can do that or would you prefer me to refactor it to be a single button `ctaButton: PropTypes.shape(Button.propTypes).isRequired,`?

richardxia: I think `ctaButtons` makes sense here, since I could see in the future how we may want to have multiple buttons in a 2-paragraph block.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: I came across a couple of articles that suggested setting `height: auto` to [maintain the aspect ratio](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/)  of the image so that it would not be stretched or squashed and to [help browsers](https://html.com/attributes/img-height/)  render the page faster and more accurately but I can take it out since you're right on height being set to [auto by default](https://developer.mozilla.org/en-US/docs/Web/CSS/height).

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: Thanks for noticing this! I made the change for `.paragraph1Wrapper` mobile font size to 16px. However, I noticed on the Figma design that the mobile font size for `.paragraph2Wrapper` is 16px which is what it's currently set to.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: Thanks for the links to the articles, they were pretty interesting reads! Images are surprisingly one of the more complicated parts of HTML and CSS, since they have different default behavior from other kinds of elements and very often require us to preserve the aspect ratio in order to look nice, unlike most other elements in HTML.

I think we're safe to omit the `height: auto` here, since we're not explicitly setting the dimensions in the `` tag itself. From your first article, it seems like we may actually want to start setting that, since it'll prevent the "pop in" problem, but I think this is something we can address later, especially after we've integrated with the CMS.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
richardxia: You're right that the font size should be 16px, but when I check out and run this branch, I see that it's getting a size of 14px.

I think there's a slight inconsistency in the mobile Figma design, where the Typography section in the right side bar says that the name of the style is "Web / Small Font", even though it's a mobile design, and therefore should use the Mobile font styles. "Web / Small Font" is 16px, but "Mobile / Small Font" is 14px, and instead "Mobile / Medium Font" is actually 16px.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: Sounds great!

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,159 @@
+/* Background bleed */
+
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr;
```
richardxia: Sorry if this got lost in the shuffle from earlier reviews, but I think this isn't doing what we want it to do, since in the designs, the gray background only spans the left and middle areas, but not the right bleed area.

I think this might have happened because of my comment in https://github.com/ShelterTechSF/sheltertech.org/pull/87#discussion_r518604160, where I probably wasn't clear enough. I should have been more explicit that my suggestion to:

> In more specific terms, we could have the `.bleedWrapper` only have a single column which spans the entire width of the page (e.g. `grid-template-columns: 1fr`), and then we could have the `.gridParent` define a real grid which sets the fixed side margins.

I don't remember exactly what I was thinking when I wrote that comment, but I think I probably only meant for that to apply to the mobile media query version of this. Could you add the normal bleed grid column definition to the `.bleedWrapper` in desktop?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: auto 590px;
```
richardxia: I think you probably don't want to set the height of the second row to a fixed height of 590px, since this means that if you have more text, it'll spill outside of the component. If you want to make sure it's no shorter than 590px, you could use the [`minmax()`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax) function.

```suggestion
  grid-template-rows: auto minmax(590px, auto);
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,159 @@
...
+/* Component wrappers */
+
+.paragraph1Wrapper {
+  font: var(--font-body-large);
+  margin-top: 0;
+}
+
+.paragraph2Wrapper {
+  font: var(--font-body-small);
+  margin-top: 75px;
```
richardxia: Is this still desirable? I know that the component looks slightly different between the Design System mockup and the mockup of the Volunteer page, but I don't think this is needed if we have the `row-gap` defined on the main grid.

Edit: Ah, I see, in mobile, you need it in order to make sure the image doesn't overlap the paragraph. Do you think you could take this off `.paragraph2Wrapper`, and instead add it to the mobile media query version of `.image1Wrapper` as a `margin-bottom`? I think that would make it more obvious why this margin is here.

Also, I think you could reset the `top:` property to `0` on `.image1Wrapper` in mobile, since it makes the margins more complicated to reason about, and instead, you could just use normal margins. That will make it so that you don't have to do math to account for the relative offsetting, like you did with the `margin-top: 5px` in that element.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+.paragraph2Wrapper :first-child {
+  margin: 0;
+}
+
+.paragraph2Wrapper *:not(:first-child) {
```
richardxia: The `*` isn't needed on pseudo-classes.

```suggestion
.paragraph2Wrapper :not(:first-child) {
```

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+.image1Wrapper {
+  position: relative;
+  top: 35px;
+}
+
+@media (--tablet-and-down) {
+  .paragraph1Wrapper {
+    font-size: 16px;
```
richardxia: Could you just set this to `font: var(--font-body-medium)`? In mobile, that's set to 16px.

Also, can you add a media query for `.paragraph2Wrapper` to set the `font:` property to the same value? In mobile, both paragraphs should use `--font-body-medium`.

richardxia (commented):
Just a few more comments about the CSS, but this almost LGTM! Sorry for taking a while to get back to reviewing this, but I think we can get this in once these last comments are addressed!

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: If I add a `margin-bottom` property to `.image1Wrapper` in mobile it doesn't create a space between the image and the second paragraph block but if I add a `margin-top` property to `.paragraph2Wrapper` in mobile, it will create the spacing needed. Is this due to collapsing margins and what would be the best solution?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: Yes! Sorry, I did see it in the previous review but have yet to make the changes.

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: Thanks for going over this with me! 👻

richardxia (approved):
This LGTM! Note that you have a new merge conflict with `main` because of your other PR that merged in, since they both touched the Volunteer page, but that should hopefully be quick to resolve.

PR #88, opened 2020-11-05
1682 create image header component

candywang:
I created the ImageHeader component and instantiated it onto the Volunteer's page. There are two designs for this component. I followed the [Image Header design](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14) while referencing the grid layout for the [Volunteer page design](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1764). In this component, there are two images that overlap and I found the [MDN CSS position documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/position) to be extremely helpful. Below are previews of the component in both Storybook and Gatsby.

## Storybook

**Web view**:
Screen Shot 2020-11-04 at 3 56 59 PM
Screen Shot 2020-11-04 at 3 56 33 PM
Screen Shot 2020-11-04 at 3 56 45 PM

**Tablet view**:
screencapture-localhost-6006-2020-11-04-16_44_40

**Large mobile view**:
screencapture-localhost-6006-2020-11-04-16_41_05

**Small mobile view**:
screencapture-localhost-6006-2020-11-04-16_40_36

## Gatsby
screencapture-localhost-8000-new-volunteer-2020-11-04-16_50_33


src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,125 @@
...
+
+.image {
+  height: auto;
+  max-width: 100%;
+}
+
+.image1Wrapper {
+  margin-top: 50px;
+  position: relative;
+  z-index: 1;
```
davidagustin: The use of z-index is discouraged. If a sticky navbar is used, z-index would have to take into account items such as that.

The example in this link may help you get the overlapping behavior with two images: https://mastery.games/post/overlapping-grid-items/

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.image2Wrapper {
+  float: right;
+  left: 75px;
+  margin-bottom: 80px;
+  position: relative;
+  top: 271px;
+  z-index: 2;
```
davidagustin: See comment above

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,125 @@
...
+    margin-top: 0;
+    right: 40px;
+  }
+
+  .image2Wrapper {
+    bottom: 14px;
+    left: 40px;
+    margin-bottom: 40px;
+    top: 0%;
+    z-index: -1;
```
davidagustin: see comment above

davidagustin (changes requested):

candywang:
I removed the `z-index` properties in the CSS file and repositioned and resized image2 in the mobile views based on the Volunteer page design. There are discrepancies between the two mobile designs ([Image Header design](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14) and [Volunteer page design](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1764)) which I have asked Derek to clarify on Figma and am waiting for his response.

**Current preview of the component with new changes**:

Screen Shot 2020-11-05 at 12 48 11 AM

**You can see the differences between the two mobile designs below**:
| **Volunteer page design**  | **Image header design** |
| ------------- | ------------- |
| Screen Shot 2020-11-05 at 12 49 38 AM  | Screen Shot 2020-11-05 at 12 50 28 AM  |

davidagustin (approved):

davidagustin:
Looks good to me if the Volunteering design is what is wanted.

src/components/grid-aware/ImageHeader/ImageHeader.js
```diff
@@ -0,0 +1,71 @@
...
+        ))}
+      </div>
+    </div>
+  );
+
+  return (
+    <div className={s.bleedWrapper}>
+      <div className={s.bleedMainContent}>
+        <section className={s.gridParent}>
+          <div className={s.gridAreaRight}>
```
richardxia: `gridAreaRight` might not be the best name for this, since the positions vary between desktop and mobile. Also, you have two separate instances of it, wrapping each image separately, so it's a bit unintuitive to read HTML that looks more or less like this:

```html
<div class="gridAreaRight">...</div>
<div class="gridAreaLeft">...</div>
<div class="gridAreaRight">...</div>
```

How about you give these names that are more descriptive of their content rather than their location? Also, you may just want to give different CSS classes to the two grid areas that wrap the images, since that will allow us to target them separately.

Here are some suggestions for names: `gridAreaImage1`, `gridAreaImage2`, `gridAreaText`.

src/components/grid-aware/ImageHeader/ImageHeader.js
```diff
@@ -0,0 +1,71 @@
...
+  ctaButton,
+}) => {
+  const GridAreaLeft = () => (
+    <div className={s.gridAreaLeft}>
+      <h1 className={s.titleWrapper}>{title}</h1>
+      <div className={s.subtitleWrapper}>{subtitle}</div>
+      <div className={s.descriptionWrapper}>{description}</div>
+      <div className={s.ctaButtonWrapper}>
+        {ctaButton.map((button) => (
+          <div key={button.text}>
```
richardxia: You won't want to wrap the buttons with `<div>`s here; this is going cause the buttons to stack vertically if you have more than one, since each `<div>` is a block-level element.

Depending on how you want to do the CSS, you may still want a wrapper for each button, and you may even want those wrappers to be `<div>`s, but if you end up doing that, you will most certainly want some custom CSS for the wrapper.

Why don't you play around with this to make it look right with multiple buttons, and then we can discuss how this part should be structured? You may want to add an extra button in the Storybook story so that it's easier to test how that looks.

src/components/grid-aware/ImageHeader/ImageHeader.js
```diff
@@ -0,0 +1,71 @@
...
+  );
+};
+
+ImageHeader.propTypes = {
+  title: PropTypes.string.isRequired,
+  subtitle: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  image1: ImagePropType.isRequired,
+  image2: ImagePropType.isRequired,
+  ctaButton: PropTypes.arrayOf(Button.propTypes).isRequired,
```
richardxia: I forget if this got buried in the comments somewhere, but could you pluralize this to be `ctaButtons`? Pluralizing the names of props that hold arrays of items makes it easier to differentiate between singleton and array props.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedMainContent {
+  grid-column: 2;
+  grid-row: 1;
```
richardxia: There's no need to set this if you only have a single row (and in fact, a single element).

```suggestion
```

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+  }
+}
+
+
+/* Main grid */
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-column-start: 2;
```
richardxia: This property only has an effect on elements that are direct children of a grid. Although this element _is_ a grid parent (via the `display: grid` one line above) and although this element is the _grandchild_ of a grid parent (`.bleedWrapper`), it is itself not a grid child.

```suggestion
```

src/components/grid-aware/ImageHeader/ImageHeader.js
```diff
@@ -0,0 +1,71 @@
...
+  title,
+  subtitle,
+  description,
+  image1,
+  image2,
+  ctaButton,
+}) => {
+  const GridAreaLeft = () => (
+    <div className={s.gridAreaLeft}>
+      <h1 className={s.titleWrapper}>{title}</h1>
```
richardxia: Instead of calling this the `titleWrapper`, I would just call it `title`, since that what this _is_. Same goes with all the other elements here.

src/components/grid-aware/ImageHeader/ImageHeader.js
```diff
@@ -0,0 +1,71 @@
...
+      <div className={s.subtitleWrapper}>{subtitle}</div>
+      <div className={s.descriptionWrapper}>{description}</div>
+      <div className={s.ctaButtonWrapper}>
```
richardxia: One exception to what I said above about `*Wrapper`: this wouldn't make sense as just `ctaButton`, since it may contain multiple buttons. I think in similar situations in other components, we call this the `ctaButtonRow`, so you may want to use that name here.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+  margin-top: 15px;
+}
+
+.descriptionWrapper {
+  font: var(--font-body-medium);
+  margin-top: 20px;
+}
+
+.ctaButtonWrapper {
+  margin-bottom: 83px;
```
richardxia: This is a little bit tricky, but upon closer inspection of the design, I think the spacing between the button and the bottom of the component is not coming from the button itself, but rather the image on the right, which has an 80px margin to the bottom.

How about we set this to 80px, and we let larger of either these buttons or the images determine the bottom spacing? Another approach may be to define padding on a higher-level element (maybe the `gridParent` in our case), which would set the minimum spacing between the content and the bounding box.

Note that neither of these approaches will completely work for you without making some slight changes to how the images are being position, which I'll comment on separately.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+/* Right grid area*/
+
+.image {
+  height: auto;
+  max-width: 100%;
+}
+
+.image1Wrapper {
+  margin-top: 50px;
+  position: relative;
```
richardxia: Setting this by itself won't do anything. `position: relative` only does something useful if you set at least one of the `top/right/bottom/left` properties. Since you're using a top margin to deal with the top spacing, you can just get rid of this.

```suggestion
```

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+}
+
+.image2Wrapper {
+  float: right;
```
richardxia: Let's get rid of the `float` here. [Float](https://developer.mozilla.org/en-US/docs/Web/CSS/float) was really originally designed to let text wrap around an image, much like you would see in a print article or a textbook. For a very long part of the web's history, float was used in more advanced layouts, since we didn't have things like grid or flexbox, but now that we have these more precise and powerful tools, I think we should generally avoid using float unless if we are specifically trying to allow inline content fill around the floated item.

In your case, this float has an effect on the positioning of the image, but it's very complicated to reason about because it is being taken out of the normal flow of the document without any actual other content there. Let's use other techniques to get things positioned appropriately.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+
+/* Main grid areas */
+
+.gridAreaLeft {
+  grid-column: 2 / span 6;
+  grid-row: 1;
+}
+
+.gridAreaRight {
+  grid-column: 9 / span 4;
```
richardxia, line 52: If you split this CSS class up into two separate ones, one for each image, then for the second image, you may want to set `grid-column: 10 / span 3`, since it's really the width of 3 columns.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,121 @@
...
+@media (--tablet-and-down) {
+  .image1Wrapper {
+    margin-top: 0;
+    right: 40px;
+  }
+
+  .image2Wrapper {
+    left: 40px;
+    margin: 38px 0 40px 85px;
+    top: 0%;
```
richardxia: You probably don't need this combination of properties, since many of them are trying to push and pull the element along the same dimensions. While there are occasionally times when you may want to simultaneously set margins and relative positions for edges of the bounding box like this, it's usually best to try to stick with one method so that it's simpler to understand.

I also see that you're trying to unset some of the properties from the non-mobile view. Since the two views are trying to do very different things, it might be a good idea to wrap the desktop styles in its own media query (`@media (--laptop-and-up)`) so that we don't have to unset these here.

richardxia (commented):
Good job on this so far, since it's a pretty tricky component to lay out! I left some initial feedback on this, so please see if any of that helps with simplifying the classes and styling.

I suspect that we may need to iterate a bit more on the styling of the images, but let's try out this first set of changes before we talk more about that.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
candywang: Thanks for noticing this. I'll move this `position: relative;` property in the media query since I'm using the `right` property there.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
candywang: Thanks for the lesson and for linking the documentation!

candywang:
Made changes to the naming, styling, and positioning of elements:
- removed "Wrapper" from class names that are non-wrapper elements
- pluralized ctaButton to `ctaButtons` and changed the prop type
  - added styling to buttons wrapper and item
- renamed grid areas

Will add `width` property according to the final mobile design.

src/pages/new/volunteer/index.js
```diff
@@ -1,7 +1,31 @@
...
+        alt: "Classroom of volunteer members collaborating with one another.",
+      }}
+      image2={{
+        url: whiteboard,
+        alt:
+          "Team members collaborating together by looking and pointing at whiteboard covered with sticky notes full of ideas.",
+      }}
+      ctaButtons={[
+        {
+          text: "APPLY",
```
davidagustin: Text transform to uppercase is inherent when passing the text prop to a Button component.

src/components/grid-aware/ImageHeader/ImageHeader.stories.js
```diff
@@ -0,0 +1,53 @@
...
+    alt: "Classroom of volunteer members collaborating with one another.",
+  },
+  image2: {
+    url: whiteboard,
+    alt:
+      "Team members collaborating together by looking and pointing at whiteboard covered with sticky notes full of ideas.",
+  },
+  ctaButtons: [
+    {
+      text: "APPLY",
```
davidagustin: Text transform to uppercase is inherent when passing the text prop to a Button component.

davidagustin (approved):
Small negligible prop passing of text to button was commented but not too big of a deal. LGTM.

candywang:
Lowercased `text` value in `ctaButtons` prop.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,139 @@
...
+    text-align: right;
+  }
+}
+
+
+/* Left grid area */
+
+.title {
+  font: var(--font-headline);
+  margin-top: 90px;
```
richardxia: Rather than defining this on the `.title` element, I think it would make more sense to define this on the `. gridAreaText` element, since it makes more sense for `.gridAreaText` to describe the spacing on the "outside" of all of the elements.

In addition, you probably also want to set the bottom margin on the `.gridAreaText` element, since without that, the bottom edge of the overall `.bleedWrapper`/`<ImageHeader>` component is exactly flush with the CTA button, which means that if we placed another section just below this, there will be no spacing at all.

You can see this visually in the following screenshot, where the blue area (`.bleedWrapper`) ends immediately after the Apply button:

Screen Shot 2020-11-11 at 10 05 22 PM

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,139 @@
...
+@media (--tablet-and-down) {
+  .gridAreaImage2 {
+    text-align: right;
+  }
+}
+
+
+/* Left grid area */
+
+.title {
```
richardxia, line 74: This is a little bit subtle, but you'll also need to zero out the bottom margin on the `.title`, since `<h1>`s come with some default bottom margin. This is only visually noticeable on the mobile view, where the default 20px bottom margin of the `<h1>` is larger than the 15px top margin from the `.subtitle`, which means that 20px bottom margin wins and sets the total spacing between the elements.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -0,0 +1,139 @@
...
+
+@media (--tablet-and-down) {
+  .title {
+    margin-top: 30px;
+  }
+
+  .ctaButtonRow {
+    flex-flow: column;
+    margin-bottom: 30px;
+    margin-top: 20px;
```
richardxia: As far as I can tell from the designs, this top margin is still 30px in mobile, so I think you can just delete this property and let the desktop value of 30px take effect.

richardxia (commented):
Thanks! This is looking pretty good, and I only have a few minor CSS comments here. Be a bit careful about my requested change to the `.title` top margin, since that will have an impact in a number of other places, so you may need to fix up the spacing in other places to support that as well.

richardxia (approved):
LGTM! Thanks for adding this! It's awesome to have this role in and be the first component added to the Volunteer page!

PR #90, opened 2020-11-07
1660 create home page programs block

candywang:
According to Derek, the spacing between each program card in the mobile view should be 30px.


richardxia:
Your commit history looks a bit off here. You've got a one-line change, but you have what I think is the entire history of your previous PR. Could you make sure that you create new branches from the `main` branch? Since this is a short change, I hope you don't mind me asking you to do this over again, but from a fresh, clean branch, so that you just have a single commit with the changes that you made?

PR #91, opened 2020-11-07
Decrease grid-row-gap in .programsWrapper media query to 30px

candywang:
According to Derek, the spacing between each program card in the mobile view should be 30px.


richardxia (approved):
LGTM!

PR #92, opened 2020-11-07
1683 create stats block component

ivanleeswe:
Create the first section of the component which consists of the title.

Create the second section of the component which consists of the percent/number and description.

Style on top of a grid.


richardxia:
@ivanleeswe, it looks like you have all of the commits from https://github.com/ShelterTechSF/sheltertech.org/pull/79 included here. In the future, when you begin work on a new feature, can you make sure that you are creating the branch off the `main` branch?

For this branch, if you could remove all the commits from the Block Quote PR, that would be great, and let me know if you need any help with that.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
...
+  const NumberAndStatement = () => (
+    <div className={s.gridAreaBottom}>
+      {numberAndStatement.map((obj) => (
+        <div key={obj.number}>
+          <div className={s.number}>{obj.number}</div>
+          <div className={s.statement}>{obj.statement}</div>
+        </div>
+      ))}
+    </div>
+  );
```
richardxia: Do you think you could pull these two out as top-level React component definitions? Although I know I wrote the code in the ThreeParagraphBlock like this, with internal definitions for the components nested under the outer component, I would consider that to be the exception to rule rule. My justification for making an exception in that component is that the inner components do not really have a cohesive set of arguments (props), since some have images, some have paragraphs, and some have both, and it's a bit ad hoc which ones get which. It would add a great deal of noise to the ThreeParagraphBlock code to define the grid areas as separate, top-level components, since there would be a lot of code just passing props around.

For most of the other components on the site, I think there is more functional cohesion within the subcomponents, and therefore complexity of passing around props is lessened, and it would instead increase clarity to define them as their own components rather than nesting them inside others.

To put in more simply, pulling out the inner components will force you to explicitly pass around props, and that can get annoying in a component like the ThreeParagraphBlock.

If you all prefer consistency in our codebase, however, I'm happy to change the code in the ThreeParagraphBlock to match my ask here.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
...
+const StatsBlock = ({ title, numberAndStatement }) => {
+  const Title = () => <h1 className={s.title}>{title}</h1>;
+
+  const NumberAndStatement = () => (
+    <div className={s.gridAreaBottom}>
+      {numberAndStatement.map((obj) => (
+        <div key={obj.number}>
+          <div className={s.number}>{obj.number}</div>
+          <div className={s.statement}>{obj.statement}</div>
+        </div>
```
richardxia: You could potentially factor this out into a standalone component and give it a name, such as `StatCard`. I think doing this would help provide an extra level of organization and readability, since it would make the organization of the code more closely match the organization of the visual design.

When I look at the design, I first think "I see the title and I see a row of stats", and then I see the second level of "In the row of stats, I see three individual groupings of a number and text". You have the first-order hierarchy expressed in your code, with your `<Title>` and `<NumberAndStatement>` components, and I think you could add that second-order hierarchy that I'm describing here.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./StatsBlock.module.css";
+
+const StatsBlock = ({ title, numberAndStatement }) => {
+  const Title = () => <h1 className={s.title}>{title}</h1>;
+
+  const NumberAndStatement = () => (
```
richardxia: Nit: You should give this a plural name, since it contains a list of things. Alternatively, you could give it a singular name, but then it should be a name that describes the "container" or the "wrapper". Calling it `<NumberAndStatement>` like you have here makes it sound like there's just a single number and statement.

If you break out each individual number + statement combo into a standalone component named `<StatCard>` like I suggested in my other comment, then you call this `<StatCards>`. Alternatively, you may want to pick a name like `<StatsRow>` or `<StatsContainer>`, which would go with my alternative naming convention suggested above.

To relate this to one of the other components you are working on, the BlockQuoteBlock, you could look at how the [`pure-react-carousel`](https://github.com/express-labs/pure-react-carousel) names its subcomponents. They're following a similar convention, and you can see examples such as the `<Slider>` vs. each `<Slide>` as well as the `<DotGroup>` vs. each `<Dot>`.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
...
+  );
+};
+
+const numberAndStatement = PropTypes.shape({
+  number: PropTypes.string.isRequired,
+  statement: PropTypes.string.isRequired,
+});
+StatsBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  numberAndStatement: PropTypes.arrayOf(numberAndStatement).isRequired,
```
richardxia: Analogous to my request above about naming React components, the naming of props should also match their plurality. It's _really_ confusing to have a prop named `numberOfStatement` which is an array of `numberAndStatement`s.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  grid-template-rows: auto 0;
```
richardxia: Please remove this property, since your component only has a single row, so this doesn't do anything.

```suggestion
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+
+.gridParent {
+  background-color: var(--color-black);
+  color: var(--color-white);
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: repeat(2, auto);
+  padding-bottom: 150px;
+  padding-left: 75px;
```
richardxia: This is causing the items of this component to be misaligned from the grid. Is there a reason you are adding this in? With this property set, the component doesn't match the design, and when I disable it, the component does match the design, so I'm not sure I understand what it is doing.

```suggestion
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+
+/* Main grid */
+
+.gridParent {
+  background-color: var(--color-black);
+  color: var(--color-white);
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: repeat(2, auto);
```
richardxia: This isn't doing anything because you only have one row. Could you remove it?

```suggestion
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+.allTextWrap {
+  grid-column-end: 11;
+  grid-column-start: 2;
+}
+
+.gridAreaBottom {
+  display: flex;
+  flex-direction: row;
+  grid-row: 2;
+}
```
richardxia: Just a general comment about your grid. The `allTextWrap` element that you've introduced is causing things to be a bit awkward. You're currently halfway between two worlds, and you need to pick one, but what you have now is partially doing things one way, partially doing things another, and you have various CSS properties that have no effect because of the extra layer of indirection.

You can only set the grid child properties on the direct children of a grid parent element. In your current code, `.gridParent` is the grid parent, since it has `display: grid` set, and `.allTextWrap` is the one and only grid child. `.title` and `.gridAreaBottom` are children of `.allTextWrap`, which means that they are not grid children (they are grandchildren) and therefore none of the grid properties will do anything on them.

To be very specific, setting `grid-row: 2` will do nothing on `.gridAreaBottom` doesn't do anything because, despite its name, it is not a grid child.

There are two different approaches you can take:

1. You get rid of `.allTextWrap` altogether, both in the HTML and in CSS, and you have `.title` and `.gridAreaBottom` specify their grid row and/or column.
2. You leave `.allTextWrap` in, but get rid of all the grid-related properties on `.title` and `.gridAreaBottom`.

I'd probably prefer the former, since it reduces the number of elements in the component, which simplifies things.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+  .gridAreaBottom {
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+.title {
+  font: var(--font-title);
+  margin: 0;
+  padding-bottom: 54px;
```
richardxia: Please use margin to create spacing between elements, not padding. Padding is "part of" the element; margin is "outside of" the element.

To demonstrate, you can temporarily set a border on this element and see where it gets drawn, where the border represents the boundary of the element. With padding-bottom, the border gets drawn around the padding, which doesn't look right because it doesn't match one thinks of being "part of the element" when looking at the design.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.number {
```
richardxia: I see that this is a special typography style that is not part of our normal style guide, which makes sense, since this is a special component and we want something that adds extra emphasis to the statistic number.

However, this means that you should be setting all of the font-related styles. Right now, this is defaulting to the browser default font, which on macOS and Chrome is Times New Roman, but will vary on different browsers and OSs.

We use two typefaces on our site: Merriweather and Lato. Merriweather is a [serifed](https://en.wikipedia.org/wiki/Serif) font, whereas Lato is a [sans-serif](https://en.wikipedia.org/wiki/Sans-serif) font. In the Figma designs, I see that these stat numbers use an entirely different font, [Inter](https://fonts.google.com/specimen/Inter), which is a sans-serif font.

I don't think it's a good idea to include a third font face into our site, since that will make the site slower to load, as it's a third font that needs to be loaded, but we could try use Lato in its place, as an alternative san-serif font.

You should also appropriately set the font weight and line height, which are other properties that we normally have set with our font shorthand variables.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+  font-size: 72px;
+  height: 94px;
```
richardxia: Please do not set the height explicitly here. Setting both a fixed width and a fixed height will cause the element to be unable to expand when its content is too large to fit in the space, and cause it to [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow), which can cause odd visual bugs since the space that an element occupies will no longer match the space that it is actually drawn on screen.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+  margin-bottom: 5px;
+  margin-right: 110px;
+  width: 250px;
+}
+
+.statement {
+  font: var(--font-body-medium);
+  height: 52px;
```
richardxia: For the same reasons, please do not set a fixed height here. This is already causing issues, where the space occupied by the `<div>` is smaller than the space it takes to actually display the text, as you can see from the following screenshot, which demonstrates that the text of the element is partially outside of the blue content box:

Screen Shot 2020-11-07 at 5 33 51 PM

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+  margin-right: 110px;
+}
+
+@media (--tablet-and-down) {
+  .statement {
+    font: var(--font-body-medium);
```
richardxia: Please do not redundantly set the same properties and values in both the media query and non-media-query version of the selector. You should only override properties that need to change.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
richardxia: Also note that this spacing is not correct for mobile, so you will need to override it in a media query.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+  height: 94px;
+  margin-bottom: 5px;
+  margin-right: 110px;
+  width: 250px;
+}
+
+.statement {
+  font: var(--font-body-medium);
+  height: 52px;
+  margin-right: 110px;
```
richardxia: This margin probably does not belong on just this element. Visually, the number and the description appear to be one compound unit, so we should describe the spacing between those units at that level, not at the level of their individual subcomponents.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+@media (--tablet-and-down) {
+  .statement {
+    font: var(--font-body-medium);
+    height: 52px;
+    margin: 0;
+    margin-bottom: 40px;
+    margin-right: 85px;
```
richardxia: It's redundant to set the margin properties with the shorthand and the immediately again with the individual properties for each direction. Please either use one style or the other, but don't mix the two, since the shorthand is first setting the bottom margin to 0 and then the margin-bottom property is immediately overriding it.

richardxia (commented):
Thanks for submitting this! I left some feedback on stat block component, but in addition to looking at those, please also look at the initial comment that I left earlier about fixing up your Git history so that the unrelated changes are not also included here.

src/components/grid-aware/StatsBlock/StatsBlock.js
ivanleeswe: 👍 Makes sense!

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 I added the padding-left because of the bleed.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Agreed. Created new wrapper for both statement and number properties.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted margin 0.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: Added Lato font and line-height, but according to Figma, font-weight is bold which does not look right.

richardxia: I just left a comment on Figma just now, but I think it's OK to use Lato for now. There's consequences for requiring too many web fonts on a site, since it'll require a user to download all the fonts, so I'd like to stick to the original two fonts, if we can.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Changed to margin. Corrected spacing on mobile view.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted allTextWrap. Gave title and gridAreaBottom own grid column start and end.

src/components/grid-aware/StatsBlock/StatsBlock.js
ivanleeswe: Not sure if my new changes are what you're asking for.

src/components/grid-aware/StatsBlock/StatsBlock.js
ivanleeswe: Not sure if my new changes are what you're asking for.

src/components/grid-aware/StatsBlock/StatsBlock.js
richardxia: Yep, perfect! Thanks!

src/components/grid-aware/StatsBlock/StatsBlock.js
richardxia: Yep, this looks good. You got around the naming and plurality issue by just folding this into the main body and by pulling just the StatCard out.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,45 @@
...
+          ))}
+        </div>
+      </section>
+    </div>
+  </div>
+);
+
+StatsBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  statCards: PropTypes.arrayOf(StatCards).isRequired,
```
richardxia: You actually want this to be:

```suggestion
  statCards: PropTypes.arrayOf(PropTypes.shape(StatCards.propTypes)).isRequired,
```

I'm a little bit surprised that what you have seems to work, but I don't see it documented in the [official docs](https://reactjs.org/docs/typechecking-with-proptypes.html), so I wouldn't do it your way.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,45 @@
...
+  statement: PropTypes.string.isRequired,
+};
+
+const StatsBlock = ({ title, statCards }) => (
+  <div className={s.bleedWrapper}>
+    <div className={s.bleedMainContent}>
+      <section className={s.gridParent}>
+        <h1 className={s.title}>{title}</h1>
+        <div className={s.gridAreaBottom}>
+          {statCards.map((obj) => (
```
richardxia: Nit: I have a personal preference for destructuring objects in function arguments (although I admit I might be overusing this):

```suggestion
          {statCards.map(({ number, statement }) => (
```

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,45 @@
...
+            <div key={obj.number}>
```
richardxia: This `<div>` doesn't serve any purpose, so you can remove it. You'll still need to move the `key={object.number}` to the `<StatCard>`, though, since in React, we always have to set the [`key`](https://reactjs.org/docs/lists-and-keys.html#keys) special prop in the body of a `.map()`.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+  grid-column-start: 2;
+  margin-bottom: 54px;
+}
+
+.statCard {
+  margin-right: 85px;
+}
+
+.number {
+  font: var(--font-family-lato);
```
richardxia: This needs to be [`font-family`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family), not [`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font). The latter is a shorthand for multiple properties, but you're just trying to set the family here. This is probably why your font doesn't look right. You'll also want to make this bold.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+    flex-direction: column;
+  }
+}
+
+.title {
+  font: var(--font-title);
+  grid-column-end: 11;
+  grid-column-start: 2;
+  margin-bottom: 54px;
+}
```
richardxia, line 70: You'll need to zero out the top margin on this element, since the `<h1>` tag comes with native margins.

```suggestion
  margin-top: 0;
}
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+    display: block;
+    padding-bottom: 38px;
+    padding-top: 50px;
+  }
+}
+
+.gridAreaBottom {
+  display: flex;
+  flex-direction: row;
+  grid-column-start: 2;
```
richardxia: You need this to span multiple columns, since otherwise it ends up looking like this, where the `gridAreaBottom` is narrower than the actual items within it:

Screen Shot 2020-11-10 at 9 20 18 PM


```suggestion
  grid-column: 2 / span 10;
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+  .gridAreaBottom {
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+.title {
+  font: var(--font-title);
+  grid-column-end: 11;
+  grid-column-start: 2;
```
richardxia: Nit: I think for most of this site, we've stuck to the start + span syntax for specifying the columns for components, since we've found it to be more intuitive. Note that what you wrote down here actually means "start at and include column 2, and go through but _do not include_ column 11", meaning that this is actually 9 columns wide.

Also, not nit: According to the designs, this should be 8 columns wide. This is more noticeable in the [Design System Figma](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14), where there's enough text in the title to span two rows.

```suggestion
  grid-column: 2 / span 8;
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+  margin-bottom: 54px;
```
richardxia: According to the design, in mobile, this is only 40px, so you'll need to override this in a media query.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+}
+
+@media (--tablet-and-down) {
+  .statement {
+    margin-bottom: 40px;
+  }
+
+  .number {
+    font-weight: bold;
+    line-height: 65px;
```
richardxia: Two things:

1. You need to shrink the font size on mobile to 50px.
2. You should generally use a [unitless value for line-height](https://meyerweb.com/eric/thoughts/2006/02/08/unitless-line-heights/), because otherwise you may run into issues when nesting styles. 65/50 = 1.3, so that's the value you should use for line height.

```suggestion
    font-size: 50px;
    line-height: 1.3;
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+.statement {
+  font: var(--font-body-medium);
+}
+
+@media (--tablet-and-down) {
+  .statement {
+    margin-bottom: 40px;
+  }
+
+  .number {
```
richardxia, line 100: Could you flip the order of this and the `.statement` selectors so that they match the order that they appear in outside of the media query, which is also the order in which you see them in the HTML?

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+  width: 250px;
+}
+
+.statement {
+  font: var(--font-body-medium);
+}
+
+@media (--tablet-and-down) {
+  .statement {
+    margin-bottom: 40px;
```
richardxia: Actually, I'd probably apply this to the `.statCard`, not the `.statement`, since it's really the stat cards that are being spaced apart. This will be visually identical, but I think it makes more sense. Compare:

Screen Shot 2020-11-10 at 9 31 01 PM

vs.

Screen Shot 2020-11-10 at 9 31 14 PM

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,92 @@
...
+
+.statCard {
+  margin-right: 85px;
+}
+
+.number {
+  font: var(--font-family-lato);
+  font-size: 72px;
+  margin-bottom: 5px;
+  width: 250px;
```
richardxia: I think it makes more sense to set this on the `.statCard`, not the `.number`. It's the card that should have the fixed size, and the items within it will just take up as much space as they can take.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
richardxia: Also, I don't see the line-height actually here for the non-mobile view.

richardxia (commented):
The HTML is looking pretty good to me. Most of my remaining feedback is about the CSS.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+.number {
+  font-weight: bold;
+}
+
+.statement {
+  font: var(--font-body-medium);
+}
```
candywang: It looks like there should be a 5px spacing between the number and statement. A `margin-top: 5px;` property can be added here to make that possible.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+@media (--tablet-and-down) {
+  .title {
+    margin-bottom: 40px;
+  }
+}
+
+.statCard {
+  font-family: var(--font-family-lato);
+  font-size: 72px;
+  margin-bottom: 5px;
```
candywang: This can be taken out since it doesn't look like there's a bottom margin under each stat card item.
```suggestion
```

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./StatsBlock.module.css";
+
+const StatCards = PropTypes.shape({
```
candywang: Can you add "PropType" to the end of this variable name to stay consistent with the other components' naming convention? It'll also make the purpose of this variable more clear.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const StatsBlock = ({ title, statCards }) => (
+  <div className={s.bleedWrapper}>
+    <div className={s.bleedMainContent}>
+      <section className={s.gridParent}>
+        <h1 className={s.title}>{title}</h1>
+        <div className={s.gridAreaBottom}>
+          {statCards.map(({ number, statement }) => (
+            <StatCard key={number} number={number} statement={statement} />
```
candywang: I think this will be contingent on the final design but I think `statement` may be a bit more unique than `number` here.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+/* Main grid */
+
+.gridParent {
+  background-color: var(--color-black);
+  color: var(--color-white);
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding-bottom: 150px;
+  padding-top: 100px;
```
candywang: I would consider using the shorthand padding property if setting more than one value `padding: 100px 0 150px;`
- When three values are specified, the first padding applies to the top, the second to the right and left, the third to the bottom. - [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/padding)

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+/* Override grid item default of auto so that item contents cannot stretch the grid itself */
+.gridParent > * {
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .gridParent {
+    display: block;
+    padding-bottom: 38px;
+    padding-top: 50px;
```
candywang: As noted above, the shorthand padding property `padding: 50px 0 38px;` can be used here.

I noticed in the mobile view that the spacing at the bottom is 78px and it was accounted for by adding a media query for `.statCard` with `margin-bottom: 40px;`, which applies to the bottom of each stat card item. I would consider adding a wrapper around all the card elements and manipulate each item's gap within, then the bottom padding here can be set to 78px.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+    display: flex;
+    flex-direction: column;
+  }
+}
+
+.title {
+  font: var(--font-title);
+  grid-column: 2 / span 8;
+  margin-bottom: 54px;
+  margin-top: 0;
```
candywang: I would consider using the shorthand margin property here if setting more than one value `margin: 0 0 54px;`.
- When three values are specified, the first margin applies to the top, the second to the right and left, the third to the bottom. - [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/margin)

It's also possible to remove this `margin-bottom: 54px;` property and instead add a `margin-top: 54px;` property to the stat cards container (if implementing the wrapper element).

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+}
+
+@media (--tablet-and-down) {
+  .title {
+    margin-bottom: 40px;
+  }
+}
```
candywang: This media query can be taken out if implementing the wrapper element around the stat card items with a `margin-top` property.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,101 @@
...
+
+.statCard {
+  font-family: var(--font-family-lato);
+  font-size: 72px;
+  margin-bottom: 5px;
+  margin-right: 85px;
```
candywang: The spacing in the design has a 110px gap between each stat card item.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,43 @@
...
+          ))}
```
candywang: I would consider adding a stat card wrapper to contain all the stat card elements.

candywang (changes requested):
Looks great! I left a few comments mostly regarding spacing.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Good catch.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Deleted.

src/components/grid-aware/StatsBlock/StatsBlock.js
ivanleeswe: 👍 Changed to StatCardsPropType.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Nice!

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Changed.

ivanleeswe: Due to expanding the margin, I also had to add to grid column of the gridAreaBottom property.

src/components/grid-aware/StatsBlock/StatsBlock.js
richardxia: Good call, and I think you're right that the statement would be more unique! I can't see a situation where we'd have the exact same statement but different numbers.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,100 @@
...
+
+@media (--tablet-and-down) {
+  .statCard:first-child {
+    margin-top: 40px;
+  }
+}
+
+.statCard {
+  font-family: var(--font-family-lato);
+  font-size: 72px;
```
richardxia: I think these two font properties should probably be set on `.number`, not `.statCard`, since they only apply to the `.number`, as the `.statement` has its own font values set.

Also, you should set the line-height explicitly as well on the `.number` element. Normally our `font:` shorthand variables will set all of these properties together, but because this is a special typography in the design (nothing else is a 72px sized font), we need to make sure we set all these values.

In the Figma design, the line height is 93.6px, but we need to convert that to a unitless value by dividing that by the font size, so that results in a line-height of 1.3. This has a small, but noticeable affect on the spacing, which is why it's important to get this right.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,100 @@
...
+  .gridParent {
+    display: block;
+    padding: 50px 0 38px;
+  }
+}
+
+.gridAreaBottom {
+  display: flex;
+  flex-direction: row;
+  grid-column: 2 / span 11;
```
richardxia: Could you make this span 10, not 11 columns? Although this won't affect the layout of the current content today, if we changed the some of the other stuff, it could cause the cards to start taking up space right up until the right edge of the black container. You can see the asymmetry in the following screenshot:

Screen Shot 2020-11-11 at 10 44 50 PM


```suggestion
  grid-column: 2 / span 10;
```

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,44 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./StatsBlock.module.css";
+
+const StatCardsPropType = PropTypes.shape({
+  number: PropTypes.string.isRequired,
+  statement: PropTypes.string.isRequired,
+});
```
richardxia: This is a little bit redundant with the `StatCard.propTypes` declaration below. Could you delete this and just use `StatCard.propTypes` in the `StatsBlock.statCards` prop type definition?

src/components/grid-aware/StatsBlock/StatsBlock.module.css
richardxia: @ivanleeswe, I just unresolved this comment since I think it's still relevant but hasn't been addressed. Because the `<h1>` tag comes with default margins, it causes the total spacing between the `.title` and the stat cards below to be larger. See the following screenshot:

Screen Shot 2020-11-11 at 11 00 08 PM

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,100 @@
...
+.title {
+  font: var(--font-title);
+  grid-column: 2 / span 8;
+}
+
+@media (--tablet-and-down) {
+  .statCard:first-child {
+    margin-top: 40px;
+  }
+}
```
richardxia: Could you move this below the definition of `.statCard` (and put it in the same media query as the other `.statCard` mobile style)?

It doesn't make a difference in CSS, but I think it makes it easier to read if we have the non-media query, non-pseudo-class version of a selector appear before any other special variants. Also, even within the same media query, can we put the non-pseudo-class version before the pseudo-class versions? By [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) I mean things like `:first-child`.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
richardxia: I like this idea as well, although I think what we have is also OK.

In the future, when there's more browser support, it'd be great to use the [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) property with flexbox, since it'll help us add spacing on the insides of a flexbox container, rather than relying on the `:first-child`/`:last-child` overrides that we have to do. Support for this _just_ [landed in Chrome](https://web.dev/flexbox-gap/) a couple months ago, and Firefox has had it for a year, but Safari doesn't support it yet. Hopefully Apple adds support for it in Safari soon, since that means that we'll be able to make use of this, although we may still need to hold off from using it because not everyone updates their iPhones and Macs to get newer versions of Safari.

https://caniuse.com/flexbox-gap

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Moved to .statCard and added line-height to number element.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Changed back to span 10. However, the width of the statements is affected and does not end/start at the correct word compared to what is shown in Figma.

src/components/grid-aware/StatsBlock/StatsBlock.js
ivanleeswe: 👍 Done.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Done.

src/components/grid-aware/StatsBlock/StatsBlock.js
```diff
@@ -0,0 +1,38 @@
...
+          ))}
+        </div>
+      </section>
+    </div>
+  </div>
+);
+
+StatsBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  statCards: PropTypes.arrayOf(PropTypes.shape(StatCard.propTypes)).isRequired,
```
davidagustin, line 35: You can declare and initialize your own StatCard PropType Shape and shorten this line of code.

davidagustin (changes requested):

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,100 @@
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedMainContent {
+  grid-column: 2;
```
davidagustin, line 15: Im seeing a right bleed in the figma design. Does the component need to extend into the right column of the outer grid?

davidagustin: Screen Shot 2020-11-15 at 3 44 15 PM
Screen Shot 2020-11-15 at 3 42 34 PM

davidagustin (approved):
Found out some inconsistencies when comparing to the Figma. Pair programmed and got changes in.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,100 @@
+.bleedWrapper {
+  display: grid;
+  grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
+  margin: auto;
+  max-width: var(--grid-bleed-max-width);
+}
+
+.bleedMainContent {
+  grid-column: 2 / span 2;
```
richardxia, line 10: You can't actually set this here, since this will cause the inner grid to be misaligned. If you have the inner grid span the outer grid's center and right bleed areas, then the inner grid will evenly divide that space into 12, which means that it'll be bigger than an inner grid that is only in the center area.

Since you have the black background out in the right bleed area, you probably need to go with the approach that we've taken in a few of the other components with asymmetric bleeds, such as the TwoParagraphBlock, where we create a separate `<div>` whose job is to paint the background: https://github.com/ShelterTechSF/sheltertech.org/blob/b7c96e585126cb9c142064a18008ce1debfe1dc3/src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css#L10-L14

Note that this also has the side effect of causing the mobile view to lack any right padding, which is noticeable on a small phone screen (320px):

Screen Shot 2020-11-15 at 5 15 50 PM

richardxia (commented):
Awesome, this is almost there! I just have one last comment, and I can work with you to get this fixed, since this may require changes to a few places at once. But aside from that, this is looking really clean and good!

richardxia (approved):
LGTM!

candywang (approved):
LGTM! 👍🏼

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: 👍 Fixed with Richard's assistance! Had to create a new bleedBackground wrapper with row property.

PR #94, opened 2020-11-10
1663-create-video-spotlight-block-component

davidagustin:
Screen Shot 2020-11-10 at 3 11 25 AM
Screen Shot 2020-11-10 at 3 10 57 AM


src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
+.bleedWrapper {
+  background-color: var(--color-black);
```
richardxia: Could you leave out the black background here? It's not actually part of the component design, as shown in the [Design System Figma](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=808%3A14).

I know that on the actual Home page, you see a black background there, but I think it actually represents a more complex thing going on, where it's actually part of the Home page itself, intersecting with both the Video Spotlight and the Programs blocks. I haven't fully thought through how we should handle it, but I think we can deal with that and the overall spacing adjustments to components on the Home page at the same time.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
```diff
@@ -0,0 +1,75 @@
...
+
+const VideoSpotlightBlock = ({
+  eyebrowText,
+  description,
+  button,
+  image,
+  playButtonLink,
+}) => (
+  <div className={s.bleedWrapper}>
+    <div className={s.gridAreaImage}>
```
richardxia: This doesn't match our naming scheme, and could potentially cause a little bit of confusion. We tend to use `bleed*` as a prefix for any elements that are either the bleedWrapper itself or immediate children of the bleedWrapper. We tend to use `grid*` as a prefix for any elements that are either the `gridParent`, which sets up the 12-column grid, or its immediate children.

This element is actually part of the bleed grid, so maybe we could call this `bleedImage`?

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
...
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .bleedBackground {
+    height: 250px;
+  }
```
richardxia: This class doesn't appear to exist.

```suggestion
```

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
...
+.bleedMainContent {
+  grid-column: 2;
+  /* Override grid item default of auto so that item contents cannot stretch the grid itself */
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    display: flex;
+    flex-direction: column;
```
richardxia: Can this just be replaced with `display: block`? A column flex is unnecessary unless if you are actually trying to get items to grow proportionally with `flex-grow` or `flex-shrink`.

```suggestion
    display: block;
```

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
...
+  text-align: center;
+}
+
+@media (--laptop-and-up) {
+  .gridAreaImage {
+    align-items: center;
+    display: flex;
+    justify-content: center;
+    position: absolute;
+  }
```
richardxia: Couple notes on this:

1. As I requested in the JSX, I think this should really be named something more like `bleedImage`, and it should also be moved closer to the elements more related to it. This has more in common with `bleedMainContent` as well as `playButton` and `image` than it does with `gridAreaCard`.

2. I think your positioning and centering is a bit overly complicated here, since you have multiple layers of absolute position and weird stuff going on with flex and centering. Could you:
    1. Ditch the flexbox altogether; it's kind of confusing to reason about the position of the play button and the image with the flexbox and with the absolute positioning of the play button
    2. Also ditch the `align-items: center` and `justify-content: center`.
    3. Continue to position the `playButton` absolutely, but use the old-school way of centering an element vertically and horizontally through the use of `top: 50%`, `left: 50%`, and `transform: translate(-50%, -50%)`?

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
...
+
+  .gridAreaCard {
+    grid-column: 8 / span 5;
+    position: relative;
```
richardxia, line 57: If you're only using this to make this element appear above the image, then you should just use `z-index: 1` here.

Alternatively, you could change the `.gridAreaImage` element to be positioned via the grid, rather than absolutely, and then it would naturally lie below the card here. However, doing that will require you to add another wrapper layer in order to center the play button over the image, since positioning the `.gridAreaImage` using the grid will result in that receiving the full width of the grid columns rather than the width of the image.

I can explain this over a call in more detail, if you'd like.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  .textCardWrapper {
+    padding-top: 87px;
```
richardxia: Please use margin, not padding, to push items apart.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
```diff
@@ -0,0 +1,75 @@
+import PropTypes from "prop-types";
+import React from "react";
+import Button from "../../inline/Button";
+import playIcon from "../VideoHeader/PlayIcon.svg";
```
richardxia: Please don't import images from other components like this. This makes the import hierarchy asymmetrical, since it's only by coincidence that the VideoSpotlightBlock imports from the VideoHeader's image, even though the opposite would also be possible. Instead, as a general rule of thumb, if you find yourself doing that, you should extract the common part into a third area that they both depend on.

But this entire comment is moot, since it looks like the play icon in the design is actually much larger for this component than the Video Header component's. Could you make sure you use the properly sized icon?

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.stories.js
```diff
@@ -0,0 +1,38 @@
...
+
+export const VideoSpotlightBlockDefault = Template.bind({});
+VideoSpotlightBlockDefault.args = {
+  eyebrowText: "Our Impact",
+  description:
+    "Over 3,000 people have daily internet access in local shelters and resource centers.",
+  button: { text: "View Annual Report", internalLink: "/foo" },
+  image: {
+    url: videoSpotlightBlockImage,
+    alt: "Video spotlight of man speaking.",
```
richardxia: Just to honor him, even though it's just in a Storybook story, could you use Aaron's name here?

```suggestion
    alt: "Video spotlight of Aaron speaking.",
```

src/pages/new/index.js
```diff
@@ -23,6 +23,8 @@ import image2 from "../../components/grid-aware/ThreeParagraphBlock/stories/imag
 import image3 from "../../components/grid-aware/ThreeParagraphBlock/stories/image3.png";
 import VideoHeader from "../../components/grid-aware/VideoHeader/VideoHeader";
 import videoHeaderImage from "../../components/grid-aware/VideoHeader/stories/VideoHeader.png";
+import VideoSpotlightBlock from "../../components/grid-aware/VideoSpotlightBlock";
+import videoSpotlightBlockImage from "../../components/grid-aware/VideoSpotlightBlock/stories/VideoSpotlightBlock.svg";
```
richardxia: Could you use a PNG and not an SVG for this image? This ended up embedding a PNG into the SVG anyway, since SVGs can't be used to encode non-vector data, like an actual photograph.

richardxia (commented):
Overall, this is pretty good! I have a few comments about how the positioning of the image was done, but other than that, this looks fairly good.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: deleting

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: okay

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: I thought we were supposed to avoid the use of z-index just as a good practice. Like what if something from the navbar overlaps?

davidagustin: Will change to use z-index right now though.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.stories.js
davidagustin: Okay

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
richardxia: I would argue that using `position: relative` just to control the stacking order is kind of obtuse, since that's not a typical usage of it. `z-index` is at least very explicit about what you're trying to do with it.

I was also somewhat hoping that you would use grid positioning on the image, which would do away with this other stuff.

I guess alternatively, you could set a negative z-index on the image, which would push it below card.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: It would be complex to use the grid on the .gridAreaImage because there is a slight offset in the downward direction that overlaps with the space outside its normal boundries.

src/pages/new/index.js
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  .image {
+    width: 100%;
+  }
+
+  .playButton {
+    top: initial;
+    transform: initial;
+  }
```
richardxia: I think this, combined with the flex positioning from the `.bleedImage`, is making the positioning of this element difficult to reason about. The way it currently renders, it looks like the play button is off-center, slightly to the right.

I think it'd also be a good idea to stick to absolute positioning and transforming the element to achieve centering, rather than mixing both flex and absolute position when attempting to center the play button.

Also, could you organize your CSS so thing that are related to each other are closer to each other? It's a bit difficult to read your CSS right now because you have components within the `bleedImage` and the `bleedMainContent` mixed together, and you have the media queries for `--tablet-and-down` for the `.playButton` very far away from the `--laptop-and-up` and the non-media-query version.

Could I suggest that you group your CSS selectors to more closely match the hierarchical grouping of elements in the HTML? The two main subsections of the component are the `.bleedImage` and the `.bleedMainContent`. I think it would be helpful if you put all the classes related to the `.bleedImage` together and all the classes related to the `bleedMainContent` together.

richardxia (commented):
This looks pretty good, but some of the positioning is a little bit off in mobile, where the play button ends up being off-centered.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: Pair programmed with developer and resolved

richardxia (approved):
LGTM! Thanks for making those changes and reorganizing the file!

PR #95, opened 2020-11-11
1685 create one paragraph block component

candywang:
Created the One Paragraph Block component
- Due to comments regarding the background bleed in the Two Paragraph Block component, I may need to make a few adjustments
  - I set the background bleed to black but may need to change this and instead incorporate a theme for light and dark mode

## Storybook
**Mobile Large**:
Screen Shot 2020-11-11 at 3 03 00 AM

**Default**:
screencapture-localhost-6006-2020-11-11-02_53_19

## Gatsby
screencapture-localhost-8000-new-volunteer-2020-11-11-02_52_03


candywang:
Refactored code relating to `ctaButtons` - name, prop type, class selectors, and styling

src/pages/new/volunteer/index.js
```diff
@@ -1,7 +1,41 @@
...
+            Help us organize events and campaigns to raise donations for our
+            ShelterConnect installations that provide free wifi and hardware for
+            local housing shelters.
+          </p>
+          <p>Fundraising & Partnerships</p>
+          <p>
+            To create user-friendly websites, we need front- and back-end
+            developers and dev ops professionals who know React or Ruby on
+            Rails.
+          </p>
```
davidagustin: Are you intending to use the default properties of the p element? If not, they need to be turned off.

Screen Shot 2020-11-15 at 2 59 29 PM

https://www.w3schools.com/cssref/css_default_values.asp

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.stories.js
```diff
@@ -0,0 +1,47 @@
...
+      <p>
+        Help us organize events and campaigns to raise donations for our
+        ShelterConnect installations that provide free wifi and hardware for
+        local housing shelters.
+      </p>
+      <p>Fundraising & Partnerships</p>
+      <p>
+        To create user-friendly websites, we need front- and back-end developers
+        and dev ops professionals who know React or Ruby on Rails.
+      </p>
```
davidagustin: See comment below.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
```diff
@@ -0,0 +1,125 @@
...
+  margin-top: 20px;
+}
+
+.jobs *:nth-child(odd) {
+  font: var(--font-subtitle);
+}
+
+.jobs *:nth-child(even) {
+  font: var(--font-body-medium);
+}
```
davidagustin: Is the asterisk necessary to select the elements with the .jobs class?

richardxia: No, they are not. The pseudo-classes are enough. You may also want to use the direct child select `>` so that you don't select even more nested elements, since these will select any element at any depth of the hierarchy.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.js
```diff
@@ -0,0 +1,54 @@
...
+          <GridAreaRight />
+        </section>
+      </div>
+    </div>
+  );
+};
+
+OneParagraphBlock.propTypes = {
+  title: PropTypes.string.isRequired,
+  jobs: PropTypes.node.isRequired,
```
richardxia: Since this component has the generic name of "one-paragraph block", we should expect to use it for a number of other use cases as well, so we probably shouldn't name this `jobs`. How about `description`, which would make it match the `title` and `description` props of the 2- and 3-paragraph block components?

We should also rename the associated CSS classes.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
```diff
@@ -0,0 +1,125 @@
...
+.title {
+  font: var(--font-headline);
+}
+
+
+/* Jobs block */
+
+.jobsWrapper {
+  margin-top: 0;
+}
```
richardxia: I think this isn't necessary, since the `<div>` element doesn't come with any default margins.

davidagustin (changes requested):
CSS needs to be refactored

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
```diff
@@ -0,0 +1,125 @@
...
+
+@media (--tablet-and-down) {
+  .jobsWrapper {
+    margin-top: 40px;
+  }
+}
+
+.jobs *:first-child {
```
richardxia: OK, I see what you're doing here now. I think this is one of the few places where I think using tag selectors is appropriate, since in the future, we'll expect the actual copy to come from non-technical users through the CMS, where they'll likely be editing through a rich-text interface. It'll eventually to us in the form of [Rich Text](https://user-guides.prismic.io/en/articles/383762-rich-text) objects from Prismic, which boils down to simple HTML elements.

For now, how about you mimic this by using heading tags (`<h1>`, `<h2>`, ...) and paragraph tags (`<p>`) and styling them based on the tag name? This will be more robust than hoping that we will only have one paragraph per heading.

richardxia (commented):
Nice job on this! Most of the component looks great, and my only comments are about how the paragraph part is being handled, where I'd recommend using heading and paragraph tags in the markup rather than trying to style items based on their position in the list.

davidagustin (approved):

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
```diff
@@ -0,0 +1,127 @@
...
+.descriptions >:not(:first-child) {
+  margin-top: 20px;
+}
+
+.descriptions h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
```
richardxia: You need to add `.descriptions` for each of these, since otherwise they would affect all `h*` tags on the page. The commas separate whole selectors.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
```diff
@@ -0,0 +1,127 @@
...
+  font: var(--font-subtitle);
+}
+
+.descriptions p {
+  font: var(--font-body-medium);
+}
```
richardxia: Actually, it might be a better idea to just define this on the `.descriptions` tag directly, since that will set the "base" font that will be inherited by all elements. This would allow it to apply to other elements besides `<p>` tags.

richardxia (approved):
LGTM! I just have a couple remaining components, but I think this is good to merge once they're addressed!

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.module.css
candywang: Thanks for pointing this out!

PR #96, opened 2020-11-11
Add Modal component and integrate Mailchimp form

richardxia:
# Overview
This got a bit more complicated and messier than I had wanted, but I think this is in good enough shape to start reviewing.

This really does a few things at once, all related to the Mailchimp form modals on the site:

1. It adds the [`react-modal`](https://github.com/reactjs/react-modal) library, and it sets up a standalone, reusable Modal component that can be used for other things in our site.
2. It integrates an embedded Mailchimp form for the Volunteer signup page, which I also turned into a separate Storybook component.
3. It adds these to the Home page, specifically for the Volunteer button on the Three Paragraph Block component, which also involved setting up some React state on the Home page.

## Modal Component

Screen Shot 2020-11-11 at 8 56 13 AM

https://github.com/ShelterTechSF/sheltertech.org/commit/f9f6f6e4ee6c209795ecee7a106e2fc699d9eaca

I added a standalone Modal component to `src/components/grid-aware/Modal` based on the styles found in our [Figma designs](https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=1746%3A100).

This component just sets up the "outer" parts of the modal, and it expects the insides to be populated by the caller via the children. Specifically, this component sets up the overlay, which is the gray, partially transparent background, the modal container itself, and the close button. It's stateless, so it expects that its state is managed by its parent, passing in the `isOpen` prop as well as the `setIsOpen` function prop.

There's some weird stuff going on with this `ReactModal.setAppElement()` thing, which is something that the library [uses for accessibility purposes](http://reactcommunity.org/react-modal/accessibility/), where it wants to know the top-level element containing the application, but not containing the modal itself, so that can indicate to screen readers that the app element should be "disabled" when the modal is open. For us, unfortunately, this causes a chicken-and-the-egg problem, because on the very first render, our modal component is actually rendered _before_ the app element, meaning that if you try setting the app element before the app element is actually instantiated on the page, react-modal will hit an error. In Storybook, I worked around this by disabling the accessibility feature, but in our actual app, I had to do something weird where I used `useEffect()` to only set the app element after the first render.

I haven't fully researched all of the properties from react-modal, and it's possible that we may want to set a few more of those, so I can look into that next.

## Mailchimp form

Screen Shot 2020-11-11 at 9 05 53 AM

https://github.com/ShelterTechSF/sheltertech.org/commit/5d63d660c60006960dae64876958958717ce432b

This does a direct embedding of the Mailchimp forms using the snippet that we get directly from Mailchimp. Initially, I thought that this would be more maintainable, since it allows the non-technical teams to just send the technical team a direct copy of the snippets from Mailchimp, but this got a bit messier than I expected with overriding styles, and I also realized that the snippets just directly contain the input fields, rather than totally generating them at runtime with JavaScript.

Let me at least explain what's going on in the current approach, regardless of what we choose to do with this in the future. The first is that I am using [React's `dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) prop, which allows you to embed any raw HTML as a string, including in-document `<style>` or `<script>` tags. Next, I have two different CSS files: a CSS modules file and then a non-module, global CSS file. The former is for the title and the description of the form, which do not come with the Mailchimp embed code, and the latter is for overriding the styles from Mailchimp.

I made liberal use of `!important`, since Mailchimp has some pretty high-specificity styles, although I probably could have increased my own specificity by adding an extra global CSS class to wrap the whole thing. I had to duplicate our site styles for the buttons and input elements, since I had to make them apply to and override Mailchimp's styles.

## Adding the modal to the Home page

Screen Shot 2020-11-11 at 9 25 08 AM

https://github.com/ShelterTechSF/sheltertech.org/commit/552348018df15a3191d314cdc75d97c88380a7c0

This was mostly straightforward, but I did have to ident almost every line in the Home page source code, since I needed to set up some React state with `useState()` to control the modal `isOpen` prop. For me, rebasing this branch on top of `main` has been a bit of a pain, since I'm constantly getting merge conflicts, but I don't think there's much I can do here until this is merged in.

## Things that still need to be addressed

1. The form submission doesn't feel great, since it does an actual page load and takes you off our site and onto Mailchimp:

Screen Shot 2020-11-11 at 9 27 13 AM

  @derekfidler, I'm guessing this is not a great UX, so maybe we really should just replace the embedded form with our own, so that we can handle the form submission entirely in JavaScript.

2. There's some weird re-rendering and pop-in that happens when the modal opens on the actual Home page. I'm guessing that for some reason, a re-render is occurring for all the elements on the base Home page, and because our images are getting reloaded, it causes the page to re-layout with the images popping in and out, shifting elements on the page.

3. The Mailchimp form styles are completely not loading in Firefox for some reason.

4. I am apparently failing some tests, which I think is related to https://github.com/reactjs/react-modal/issues/553.

### DIFF
--- a/src/components/grid-aware/Modal/Modal.module.css
+++ b/src/components/grid-aware/Modal/Modal.module.css
@@ -1,0 +1,45 @@
+.overlay {
+  background-color: rgba(0, 0, 0, 0.5);
+  bottom: 0;
+  left: 0;
+  overflow: auto;
+  position: fixed;
+  right: 0;
+  top: 0;
+}
+
+.content {
+  background: var(--color-white);
+  border-radius: 16px;
+  box-shadow: 0 12px 32px rgba(0, 0, 0, 0.2);
+  box-sizing: border-box;
+  margin: 80px auto 40px;
+  max-width: 850px;
+  outline: none;
+  padding: 50px;
+  position: relative;
+}
+
+.closeButton {
+  background: none;
+  border: 0;
+  cursor: pointer;
+  display: inline-block;
+  padding: 0;
+  position: absolute;
+  right: 30px;
+  top: 30px;
+}
+
+@media (--tablet-and-down) {
+  .content {
+    margin-left: 20px;
+    margin-right: 20px;
+    padding: 35px 25px;
+  }
+
+  .closeButton {
+    right: 25px;
+    top: 25px;
+  }
+}
--- a/src/components/thirdparty/mailchimp/VolunteerSignupForm.js
+++ b/src/components/thirdparty/mailchimp/VolunteerSignupForm.js
@@ -1,0 +1,62 @@
+import React from "react";
+
+import "./VolunteerSignupForm.css";
+import s from "./VolunteerSignupForm.module.css";
+
+/* eslint-disable react/no-danger */
+// This entire file is just about embedding an external form
+
+const rawInnerHtml = {
+  __html: `
+<!-- Begin Mailchimp Signup Form -->
+<link href="//cdn-images.mailchimp.com/embedcode/classic-10_7.css" rel="stylesheet" type="text/css">
+<style type="text/css">
+	#mc_embed_signup{background:#fff; clear:left; font:14px Helvetica,Arial,sans-serif; }
+	/* Add your own Mailchimp form style overrides in your site stylesheet or in this style block.
+	   We recommend moving this block and the preceding CSS link to the HEAD of your HTML file. */
+</style>
+<div id="mc_embed_signup">
+<form action="https://sheltertech.us19.list-manage.com/subscribe/post?u=c47829732a0bea5c8e8a94604&amp;id=08f60e42ef" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" class="validate" target="_blank" novalidate>
+    <div id="mc_embed_signup_scroll">
+
+<div class="mc-field-group">
+	<label for="mce-EMAIL">Email Address </label>
+	<input type="email" value="" name="EMAIL" class="required email" id="mce-EMAIL">
+</div>
+<div class="mc-field-group">
+	<label for="mce-FNAME">First Name </label>
+	<input type="text" value="" name="FNAME" class="" id="mce-FNAME">
+</div>
+<div class="mc-field-group">
+	<label for="mce-LNAME">Last Name </label>
+	<input type="text" value="" name="LNAME" class="" id="mce-LNAME">
+</div>
+<div class="mc-field-group">
+	<label for="mce-MMERGE12">Volunteer Interests </label>
+	<input type="text" value="" name="MMERGE12" class="" id="mce-MMERGE12">
+</div>
+	<div id="mce-responses" class="clear">
+		<div class="response" id="mce-error-response" style="display:none"></div>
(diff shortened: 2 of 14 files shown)


src/components/thirdparty/mailchimp/VolunteerSignupForm.js
```diff
@@ -0,0 +1,62 @@
+import React from "react";
+
+import "./VolunteerSignupForm.css";
+import s from "./VolunteerSignupForm.module.css";
+
+/* eslint-disable react/no-danger */
+// This entire fire is just about embedding an external form
```
davidagustin, line 7: This entire file

davidagustin: This entire "file"?

src/components/grid-aware/Modal/Modal.module.css
```diff
@@ -0,0 +1,45 @@
...
+  padding: 0;
+  position: absolute;
+  right: 30px;
+  top: 30px;
+}
+
+@media (--tablet-and-down) {
+  .content {
+    margin-left: 20px;
+    margin-right: 20px;
```
davidagustin, line 37: Possible shorthand can be used?

src/components/grid-aware/Modal/Modal.module.css
```diff
@@ -0,0 +1,45 @@
+.overlay {
+  background-color: rgba(0, 0, 0, 0.5);
+  bottom: 0;
+  left: 0;
+  overflow: auto;
+  position: fixed;
+  right: 0;
+  top: 0;
```
davidagustin, line 8: I assume you tried other methods in completely covering the whole view of the browser with the darkened hue of the modal  but is there a shorter way to cover the whole screen than putting all bottom, top, left and right positions to 0?

src/components/thirdparty/mailchimp/VolunteerSignupForm.js
richardxia: Good catch! (Though the file kind of looks like a fire with all the inline HTML) 🔥 Will fix it now.

src/components/grid-aware/Modal/Modal.module.css
richardxia: I actually didn't try anything else because I had copied these styles over from the default ones that the library comes with. I'm not sure if there's a shorter way to cover it, since it's `position: fixed`. You still need the `top: 0` since the element itself is placed based on where its parent is, not based on the page. The `right` and the `bottom` could be replaced with `width: 100vw` and `height: 100vh`, which doesn't save us much because it's a one-for-one tradeoff.

You could possibly leave out `left:`, but it will only work if you don't have anything that pushes the modal's parent's position away from the left edge, which makes it a little bit riskier to do.

Overall, I think setting the position across all four dimensions is pretty much as compact and reliable as we're going to get.

src/components/grid-aware/Modal/Modal.module.css
richardxia: I didn't want to override the top and bottom margin settings from the non-media-query version of this, which are set to 80px and 40px respectively.

davidagustin (approved):
Great fix!

PR #97, opened 2020-11-11
1705 create covid 19 info box component

ivanleeswe:
Create Covid19InfoBox Component

Create styling for Covid19InfoBox Component


src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.module.css
```diff
@@ -0,0 +1,30 @@
...
+
+@media (--tablet-and-down) {
+  .bleedWrapper {
+    display: block;
+    padding: 50px 40px;
+  }
+}
+
+.bleedMainContent {
+  grid-column: 2;
```
davidagustin, line 18: Screen Shot 2020-11-15 at 3 08 46 PM

I see a right bleed here. I don't think this component is just in column 2 but extends also to touch column 3

davidagustin (changes requested):

davidagustin (changes requested):
The left and right bleed columns are small and the component in the figma does not have any overlap into the left and right columns

Screen Shot 2020-11-15 at 3 53 51 PM

Screen Shot 2020-11-15 at 3 50 30 PM

Screen Shot 2020-11-15 at 3 49 45 PM

ivanleeswe:
> The left and right bleed columns are small and the component in the figma does not have any overlap into the left and right columns
> 
> Screen Shot 2020-11-15 at 3 53 51 PM
> 
> Screen Shot 2020-11-15 at 3 50 30 PM
> 
> Screen Shot 2020-11-15 at 3 49 45 PM

Pair programmed to fix padding by moving it down from bleedWrapper property to bleedMainContent property.

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./Covid19InfoBoxBlock.module.css";
+
+const Covid19InfoBoxBlock = ({ title, description }) => {
```
richardxia: Nit: Let's capitalize this as `COVID19InfoBoxBlock`. With camelCase, there different styles for how to handle acronyms, but I have a personal preference for making acronyms all caps. [This also matches Airbnb's style guide](https://github.com/airbnb/javascript/tree/f0df3a8680479ff0b897cd98a1eab6b156899214#naming--Acronyms-and-Initialisms), which we mostly follow and have configured in our linters.

```suggestion
const COVID19InfoBoxBlock = ({ title, description }) => {
```

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.module.css
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  .bleedMainContent {
+    padding: 50px 40px;
+  }
+}
+
+.title {
+  font: var(--font-title);
+  font-size: 30px;
```
richardxia: Let's remove this, since it's already set by the `font:` shorthand above. We generally don't need to manually set the font size, font family, font weight, or line-height, since our [global font variables](https://github.com/ShelterTechSF/sheltertech.org/blob/ffe2ad112811ae214664cdc80990a9c03dd1c142/src/stylesheets/global.css#L54-L62) should already cover them.

Edit: Actually, I see why you put this here, since in mobile, we still want this to be 30px tall, but in mobile, `--font-title` is smaller. Could you instead create a media query and set the font to `--font-headline` in mobile, which has a size of 30px?

```suggestion
```

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.module.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.description {
+  font: var(--font-body-medium);
+  font-size: 20px;
```
richardxia: Same with this, except it doesn't look like we have an appropriate font typography in mobile, so instead, could you only set this `font-size` under a media query?

```suggestion
```

richardxia (commented):
This looks pretty good, aside from a few small comments from me! Do you think you could also add this component to the appropriate place on the Volunteer page?

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.js
ivanleeswe: 👍 Good to know about caps on acronyms. Changed to COVID19InfoBoxBlock.

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.module.css
ivanleeswe: 👍 Created media query using font headline for title.

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.module.css
ivanleeswe: 👍 Deleted and moved to media query.

src/components/grid-aware/Covid19InfoBoxBlock/Covid19InfoBoxBlock.js
ivanleeswe: 👍 Add to volunteer page as well!

richardxia (approved):
LGTM!

davidagustin (approved):

ivanleeswe:
LET'S GO!

PR #98, opened 2020-11-16
1661 create title block component

ivanleeswe:
Create Title Block which consists of one line of text


src/components/grid-aware/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,39 @@
...
+    grid-template-columns: 20px 1fr 20px;
+    padding-bottom: 30px;
+    padding-top: 80px;
+  }
+}
+
+.textWrapper {
+  grid-column-end: 12;
+  grid-column-start: 1;
+}
```
richardxia: This doesn't do anything because the parent of `.textWrapper` isn't a grid parent. You can just delete these styles and also delete the `textWrapper` element from the HTML as well.

src/components/grid-aware/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,39 @@
...
+
+.title {
+  font: var(--font-headline);
+  font-size: 50px;
```
richardxia: Can you remove the font size? You shouldn't need to set it because the font shorthand already takes care of it.

src/components/grid-aware/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+
+@media (--tablet-and-down) {
+  .title {
+    font-size: 30px;
+  }
+}
```
richardxia: You can get rid of this media query altogether because you don't need to explicitly set the font size, as the variables themselves are already set to the appropriate size in mobile as well.

richardxia (commented):
Looks pretty good! If you address the few things that I commented on, as well as fixing the lint error, then this looks good to me for merging!

src/components/grid-aware/TitleBlock/TitleBlock.module.css
ivanleeswe: 👍  Deleted.

src/components/grid-aware/TitleBlock/TitleBlock.module.css
ivanleeswe: 👍  Deleted.

src/components/grid-aware/TitleBlock/TitleBlock.module.css
ivanleeswe: 👍  Deleted.

richardxia (approved):
LGTM!

ivanleeswe:
Merged! 🙌

PR #99, opened 2020-11-16
Update images and texts

candywang:
I updated the texts and did a quick comparison overview with the [design here](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1601).

Here are some differences I've found:

## Home Page - Web
| Figma  | Gatsby |
| ------------- | ------------- |
| "Programs" text color is Gray 700 | "Programs" text color is ShelterTech Blue, button height looks short |
| fhnav  | ghnav |
| Two images with button below each paragraph block | Three images with buttons grouped together |
| rf1 | rg1 |
| Text wrapping is tighter and black container is a bit taller and narrower | Text wrapping and black container is a bit shorter wider |
| fd1 | gd1 |

## Volunteer Page - Web
| Figma  | Gatsby |
| ------------- | ------------- |
| "Work With Us" text color is Gray 700 | "Programs" text color is ShelterTech Blue and text should be "Work With Us", button height looks short |
| fvnav | gvnav |
| Text wraps after "Partners"  | Text wraps after "and"  |
|  fc3 | gc3 |

## Volunteer Page - Mobile
| Figma  | Gatsby |
| ------------- | ------------- |
| Left side of image aligns with text blocks  | Left side of image has extra spacing  |
| fc2 | gc2 |
| Image looks resized based on viewport | Image currently resizing to max-width |
| fc1 | gc1 |

## Other details
| Figma  | Gatsby |
| ------------- | ------------- |
| Button height looks taller | Button height looks shorter |
| fig | gat |

For a few components, the button text for web vs mobile is different. I commented on this issue in various Figma design areas and hope to receive responses soon so that I can further update our code appropriately!


src/pages/new/index.js
```diff
@@ -111,9 +111,9 @@ export default () => (
       quotes={[
         {
           quote:
-            "We've heard from San Franciscans, including as part of our recent strategic planning community outreach that it's hard to know where to go when you're looking for supportive services. It's hard to know where to start, and hard to navigate different resources to find accurate information. We're glad to be supporting ShelterTech and the broader community to build SF Service Guide - a one-stop, reliable place to help connect residents to the services they need.",
+            "We’ve heard from San Franciscans, including as part of our recent strategic planning community outreach that It’s hard to know where to go when you’re looking for supportive services. It’s hard to know where to start, and hard to navigate different resources to find accurate information. We’re glad to be supporting ShelterTech and the broader community to build SF Service Guide – a one-stop, reliable place to help connect residents to the services they need.",
```
richardxia: You ~accidentally~ capitalized one of the "It's" in the quote. It's the first one, in the middle of:

> strategic planning community outreach that <b>It’s</b> hard to know

Double checking the Figma design, I see that it also has the capitalization error. Can you leave the the "it's" uncapitalized here, since I'm pretty sure it's a typo in the design?

src/pages/new/index.js
```diff
@@ -172,7 +172,7 @@ export default () => (
     <ArticleSpotlightCard
       eyebrowText="Partnership Spotlight"
       title="Digital Equity Team of San Francisco"
-      description="Free Fiber Initiative summary we're working together to find new and different ways of building lasting infrastructure"
+      description="Free Fiber Inititiative summary we’re working together to find new and different ways of building lasting infrastructure."
```
richardxia: The original spelling of "Initiative" was correct. I see that you've copied this from the Figma design, but in this case, there's a typo there. Could you revert the spelling of that word back to the correct one?

richardxia (approved):
Wow, thank you for the detailed analysis and for putting that all in a table in Markdown! I think your text changes are mostly fine, but I'd suggest we don't reintroduce typos that are present in the design but that we had corrected in our implementations. I'll preemptively approve this PR, assuming that you make the text changes before merging.

For the rest of the things you discovered, here are my comments:

> "Programs" text color is ShelterTech Blue

This is intentional, since the "active page" is supposed to be highlighted in blue, and the Programs link is meant to take you to the Home page (at least for now).

Regarding the button height, on my computer they look fine. What browser are you using? It looks fine for me on Firefox, Chrome, and Safari.

Screen Shot 2020-11-15 at 10 01 32 PM

### Three-Paragraph Block
> Three images with buttons grouped together

Yeah, looks like the design changed since when we first implemented it. I created a separate ticket to track it, but if you're down for it, you can take it: https://app.clubhouse.io/sheltertech/story/1735/update-three-paragraph-component-to-match-latest-designs

### Article Spotlight
> Text wrapping and black container is a bit shorter wider

For this, I think I intentionally OK'd what we had, since the current Figma design is not aligned with the grid columns. Originally, the site had a different number of grid columns (15), so it may have been aligned to that. I think we can stick with what we have for now, unless if someone from design asks that we change the size of the card.

### Partners and Sponsors Block
> Text wraps after "and"

I'm not sure if it'd be possible to get it to wrap after "Partners" based on the current grid. If we make it three grid columns wide, it'll force the whole title to take up three separate lines, since "and sponsors" would also wrap after "and". I think what we have is probably fine, since I don't want us to get into manually inserting line breaks unless if it's necessary.

### Two-Paragraph Block
> Left side of image has extra spacing

I think that's going to be a bit tricky to set up, but we could try to do it, now that we've gotten more of the higher priority items out of the way. I created https://app.clubhouse.io/sheltertech/story/1736/get-two-paragraph-block-image-on-mobile-to-line-up-with-text to track this.

I think the way I would suggest doing it is to remove the outer grid on mobile and change the "inner" grid to behave like the outer 3-column grid. Then you can have the image aligned to columns 2 and 3.

### Image Header
> Image currently resizing to max-width

Yeah, I remember us having a lot of difficulty with this. I think we can tweak it to look more similar to the design on a 375px-wide screen, but it'll be difficult to have this image work well across all the different resolutions we want to support. Created a separate ticket in https://app.clubhouse.io/sheltertech/story/1737/tweak-image-header-image-alignment-on-mobile.

src/pages/new/index.js
candywang: Yes! Thanks for noticing this!!

src/pages/new/index.js
candywang: Yes, sorry!

candywang:
Thank you for the quick review and for creating new tickets to track the changes! I can definitely take these tickets and start making the updates.

>the "active page" is supposed to be highlighted in blue

This makes perfect sense.

>What browser are you using?

I'm using Chrome. Could it be my WebKit not working properly? Moving forward, I'll check other browsers on top of Chrome.

richardxia:
> I'm using Chrome. Could it be my WebKit not working properly? Moving forward, I'll check other browsers on top of Chrome.

I would find that to be unlikely. @davidagustin, @ivanleeswe, do your buttons look like @cw323's?

I can also double check on my end, since I didn't do an exact measurement, and now I'm second guessing my ability to just eyeball this.

davidagustin (approved):
Good eye!

davidagustin:
> > I'm using Chrome. Could it be my WebKit not working properly? Moving forward, I'll check other browsers on top of Chrome.
> 
> I would find that to be unlikely. @davidagustin, @ivanleeswe, do your buttons look like @cw323's?
> 
> I can also double check on my end, since I didn't do an exact measurement, and now I'm second guessing my ability to just eyeball this.

It does not

candywang:
I've compared and updated the main and volunteer pages. I will conduct a final review once all the components are instantiated. Thanks!

PR #100, opened 2020-11-16
Modularize Image and Title/Description PropTypes, update code

candywang:
I created a separate file for PropTypes that have the shape of `url, alt` and `title, description`. I named them `ImagePropType` and `TitleDescriptionPropType`, respectively, and components have been updated to use these PropTypes.

The `TitleDescriptionPropType` name is a bit long, perhaps it can be shorter.


src/PropTypeShapes.js
```diff
@@ -0,0 +1,11 @@
+import PropTypes from "prop-types";
```
richardxia: This is a comment on the file name, not this line, but I can't leave a comment on the file name itself.

I'm not actually sure what the most common conventions are here, but I feel that it might be better to give this file a kebab-case name, like `prop-types.js`. The reason behind my thinking is that in most of our files that are `CamelCase.js`, we have a default export of a class whose name matches the file name, so the file name describes the type of thing being exported.

In this file, we don't have a natural default export, so we're only providing named exports. We don't actually have a class named `PropTypeShapes`, so it might be strange to name the file after that.

What do you think?

src/PropTypeShapes.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const ImagePropType = PropTypes.shape({
```
richardxia: Something I've been thinking about is whether we actually want to define these with `PropTypes.shape()` wrapped around it. When we do this, we lose the ability to reference the individual fields outside of the `shape()`, since they come "prewrapped" with `shape()`.

For a concrete example, take `ParagraphBlock` in `ThreeParagraphBlock`. Its prop types are the usual `title` and `description` prop type that we have, but because its props are directly listed there, we can't make use of the `TitleDescriptionPropType` because it comes with a `shape()`.

```js
// What we have today
ParagraphBlock.propTypes = {
};

// Invalid, since this is actually short for PropTypes.shape({ title: ..., description: ...})
ParagraphBlock.propTypes = TitleDescriptionPropType;
```

If we instead defined `TitleDescriptionPropType` as just the raw object without the `PropTypes.shape()`, then we could reuse it here while still retaining the ability to wrap it with a `shape()` in the places that we need it. The downside to this approach is that the other places become more verbose.

Another benefit to this approach of using plain objects is that we could actually mix multiple objects together. I wouldn't recommend doing this very often, since it could get hard to read, but in theory, we could write things like this:

```js
const MyImageWithText = () => {}

MyImageWithText.propTypes = {
  ...ImagePropType,
  ...TitleDescriptionPropType,
};
```

What do you think?

src/legacy/layouts/Layout.js
```diff
@@ -2,6 +2,7 @@ import { StaticQuery, graphql } from "gatsby";
 import PropTypes from "prop-types";
 import React from "react";
 import { Helmet } from "react-helmet";
+import { TitleDescriptionPropType } from "../../PropTypeShapes";
```
richardxia: Let's avoid touching the stuff under `legacy`, since we'll be deleting it soon, once we do the real site launch.

src/PropTypeShapes.js
candywang: I agree! I'll change the file name to `prop-types.js`.

candywang:
I am closing this PR due to receiving false positives from ESLint when destructuring variables that contain a raw object defining a PropType.

The goal was to create a separate file `prop-types.js` defining common PropTypes as raw objects to be reused throughout our components. However, when these variables are imported into a component page and destructured within a PropTypes definition, we run into ESLint errors. With the correct `eslint-plugin-react` version, the issue itself seems like it solely pertains to the fact that we are using destructuring where [ESLint thinks that the imported PropTypes definition are components of its own and needs props of its own](https://github.com/yannickcr/eslint-plugin-react/issues/2357). After some research and discussion, this issue can be temporarily fixed by not using destructuring but this would partially defeat the purpose of creating a separate file for common PropTypes. Since the priority of this assignment is currently low, we may consider looking into implementing TypeScript in the near future as a more elegant way to validate props.

### Examples
---
##### Defining a raw object in `prop-types.js`
```javascript
export const TitleDescriptionPropType = {
};
```

##### Using `TitleDescriptionPropType` in `ExampleComponent.js`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;_Without destructing, ESLint **will not** throw an error_
```javascript
import { TitleDescriptionPropType } from '../../../prop-types.js';

ExampleComponent.propTypes = {
  header: PropTypes.shape(TitleDescriptionPropType).isRequired,
  image: PropTypes.shape({
    url: PropTypes.string,
    alt: PropTypes.string,
  }),
  paragraph: PropTypes.string.isRequired,
};
```

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;_With destructing, ESLint **will** throw an error_
```javascript
import { TitleDescriptionPropType } from '../../../prop-types.js';

// Defining shape of a Paragraph using destructuring of TitleDescriptionPropType
const ParagraphPropType = PropTypes.shape({
  ...TitleDescriptionPropType,
 paragraphText: PropType.string.isRequired,
});

ExampleComponent.propTypes = {
  header: PropTypes.shape(TitleDescriptionPropType).isRequired,
  image: PropTypes.shape({
    url: PropTypes.string,
    alt: PropTypes.string,
  }),
  paragraph: ParagraphPropType.isRequired, // Using ParagraphPropType
};
```

richardxia:
Sorry about sending you off on wild goose chase, but I hope it was at least educational!

candywang:
No worries at all! It definitely deepened my learning of PropTypes and using effective search terms 😊

PR #101, opened 2020-11-16
Fix image alignment on mobile view for TwoParagraphBlock component

candywang:
I used negative margins to stick the image to the right of the screen but as the viewport shrinks it'll align the image with the text and resize accordingly.

### DIFF
--- a/src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
+++ b/src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
@@ -84,6 +84,7 @@
   .gridAreaImage {
     margin-right: -40px;
     margin-top: 40px;
+    text-align: right;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -123,7 +124,6 @@
   }
 
   .image1Wrapper {
-    left: 40px;
     margin-bottom: 0;
     position: relative;
```
richardxia: Since you're not setting any of the left/top/right/bottom positions anymore, you don't need this `position: relative` either.

```suggestion
```

richardxia (approved):
Nice, this is pretty creative, and I think it works well here!

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
candywang: Thanks for catching this!

PR #102, opened 2020-11-16
Remove roles per volunteer team request

derekfidler:
Jasmine requested we remove these two roles from the current page.


richardxia (approved):
LGTM!

PR #103, opened 2020-11-16
Update ThreeParagraphBlock component with latest design

candywang:
I updated the layout, images, and texts for the web and mobile view to match the latest design.


src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -62,10 +62,7 @@
   column-gap: var(--grid-column-gap);
   display: grid;
   grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
-  grid-template-rows: repeat(2, auto);
-  padding-bottom: 100px;
-  padding-top: 100px;
-  row-gap: 50px;
+  padding: 100px 0 98px;
```
richardxia: How about we leave the bottom padding as 100px? Even if the design shows 98px, I think that may just be a small inconsistency, and 100px would be almost visually identically yet more consistent in the code.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -148,22 +126,19 @@
...
   .image1Wrapper {
     display: none;
   }
 
   .image2Wrapper {
-    display: none;
-  }
-
-  .image3Wrapper {
+    /* overflow: hidden; */
```
richardxia: Let's delete commented out code instead of committing it. See https://kentcdodds.com/blog/please-dont-commit-commented-out-code for a short article describing why keeping commented out code in the codebase is not generally a good idea.

richardxia (commented):
Looking at the designs for the Home page and the Volunteer page, the Volunteer page actually still uses the original design for the ThreeParagraphBlock. I think we'll need to simultaneously support both use cases.

I think this isn't too difficult to do, so how about we do the following?

- Keep the `ctaButtons` prop as describing the extra row at the bottom, but make it optional, along with the `ctaTitle`? If they aren't present, we don't render that bottom grid area at all.
- Add a `button` prop to `paragraph1`, `paragraph2`, and `paragraph3`, which is also optional. This would match what you've already changed about the `<ParagraphBlock>` subcomponent, and it would allow us to optionally render a CTA button with each paragraph.
- Make image3 optional

src/pages/new/volunteer/index.js
```diff
@@ -5,6 +5,10 @@ import ImageHeader from "../../../components/grid-aware/ImageHeader";
 import classroom from "../../../components/grid-aware/ImageHeader/stories/classroom.png";
 import whiteboard from "../../../components/grid-aware/ImageHeader/stories/whiteboard-sticky-notes.png";
 import StatsBlock from "../../../components/grid-aware/StatsBlock";
+import ThreeParagraphBlock from "../../../components/grid-aware/ThreeParagraphBlock";
+import teamActivityImage from "../../../components/grid-aware/ThreeParagraphBlock/stories/image4.png";
+import pairProgrammingImage from "../../../components/grid-aware/ThreeParagraphBlock/stories/image5.png";
+import teamSnackingImage from "../../../components/grid-aware/ThreeParagraphBlock/stories/image6.png";
```
richardxia: Actually, since these images are specific to this instance of the component on this page, how about we move them here to this directory?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -19,16 +33,39 @@ const ImagePropType = PropTypes.shape({
...
 /* Subcomponents */
 
-const ParagraphBlock = ({ title, description }) => (
-  <div>
-    <div className={s.paragraphTitle}>{title}</div>
-    <div className={s.paragraphDescription}>{description}</div>
-  </div>
-);
+const ParagraphBlock = ({ title, description, button }) => {
+  let hasButton;
```
richardxia: `hasButton` is a bit of a misleading name for this variable, since it makes it sound boolean. A common naming convention across many languages is to name boolean variables using grammatical [predicates](https://en.wikipedia.org/wiki/Predicate_(grammar)), most commonly `hasFoo` or `isFoo`.

Since this variable actually (possibly) holds a real thing, we should just name it after the thing. How about `buttonWrapper`?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -19,16 +33,39 @@ const ImagePropType = PropTypes.shape({
...
+
+  if (button) {
+    hasButton = (
+      <div className={s.ctaButton}>
```
richardxia: Since there are several different kinds of CTA buttons on the page, how about we give this CSS class a more specific name, like `paragraphCTAButton`, to make it easier to differentiate between the different instances?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -58,20 +95,41 @@ const ThreeParagraphBlock = ({
...
+      <div className={s.leftTopImageWrapper}>
+        <img
+          className={s.image}
+          src={leftTopImage.url}
+          alt={leftTopImage.alt}
+        />
+      </div>
+    );
+
+    leftBottomImageWrapperClassName += ` ${s.v1}`;
```
richardxia: Could we give the CSS class a more descriptive name? It'll be difficult to remember which is `v1` and `v2`. How about `solo` and `paired`? Feel free to come up with your own names, too.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -99,29 +159,37 @@ const ThreeParagraphBlock = ({
...
       </div>
     </div>
   );
 
-  const GridAreaBottom = () => (
-    <div className={s.gridAreaBottom}>
-      <CTABlock title={ctaTitle} buttons={ctaButtons} />
-    </div>
-  );
+  let GridAreaBottom;
```
richardxia: I think we should reserve `CapitalCamelCased` names for things that act as classes (including React function components). The change you made here causes it to be a normal variable that happens to hold JSX content. I think you have two options:

1. Rename this to `gridAreaBottom`
2. Keep the definition as a React function component with no props, and put the conditional inside of the body, kind of like this:

```jsx
const GridAreaBottom = () => {
  if (!ctaTitle || !ctaButtons) {
    return null;
  }
  return (
    <div>...</div>
  );
}

File: src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css [hunk 8]
```diff
@@ -76,8 +79,11 @@
 @media (--tablet-and-down) {
   .gridParent {
     display: block;
-    padding-bottom: 80px;
-    padding-top: 40px;
+    padding: 50px 0 80px;
+  }
+
+  .gridParent.v2 {
+    padding: 40px 0 80px;
```
richardxia: I think this isn't actually necessary, since it looks like both versions have 50px top padding. This may have very subtly changed in the designs between when we first implemented the component and now.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -62,9 +62,12 @@
   column-gap: var(--grid-column-gap);
   display: grid;
   grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding: 100px 0;
+}
+
+/* Grid parent for optional ctaButtons block */
+.gridParent.v2 {
```
richardxia: Is there a way for us to get rid of this variant altogether? I think we could write the CSS in a way such that just the presence or absence of the `<GridBottomArea>` will be sufficient for laying it out properly.

I think we can do this by getting rid of the `row-gap` and just having the `gridAreaBottom` set a top margin. That way, if it doesn't exist, then the extra spacing isn't there.

candywang:
I reverted the original code that @richardxia created and added the following optional props: `leftTopImage`, `ctaTitle`, `ctaButtons`, and `button` within `paragraph1`, `paragraph2`, and `paragraph3`. The `propTypes` validation of `.isRequired` has been removed from these optional props with default values being set.

Since the `<ThreeParagraphBlock />` component is instantiated on both the Home page (with only the optional `button` props for each paragraph) and on the Volunteer page (with only the optional `leftTopImage`, `ctaTitle`, and `ctaButtons` props), conditional statements have been added and will render the appropriate block of code with the updated styling if any of these optional props are being passed in.

- **WIP**:
  - I noticed when `leftTopImage` is being passed in and I manually resize the screen to be more narrow, the image overlaps with the second column (`GridAreaMiddle`). I need to figure out a way to fix this.
  - In order to render the bottom row where `ctaTitle` and `ctaButtons` exists, both props must be passed in at the same time. Would we want this to be changed to making them both independently optional?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -148,22 +162,26 @@
...
-  .image2Wrapper {
+  .leftBottomImageWrapper {
     display: none;
   }
 
-  .image3Wrapper {
+  .rightImageWrapper {
+    margin-right: -40px;
     margin-top: 40px;
-    overflow: hidden;
```
richardxia, line 166: I think if we take this out, it causes the whole page to scroll if this thing expands off screen. That said, I think we actually need to go through and just set `overflow-x: hidden` on all the bleed wrappers, since that would allow us to enforce this for all content under each bleed wrapper, rather than individually doing this for each thing that may possible spill off screen.

src/pages/new/volunteer/index.js
```diff
@@ -57,6 +61,38 @@ export default () => (
...
+        alt: "Multiple volunteers working at a datathon.",
+      }}
+      leftBottomImage={{
+        url: teamActivityImage,
+        alt: "Team posing for a photo after a design workshop.",
+      }}
+      rightImage={{
+        url: teamSnackingImage,
+        alt: "Multiple volunteers working at a datathon.",
+      }}
```
richardxia: I think the top left and bottom right images are swapped, except I'm not sure if the file names are swapped or if the code is swapped.

richardxia (commented):
Overall, this looks like it's in the right direction. I left a few small comments, but I think there are no major issues here. Thanks for hoping on this!

richardxia:
>   * I noticed when `leftTopImage` is being passed in and I manually resize the screen to be more narrow, the image overlaps with the second column (`GridAreaMiddle`). I need to figure out a way to fix this.

I think you can set `max-width: 100%` on the image.

>   * In order to render the bottom row where `ctaTitle` and `ctaButtons` exists, both props must be passed in at the same time. Would we want this to be changed to making them both independently optional?

That sounds reasonable to me. It'd probably be more intuitive to someone who is passing in the props that if either prop is passed in, then it should render the row.

src/pages/new/volunteer/index.js
candywang: That would be better! Should I create an `image_assets` directory within the `src/pages/new/volunteer` directory?

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
candywang: I like `buttonWrapper`. Thanks!

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
candywang: I like `solo` and `paired`. Thanks!

src/pages/new/volunteer/index.js
candywang: You're right! I also need to change the `alt` text. Thanks for catching this!

src/pages/new/volunteer/index.js
richardxia: Since I don't anticipate having too many files in there, why don't we just leave it directly in the `volunteer/` directory? There's no need to add extra hierarchy unless we actually need it.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
candywang: I like #2, it looks cleaner.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
candywang: I added `overflow-x: hidden` to the bleed wrapper. Do I still need to set `overflow: hidden;` to `.leftBottomImageWrapper.solo` if I have `.image` set to `max-width: 100%;`?

richardxia: Let's not worry about that for now. Technically, it could overflow if the minimum width is greater than 100%, and also possibly in a few other cases, but it's unlikely to happen in our situation.

candywang: Okay, I'll keep the `overflow` property. Thanks!

src/pages/new/index.js
```diff
@@ -63,35 +62,28 @@ export default () => {
           title: "Volunteer",
           description:
             "Volunteers make our work possible. There are several ways to support our mission. Learn more and get involved.",
+          button: { text: "Become a Volunteer", internalLink: "/foo" },
```
davidagustin: You can use the /new/volunteer

src/pages/new/index.js
```diff
@@ -63,35 +62,28 @@ export default () => {
...
           title: "Partnerships",
           description:
             "We work with companies, nonprofits, and local governments to empower the community. Reach out to us.",
+          button: { text: "Work With Us", internalLink: "/bar" },
         }}
         paragraph3={{
           title: "Donate",
           description:
             "Our programs are largely funded by donations from people who care about bridging the digital divide. Support ShelterTech today.",
+          button: { text: "Donate", internalLink: "/fooBar" },
```
davidagustin: /new/donate here also

davidagustin (changes requested):

davidagustin (commented):
Don't know what should go here

src/pages/new/index.js
```diff
@@ -63,35 +62,28 @@ export default () => {
           title: "Volunteer",
           description:
             "Volunteers make our work possible. There are several ways to support our mission. Learn more and get involved.",
+          button: { text: "Become a Volunteer", internalLink: "/foo" },
         }}
         paragraph2={{
           title: "Partnerships",
           description:
             "We work with companies, nonprofits, and local governments to empower the community. Reach out to us.",
+          button: { text: "Work With Us", internalLink: "/bar" },
```
richardxia: Similar to @davidagustin's comments, this button should open up the volunteer modal like before.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -10,6 +10,20 @@ import s from "./ThreeParagraphBlock.module.css";
...
+    }),
+    PropTypes.exact({
+      text: PropTypes.string,
+      internalLink: PropTypes.string,
+    }),
+    PropTypes.exact({
+      text: PropTypes.string,
+      onClick: PropTypes.func,
+    }),
+  ]),
```
davidagustin, line 26: There will be a common PropType file right? @richardxia @cw323  so will need to change this in the future then.

davidagustin (approved):

richardxia (approved):
LGTM!

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
candywang: Yep! I'll be working on that after the `OneParagraphBlock` component.

PR #104, opened 2020-11-17
Add broader set of viewports in Storybook

richardxia:
This has been bothering me for a while, but we haven't had a viewport in Storybook that actually matched the size of the mobile designs in Figma, which is 375px. For this PR, I followed the instructions in https://storybook.js.org/docs/react/essentials/viewport#use-detailed-set-of-devices to add a larger set of device viewports. You can see what this expands to in the following screenshot:

Screen Shot 2020-11-16 at 8 49 00 PM

The constant that I imported in this PR is defined [here](https://github.com/storybookjs/storybook/blob/5a257808c3fdf03b34c47bded5b9d1472e03e17e/addons/viewport/src/defaults.ts#L3), so if we feel that this is too overwhelming, I can cut it down to a smaller set of device types. Also, if we feel that having to think about the actual device screen names is too much, I could go back to the original list but just add a "Medium mobile" viewport size that has a width of 375px.

Let me know what you think!

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,4 +1,5 @@
 import { action } from "@storybook/addon-actions";
+import { INITIAL_VIEWPORTS } from "@storybook/addon-viewport";
 
 import "../src/stylesheets/global.css";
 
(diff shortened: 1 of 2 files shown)


davidagustin (approved):
Much needed, thankyou

PR #105, opened 2020-11-18
Create Spacer component and correct spacings

davidagustin:
Since grid-aware components are reusable, there needs to be customized spacing between grid-aware components when needed.

This Figma specification request a spacing of 100px between these two grid aware components
Screen Shot 2020-11-18 at 12 52 33 AM

Before the Spacer:
Screen Shot 2020-11-18 at 12 49 34 AM

After the Spacer:
Screen Shot 2020-11-18 at 12 49 12 AM

For this first commit I have specified 2 required inputs and 1 optional for the spacer component
Screen Shot 2020-11-18 at 12 56 46 AM

This commit uses props that influence the value of the css variables. I have made a few subtle corrections on the spacings
For example:
The padding of the bottom grid-aware component should not touch the overlapping picture of the top grid aware component
Screen Shot 2020-11-18 at 12 32 58 AM

I made the measurements:
Screen Shot 2020-11-18 at 12 32 35 AM

And corrections:
Screen Shot 2020-11-18 at 1 07 15 AM

Screen Shot 2020-11-18 at 12 31 36 AM

More spacings between grid-aware components may need to be implemented in the future and some may have been possibly overlooked.

This PR can potentially addresses the ticket:
https://app.clubhouse.io/sheltertech/story/1706/fix-up-spacing-for-the-article-spotlight-on-the-home-page


src/components/grid-aware/Spacer/Spacer.js
```diff
@@ -0,0 +1,29 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./Spacer.module.css";
+
+// color should be a value from global.css
```
richardxia: This comment probably doesn't need to be here. This is stating something that is generally true across our codebase and isn't specific to this component.

Also, there's nothing that functionally prevents you from using using a color that's not defined in globals.css. Coding style guidelines should go into style guides, not in the code like this.

src/components/grid-aware/Spacer/Spacer.js
```diff
@@ -0,0 +1,29 @@
...
+
+const Spacer = ({ heightDesktop, heightMobile, color }) => (
+  <div
+    style={{
+      "--spacer-height-desktop": `${heightDesktop}`,
```
richardxia: Please don't use string variable interpolation in a string that doesn't have anything else in it. You can just directly use the variable:

```suggestion
      "--spacer-height-desktop": heightDesktop,
```

src/components/grid-aware/Spacer/Spacer.js
```diff
@@ -0,0 +1,29 @@
...
+);
+
+Spacer.propTypes = {
+  heightDesktop: PropTypes.string.isRequired,
+  heightMobile: PropTypes.string.isRequired,
+  color: PropTypes.string,
+};
+
+Spacer.defaultProps = {
+  color: "#fff", // default color is white
```
richardxia: Actually, I think the default color should be `transparent`, not white.

Also, you're not following your own comment about only using the color variables defined in `globals.css` 😛

src/pages/new/index.js
```diff
@@ -121,6 +122,7 @@ export default () => {
           },
         ]}
       />
+      <Spacer heightDesktop="53px" heightMobile="19px" color="#000" />
```
richardxia: Shouldn't this be `color="var(--color-black)"`, according to your comment about using the variables we defined?

Also, where did these numbers come from? I'm counting 191 pixels between the bottom of the program card text to the top of the Video, and with 100px of padding within the Program Block itself, this should be 91px tall?

richardxia: As a completely unrelated comment about this particular spacer, it looks like we need to do something a bit more special to cover the whitespace under the Video Spotlight Block, behind the card on the right.

src/components/layout.js
```diff
@@ -35,6 +36,7 @@ const Layout = ({ children }) => {
         />
         {children}
       </div>
+      <Spacer heightDesktop="170px" heightMobile="80px" />
```
richardxia: I don't understand what this is for? There's nothing like this in the design, and I don't think we should be adding space to the Layout component like this because you won't be able to control it/undo it on a per page basis.

src/pages/new/index.js
```diff
@@ -131,6 +133,7 @@ export default () => {
         }}
         playButtonLink="/foo"
       />
+      <Spacer heightDesktop="42px" heightMobile="0" />
```
richardxia, line 140: This just feels like it should be part of the Video Spotlight component itself, since this appear to be 100% overlapping with the bottom section of the video background image.

src/pages/new/index.js
```diff
@@ -154,6 +157,7 @@ export default () => {
           },
         ]}
       />
+      <Spacer heightDesktop="20px" heightMobile="0" />
```
richardxia, line 164: It seems like there's an extra 10px of margin being caused by the bottom of the carousel dots. Should we correct that in that component?

Also, shouldn't the mobile height be set to 40px here, to combine with the 40px of padding in the Partners and Sponsors block in order to get a total of 80px between the carousel and the Partners and Sponsors block?

src/pages/new/volunteer/index.js
```diff
@@ -35,6 +36,7 @@ export default () => (
       title="Volunteering during COVID-19"
       description="Our programs have shifted to remote alternatives. While our means of coming together are different, our research and Datathon events will be conducted online. Simply put, our goal is still the same: to connect this community with resources that can help them address their challenges, become permanently housed, and on track to live happy and productive lives."
     />
+    <Spacer heightDesktop="120px" heightMobile="0" />
```
richardxia: Shouldn't there be 100px of spacing in desktop here and 80px of spacing in mobile?

src/pages/new/volunteer/index.js
```diff
@@ -55,6 +57,7 @@ export default () => (
         },
       ]}
     />
+    <Spacer heightDesktop="100px" heightMobile="80px" />
```
richardxia: Shouldn't there be 120px of spacing in desktop here and 0 in mobile?

src/components/grid-aware/Spacer/Spacer.js
davidagustin: ok

src/components/grid-aware/Spacer/Spacer.js
davidagustin: ok

src/components/grid-aware/Spacer/Spacer.js
davidagustin: I meant the values, not the variables

davidagustin: But I changed it

src/pages/new/index.js
davidagustin: ok

src/components/layout.js
davidagustin: ok

src/pages/new/index.js
davidagustin: setting heightDesktop to 0 and heightMobile to 40px

src/pages/new/volunteer/index.js
davidagustin: ok

src/pages/new/volunteer/index.js
davidagustin: ok

src/pages/new/index.js
davidagustin: Ill make tickets for the subtle corrections

src/pages/new/index.js
davidagustin: Ill make a ticket for the subtle correction

src/pages/new/index.js
davidagustin: Will create a ticket for it

davidagustin: https://app.clubhouse.io/sheltertech/story/1747/fix-padding-and-background-for-the-video-spotlight-component

src/pages/new/index.js
davidagustin: https://app.clubhouse.io/sheltertech/story/1747/fix-padding-and-background-for-the-video-spotlight-component

src/pages/new/index.js
davidagustin: https://app.clubhouse.io/sheltertech/story/1748/fix-spacing-in-carousel

src/components/grid-aware/Spacer/Spacer.js
```diff
@@ -0,0 +1,27 @@
...
+import React from "react";
+
+import s from "./Spacer.module.css";
+
+const Spacer = ({ heightDesktop, heightMobile, color }) => (
+  <div
+    style={{
+      "--spacer-height-desktop": heightDesktop,
+      "--spacer-height-mobile": heightMobile,
+      "--spacer-color": `${color}`,
```
richardxia: This one also doesn't require the string interpolation, and you can just directly pass in the variable.

```suggestion
      "--spacer-color": color,
```

richardxia (approved):
One last nit, but this LGTM! I'm guessing we may need to do a second pass when we get the rest of the components instantiated on the pages, but thanks for fixing up the spacing everywhere else!

PR #107, opened 2020-11-19
Integrate Little Green Light donation form

richardxia:
## Overview

This PR is an attempt at implementing the main component on the Donate page, which includes the Little Green Light form.

Note that it looks fairly different from the [Figma design](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1680%3A115), since there's no way for me to style the embedded form from Little Green Light, since it's located in an `<iframe>`, and for security reasons, web browsers do not allow you to interact with the content in an iframe if it comes from a different domain name.

## Caveats and issues

This is not my ideal way of handling this, and there's a few things that make it quite awkward. Aside from the form content just having a different style (typography-wise), I think the major issue with this is the height of the iframe. For security reasons, one cannot query the iframe content for its height, so it's not possible for our site to query for how much vertical space is required for the iframe, even with JavaScript.

The original iframe embed code had scrolling turned off, but this has a major drawback where if we guess wrong for the iframe height, the bottom of the form will get cut off, and the submit button itself will be the first thing to be truncated. That's not great for getting donations from people. In this code, I enabled scrolling, since I thought that the risk of having the content be cut off would be too large, so it would be better to err on the side of showing a scrollbar internal to the iframe if that happens.

One annoying issue is that there just appears to be a bunch of empty space at the bottom of the iframe, below the Submit button, so even when the content appears to fit, there will be an internal scroll region.

The other thing I found problematic about this is that different web browsers will have a different content height. Firefox seems to occupy the most vertical space, while Chrome and Safari occupy less. We have to err on the side of fully containing the form on Firefox, which means that there will be excessive padding on Chrome and Safari.

Another problem I encountered is that the height of the form is dependent on the width of our iframe, and since we're already going for a two-column layout, we really only have half the horizontal space available. On narrower screens, the form itself will take up more vertical space because the input text fields will switch from two-column layout to a single-column layout.

Finally, on mobile, the form is very awkward to use, since the static, fixed height of the iframe — which I hand-tuned for desktop — is definitely too small for mobile, and there will always be scrolling in mobile.

If you have time, @derekfidler, I'd ask that you check out this branch and play around with it yourself to see if you have any opinions on what we should do here.

## Alternate plan?
I do have one idea for a completely different approach to this. According to https://help.littlegreenlight.com/article/224-advanced-pass-values-into-a-form-via-email, there's a way to prefill form fields from an email link, where you can place the form values in the query parameters of a URL for the form that's hosted on Little Green Light's site. We could possibly implement a pure React form on our end, and when a user submits the form on our site, instead of actually submitting a real form request, we would just take the user to LGL's site with the form fields prefilled with what the user already set on our site.

I think the main drawbacks with this approach is 1) we would have to be _very_ careful to not change the form definition in LGL at all, since the form fields just have names like `field_1`, `field_2`, etc., and 2) this is going to be a bit more work to implement, and we're already coming up pretty close to when we want the site shipped by, especially since I'm getting requests to have the site live by next Tuesday (11/24) for an email campaign.

## Screenshots
screencapture-localhost-8000-new-donate-2020-11-18-20_46_57
Screen Shot 2020-11-18 at 8 49 10 PM

### DIFF
--- a/src/components/grid-aware/DonationBlock/DonationBlock.js
+++ b/src/components/grid-aware/DonationBlock/DonationBlock.js
@@ -1,0 +1,71 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./DonationBlock.module.css";
+
+const rawLGLForm = {
+  // There are two manual modifications to this snippet from Little Green Light:
+  // - The height has been manually modified, since the original value didn't seem
+  //   to work. Note that Firefox's text box sizes seem to be slightly larger,
+  //   so the height needs to be large enough to accommodate Firefox
+  // - Scrolling has been enabled, since in a worst-case situation, we don't
+  //   want the submit button to be cut off and unable to be scrolled to.
+  __html: `
+<iframe onload="window.parent.scrollTo(0,0)" height="1040" allowTransparency="true" frameborder="0" scrolling="yes" style="width:100%;border:none"  src="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw"><a href="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw">Fill out my LGL Form!</a></iframe>
+`,
+};
+
+const LGLForm = () => (
+  // eslint-disable-next-line react/no-danger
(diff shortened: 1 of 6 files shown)


davidagustin (approved):
Looks good to me

PR #110, opened 2020-11-19
Fix background and spacing for Video Spotlight

davidagustin:
Screen Shot 2020-11-19 at 3 40 06 AM

### DIFF
--- a/src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
+++ b/src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
@@ -38,31 +38,38 @@ const VideoSpotlightBlock = ({
   button,
   image,
   playButtonLink,
-}) => (
-  <div className={s.bleedWrapper}>
-    <div className={s.bleedImage}>
-      <div className={s.playButton}>
-        <a rel="noreferrer" href={playButtonLink} target="_blank">
-          <img className={s.playIcon} src={playIcon} alt="Play Video" />
-        </a>
+  blackBackground,
+}) => {
+  const bleedBackgroundWrapper = `${s.bleedWrapper} ${
+    blackBackground ? s.blackBackground : ""
+  }`;
+
+  return (
+    <div className={bleedBackgroundWrapper}>
+      <div className={s.bleedImage}>
+        <div className={s.playButton}>
+          <a rel="noreferrer" href={playButtonLink} target="_blank">
+            <img className={s.playIcon} src={playIcon} alt="Play Video" />
+          </a>
+        </div>
+        <img className={s.image} src={image.url} alt={image.alt} />
       </div>
-      <img className={s.image} src={image.url} alt={image.alt} />
-    </div>
-    <div className={s.bleedMainContent}>
-      <div className={s.gridParent}>
-        <div className={s.gridAreaCard}>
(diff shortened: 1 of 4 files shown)


src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
```diff
@@ -38,38 +38,50 @@ const VideoSpotlightBlock = ({
...
-    <div className={s.bleedImage}>
-      <div className={s.playButton}>
-        <a rel="noreferrer" href={playButtonLink} target="_blank">
-          <img className={s.playIcon} src={playIcon} alt="Play Video" />
-        </a>
+  blackBackground,
+}) => {
+  const bleedBackgroundWrapper = `${s.bleedWrapper} ${
+    blackBackground ? s.blackBackground : ""
+  }`;
```
richardxia, line 45: Just an observation here, and no need to change anything here, but maybe it's about time we look at using the [`classnames`](https://github.com/JedWatson/classnames) library, which makes this conditional class name computation a bit cleaner.

See https://github.com/JedWatson/classnames/tree/b166fd132d4a46b6ed93b3264ca638c7f14d4620#usage-with-reactjs for an example of how this can be used, but it works pretty well with React.

I can create a ticket to investigate this later.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
```diff
@@ -2,10 +2,14 @@
   display: grid;
   grid-template-columns: 1fr minmax(var(--grid-content-min-width), var(--grid-content-max-width)) 1fr;
   grid-template-rows: 465px;
-  margin: auto;
+  margin: auto auto 42px;
```
richardxia: I think we don't actually even really want the `auto` on the top side, right? Maybe this should be:

```suggestion
  margin: 0 auto 42px;
```

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.stories.js
```diff
@@ -21,6 +21,7 @@ const Template = ({
     button={button}
     image={image}
     playButtonLink={playButtonLink}
+    blackBackground
```
richardxia: I would say that this should be a toggleable option in Storybook, so that you can easily view both variants.

richardxia (approved):
Just a few minor comments, but this otherwise LGTM!

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css
davidagustin: ok

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.stories.js
davidagustin: ok

PR #111, opened 2020-11-19
Mission Hotel and Volunteer Spotlight Change

davidagustin:
I changed the view position from center to bottom. The picture is also a jpeg so I hope we can find a better quality picture. It appears pixelated.
Screen Shot 2020-11-19 at 4 06 45 AM

If I kept the display to be the center, it would look like:
Screen Shot 2020-11-19 at 4 07 43 AM


src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.stories.js
```diff
@@ -26,10 +26,14 @@ const Template = ({
 
 export const Default = Template.bind({});
 Default.args = {
-  eyebrowText: "Partnership Spotlight",
-  title: "Digital Equity Team of San Francisco",
+  eyebrowText: "ShelterConnect Spotlight",
+  title: "Mission Hotel",
```
richardxia: There's not really a need to change the Storybook story. This component will be used in multiple places, and we will periodically update the cards as we have new things to highlight, so this Storybook story doesn't have always be updated with it.

This is partially why I wanted to separate the images used by the Storybook stories from the images used in the actual pages.

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.js
```diff
@@ -39,7 +39,7 @@ const ArticleSpotlightCard = ({
     <div
       className={s.bleedBackground}
       style={{
-        background: `center / cover no-repeat url("${backgroundImage}")`,
+        background: `bottom / cover no-repeat url("${backgroundImage}")`,
```
richardxia: I'm not sure if this is a good idea. Won't this interfere with the Article Spotlight on the Volunteer page, which has a different image?

Could we add that component to the Volunteer page first so that we know what the impact of making any further CSS tweaks to the component?

richardxia (commented):
In addition to my other comments, I have one more request, which is that you actually instantiate the Article Spotlight Card on the Donate page with the same Mission Hotel article link.

We have this component on all three of the current pages on the site, pointing to two different articles:

- Home page: Mission Hotel blog post
- Volunteer page: Laura's spotlight article
- Donate page: Mission Hotel blog post

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.js
```diff
@@ -5,10 +5,11 @@ import Button from "../../inline/Button";
 
 import s from "./ArticleSpotlightCard.module.css";
 
-const TextCard = ({ eyebrowText, title, description, button }) => (
+const TextCard = ({ eyebrowText, title, subtitle, description, button }) => (
   <div className={s.textCard}>
     <div className={s.eyebrow}>{eyebrowText}</div>
     <div className={s.title}>{title}</div>
+    <div className={s.subtitle}>{subtitle}</div>
```
richardxia: Let's conditionally render this only if `subtitle` is defined.

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.module.css
```diff
@@ -76,6 +76,11 @@
   margin-top: 20px;
 }
 
+.subtitle {
+  color: var(--color-gray-600);
+  font: var(--font-body-medium);
```
richardxia: In the design, it looks like the font is actually `--font-subtitle`, not `--font-body-medium`.

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.stories.js
```diff
@@ -30,6 +30,7 @@ Default.args = {
   title: "Digital Equity Team of San Francisco",
   description:
     "Free Fiber Initiative summary we're working together to find new and different ways of building lasting infrastructure.",
+  text: "Read More",
```
richardxia: Did you mean for this to be set to `subtitle`, and for it to actually get passed into the Article Spotlight component?

src/pages/new/donate/index.js
```diff
@@ -1,5 +1,7 @@
 import React from "react";
 
+import ArticleSpotlightCard from "../../../components/grid-aware/ArticleSpotlightCard";
+import articleSpotlightImage from "../../../components/grid-aware/ArticleSpotlightCard/mission-hotel.jpeg";
```
richardxia: Instead of putting the image in the component directory, could you put it in the directory containing the page itself? The images are not part of the generic components, and they are specific to the instances of the component on the individual pages.

src/pages/new/volunteer/index.js
```diff
@@ -1,5 +1,7 @@
 import React from "react";
 
+import ArticleSpotlightCard from "../../../components/grid-aware/ArticleSpotlightCard";
+import articleSpotlightImage from "../../../components/grid-aware/ArticleSpotlightCard/laura-barerra-vera.png";
```
richardxia: For similar reasons to my last comment, could you also move this image into the actual page directory?

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.stories.js
davidagustin: This was a mistake

richardxia (approved):
LGTM, thanks!

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.js
davidagustin: ok

PR #112, opened 2020-11-19
1737 tweak image header

davidagustin:
Screen Shot 2020-11-19 at 4 59 35 AM

I had to eyeball a decimal point of the viewport width but hope I did this right.

### DIFF
--- a/src/components/grid-aware/ImageHeader/ImageHeader.module.css
+++ b/src/components/grid-aware/ImageHeader/ImageHeader.module.css
@@ -64,6 +64,7 @@
   }
 
   .gridAreaImage2 {
+    margin-left: 22.67vw;
     text-align: right;
   }
 }


richardxia:
Wait, I think @cw323 took care of this in https://github.com/ShelterTechSF/sheltertech.org/pull/101

davidagustin:
Screen Shot 2020-11-19 at 11 04 06 AM

If that were the case, staging still has incorrect dimensions.
Candy's PR at https://github.com/ShelterTechSF/sheltertech.org/pull/101/files was effecting image1

With a 375 viewport width on mobile, the image2 width should have a 210 pixel width.

richardxia:
> With a 375 viewport width on mobile, the image2 width should have a 210 pixel width.

I think it's going to be a little bit difficult to get something that is pixel precise here that would make sense in slightly larger or smaller screens. While I usually like to be pixel precise, it's normally only in the pursuit of simplicity, where consistency often results in simplicity. Here, I'm not sure if I can intuit any obvious rule that makes this better across other mobile screens, and I don't feel like `margin-left: 22.67vw` helps us with simplicity or consistency.

richardxia:
Closing, as we are in the process of porting the site to a no-code platform, Wix, so that the non-engineering teams at ShelterTech can maintain the corporate site on their own.

PR #117, opened 2020-11-20
Create spacer in Volunteer Spotlight

davidagustin:


src/pages/new/volunteer/index.js
```diff
@@ -92,6 +92,7 @@ export default () => (
         },
       ]}
     />
+    <Spacer heightDesktop="120px" heightMobile="80px" />
```
richardxia: I think this might need to be 190px in desktop? Since the image that hangs off the bottom of the component above doesn't seem to accounted for as part of the spacing.

davidagustin: ok

richardxia (approved):
LGTM!

PR #119, opened 2020-11-20
Fix padding for input

davidagustin:
Since there is a 1px border in the input, minused from 2px originally. Ive increased the top and bottom padding by 1px.
Screen Shot 2020-11-20 at 12 40 31 AM
Screen Shot 2020-11-20 at 12 40 24 AM


richardxia (approved):
LGTM!

PR #121, opened 2020-11-20
fix video header link

davidagustin:


src/pages/new/index.js
```diff
@@ -51,7 +51,7 @@ export default () => {
           { text: "Donate", internalLink: "/new/donate" },
           { text: "Volunteer", internalLink: "/new/volunteer" },
         ]}
-        playButtonLink="https://www.youtube.com/watch?v=KCduRWJ1hQo"
+        playButtonLink="https://www.youtube.com/watch?v=2aLyGwaRufY&t=2s"
```
richardxia: Maybe get rid of the `&t=2s` so that it starts from the beginning of the video and not 2 seconds in?

```suggestion
        playButtonLink="https://www.youtube.com/watch?v=2aLyGwaRufY"
```

richardxia (approved):
LGTM!

PR #122, opened 2020-11-20
Move new pages out of new/.

richardxia:
This does the final step of moving the new pages out of `new/` and into the real pages on the site.

Note that I kept most of the older pages, in case if we still want search engines to be able to link to them until we replace them with newer pages. The only two pages that I actually deleted was the Home page and the Donate page. All the other pages are still available at the older URLs.

I clicked around on the site, and I think this makes all the changes necessary to internal links between pages and imports.


davidagustin (approved):

davidagustin:
The storybook footer links need to change

richardxia:
> The storybook footer links need to change

I don't think we need to change them, especially because the links in the Storybook stories don't actually work, since they're not hooked up to the actual site router.

PR #124, opened 2020-11-22
1661 create blockquote block component

ivanleeswe:
Change quote attribute.

image


richardxia:
It looks like you started from the original branch that added this component. Could you instead create a brand new branch from `main`? In general, we should always be branching from the `main` branch whenever starting new work.

In addition, it looks like you reused the old branch name, with the old ticket number in it. Could you please use the new ticket's number (https://app.clubhouse.io/sheltertech/story/1756/minor-copy-tweaks-to-block-quote-and-one-paragraph-block) so that we can more easily keep track of it?

PR #125, opened 2020-11-22
Refactor VideoSpotlightBlock to not use z-index.

richardxia:
While working on the embedded video modals, I hit an issue due to the `z-index: 1` that we had on this component, where the card would be on top of the modal. This PR changes the VideoSpotlightBlock to not rely on `z-index`, which makes it more self-contained. I decided to split this out into its own PR first, since it seemed substantial and standalone enough to warrant splitting off into its own change.

This makes use of a trick described in https://fettblog.eu/blog/2013/06/16/preserving-aspect-ratio-for-embedded-iframes/ to allow for responsive elements that preserve the aspect ratio.

This PR makes a few structural changes to the component:

- The `` element has been removed, and instead we now set the image as a background image of the `.bleedImage` element.
- There is a new `.bleedImageWrapper` element that sits between the `.bleedWrapper` and the `.bleedImage` elements. This element has different behavior in desktop and mobile, but in both cases, it is there to help with positioning of the `.bleedImage`.
- `.bleedImage` now uses flexbox to center the play button icon.

### DIFF
--- a/src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
+++ b/src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js
@@ -4,13 +4,6 @@ import Button from "../../inline/Button";
 import playIcon from "./PlayIcon.svg";
 import s from "./VideoSpotlightBlock.module.css";
 
-/* PropType shapes */
-
-const ImagePropType = PropTypes.shape({
-  url: PropTypes.string.isRequired,
-  alt: PropTypes.string.isRequired,
-});
-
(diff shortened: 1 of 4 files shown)


davidagustin (approved):
We'll have to see the VideoModal PR to verify this

richardxia:
> We'll have to see the VideoModal PR to verify this

That's not necessary. This is a standalone PR, and it does exactly what the title says, which is to refactor it to not use z-index anymore. One can verify that this PR is correct by confirming that the appearance of this component has not changed in the browser and that there is no more `z-index` property on the elements in this component.

PR #127, opened 2020-11-23
Make videos open in modals

richardxia:
## Overview

This changes the video links on the home page to open up a modal with the video in an embedded YouTube player, allowing users to watch the video without opening the link in a separate window.

## Screenshots

Screen Shot 2020-11-22 at 5 17 50 PM
Screen Shot 2020-11-22 at 5 18 07 PM

## Detailed changes

https://github.com/ShelterTechSF/sheltertech.org/commit/36116cabe797efd5f6764a0c3fd0fb04a629dd14 and https://github.com/ShelterTechSF/sheltertech.org/commit/1e6910b2cfed0e1fa349f1e8b4dd87b13fb93900 are smaller changes to the `VideoHeader` component and the `VideoSpotlightBlock` that change the `playButtonLink` props to `playButtonOnClick`. This is needed because instead of containing external links, this should now just trigger a React state change to open the modal.

https://github.com/ShelterTechSF/sheltertech.org/commit/3b8008761879839d224c12709161f16068bb3f10 modifies the `<Modal>` component to add a new `noBezel` variant. When `noBezel` is set, then we don't render the white container box with the padding, the close button, and the rounded corners. I initially tried putting the video embed in the normal modal container, but it looked pretty weird, so instead I decided that it would probably look the best with no "bezel" at all.

https://github.com/ShelterTechSF/sheltertech.org/commit/3f6734221d0695e5de063971d528fd6a382cc8c6 implements the actual `YouTubeEmbed` component, which is just a lightweight wrapper around the embed `<iframe>` that you get from YouTube. The main thing that we add on top is a wrapper that provides an aspect-ratio-preserving style on top.

Finally, https://github.com/ShelterTechSF/sheltertech.org/commit/3332f4d98c07b716876dd1cf4ad36c92bbe6ab19 actually ties everything together by adding the new video modal onto the home page.

### DIFF
--- a/src/components/grid-aware/VideoHeader/VideoHeader.stories.js
+++ b/src/components/grid-aware/VideoHeader/VideoHeader.stories.js
@@ -8,13 +8,13 @@ export default {
   component: VideoHeader,
 };
 
-const Template = ({ title, description, ctaButtons, playButtonLink }) => (
+const Template = ({ title, description, ctaButtons }) => (
   <VideoHeader
     title={title}
     description={description}
     image={videoHeaderImage}
     ctaButtons={ctaButtons}
-    playButtonLink={playButtonLink}
+    playButtonOnClick={() => {}}
   />
 );
 
@@ -28,5 +28,4 @@ DefaultVideoHeader.args = {
     { text: "Donate", internalLink: "/new/donate" },
     { text: "Volunteer", internalLink: "/new/volunteer" },
   ],
-  playButtonLink: "https://www.youtube.com/watch?v=2aLyGwaRufY",
 };
--- a/src/components/block/YouTubeEmbed/YouTubeEmbed.js
+++ b/src/components/block/YouTubeEmbed/YouTubeEmbed.js
@@ -1,0 +1,24 @@
+import PropTypes from "prop-types";
+import React from "react";
+
+import s from "./YouTubeEmbed.module.css";
+
+const YouTubeEmbed = ({ url, title }) => (
+  <div className={s.wrapper}>
+    <iframe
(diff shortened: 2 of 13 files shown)


src/components/grid-aware/VideoHeader/VideoHeader.stories.js
```diff
@@ -28,5 +28,4 @@ DefaultVideoHeader.args = {
     { text: "Donate", internalLink: "/new/donate" },
     { text: "Volunteer", internalLink: "/new/volunteer" },
```
candywang, line 29: I understand this is the `stories.js` file and won't be used in production but did we want to update the `internalLink` paths to `"/donate"` and `"/volunteer"` respectively even if it doesn't make a difference?

candywang (approved):
LGTM!

davidagustin (approved):
All questions asked through personal conversation.

src/components/grid-aware/VideoHeader/VideoHeader.stories.js
richardxia: I don't think it's really necessary, and if we were to do something here, I'd probably want to just replace these with completely dummied out URLs, just to emphasize that the actual URLs in the links is not important in the Storybook story.

richardxia:
Thanks for quickly reviewing this!

src/components/grid-aware/VideoHeader/VideoHeader.stories.js
candywang: Okay! Thanks for the reference.

PR #129, opened 2020-11-23
Refactor donation for giving tuesday

davidagustin:
Screen Shot 2020-11-22 at 9 31 10 PM

Screen Shot 2020-11-22 at 9 38 28 PM

I made my own best judgement on the CSS @derekfidler please take a look


src/components/grid-aware/DonationBlock/DonationBlock.js
```diff
@@ -22,50 +22,96 @@ const LGLForm = () => (
...
+          <div className={s.mainDescription}>{mainDescription}</div>
+          {givingTuesdayParagraph &&
+            givingTuesdayParagraph.map((paragraph, index) => (
+              <div
+                className={s.givingTuesdayParagraph}
+                key={`${paragraph + index}`}
+              >
+                {paragraph}
+              </div>
+            ))}
```
richardxia: Let me make a few independent comments about this:

1. If you ever find yourself doing `foo && foo.map(...)`, then you should just give `foo` a default value of the empty array, as that would allow you to just call `foo.map(...)` without the guard because an empty array will map to the empty array anyway.
2. You should never be generating a React [key](https://reactjs.org/docs/lists-and-keys.html#keys) with the iteration index, since that defeats the purpose of the key in the first place, which is to have an identifier that is independent of the index.
3. I wouldn't advise mapping over all of these in the first place. This is one of the rare instances where I think tag selectors would be appropriate, as long as they are scoped by some other ancestor class name.

Could you just directly render these paragraphs, possibly wrapped by a single div with a class name? See @cw323's One-Paragraph Block component for an example of this: https://github.com/ShelterTechSF/sheltertech.org/pull/95

src/components/grid-aware/DonationBlock/DonationBlock.js
```diff
@@ -22,50 +22,96 @@ const LGLForm = () => (
 
 const DonationBlock = ({
   mainTitle,
+  givingTuesdayImage,
+  givingTuesdayParagraph,
```
richardxia: Could you give these names that are less specific to Giving Tuesday and more general to the role which they play on this page?

Also, please pluralize the names of any props that hold an array of items. It's confusing that `givingTuesdayParagraph` is singular but contains an array of paragraphs.

src/components/grid-aware/DonationBlock/DonationBlock.module.css
```diff
@@ -74,6 +74,18 @@
   margin-top: 5px;
 }
 
+.mainTitleImage {
+  height: auto;
+  margin-bottom: 0;
```
richardxia: Could you increase the bottom margin? Compared to the normal `.mainTitle`, this ends up having less effective space between it and the subtitle. The `.mainTitle` has a font size of 50px and a line height of 1.3 = 65px, which means that there's actually (65-50)/2 = 7.5px of effective space between the baseline of the text and the bottom of the element.

Without additional spacing, it makes the page feel cramped.

src/components/grid-aware/DonationBlock/DonationBlock.module.css
```diff
@@ -74,6 +74,18 @@
...
+  margin-top: 5px;
+  max-width: 100%;
+}
+
+.givingTuesdayParagraph {
+  font: var(--font-body-medium);
+  margin-top: 40px;
```
richardxia, line 91: This needs to be decreased in mobile because 40px is far too large.

src/components/grid-aware/DonationBlock/DonationBlock.stories.js
```diff
@@ -8,25 +9,29 @@ export default {
...
+    givingTuesdayParagraph={givingTuesdayParagraph}
     whyDonateTitle={whyDonateTitle}
     whyDonateList={whyDonateList}
     impactTitle={impactTitle}
     impactList={impactList}
   />
 );
 
-export const Default = Template.bind({});
+export const DonationBlockDefault = Template.bind({});
```
richardxia: Is there a reason you renamed this?

src/pages/donate/index.js
```diff
@@ -1,48 +1,98 @@
...
+        />
+      </Modal>
+      <DonationBlock
+        givingTuesdayImage={{
+          url: image,
+          alt: "Giving Tuesday logo.",
+        }}
+        mainDescription="For 2020, ShelterTech has a $20,000 fundraising goal to cover the installation costs for WiFi for the next SF building on our list. We have already raised half of that goal, and ask your support to raise the remaining $10,000."
+        givingTuesdayParagraph={[
+          <span>
```
richardxia: Why are these `<span>`s? We want these to behave like block elements, so could you just have these be regular `<p>` tags?

I would also recommend that you just make this whole thing a React Fragment rather than an array of JSX elements, similar to https://github.com/ShelterTechSF/sheltertech.org/pull/95.

src/pages/donate/index.js
```diff
@@ -1,48 +1,98 @@
...
+      >
+        <YouTubeEmbed
+          url="https://www.youtube.com/embed/2aLyGwaRufY"
+          title="Why Internet - Learn about the digital divide and how you can help"
+        />
+      </Modal>
+      <DonationBlock
+        givingTuesdayImage={{
+          url: image,
+          alt: "Giving Tuesday logo.",
```
richardxia: I'd probably just call this "Giving Tuesday", without the "logo.", since this plays the role of a heading.

src/pages/donate/index.js
```diff
@@ -1,48 +1,98 @@
...
+            our video&nbsp;
+            {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */}
+            <span
+              style={{
+                color: "#069",
+                textDecoration: "underline",
+                cursor: "pointer",
+              }}
+              onClick={() => setVideoHeaderModalIsOpen(true)}
+            >
```
richardxia: I don't think `<span>` is an appropriate tag to use here. `<a>` is probably fine, although from a UX perspective, it's actually kind of weird for an inline link like this to be popping open a modal on the page in the first place, so maybe we should just make this a normal link.

src/pages/donate/index.js
```diff
@@ -1,48 +1,98 @@
...
+              here
+            </span>
+            .
+          </span>,
+          <span>
+            Last year, <span style={{ textDecoration: "underline" }}>you</span>{" "}
```
richardxia: Please don't use spans to do stuff like this. The correct tag to use here is `<em>`. Underlines also don't really translate well to the web, since they are often used to indicate hyperlinks.

richardxia (commented):
Overall, this approach looks fine to me, but I'd request that we make some changes to this.

richardxia:
A few last comments that I got from talking to Derek about this just now:

- He suggested that we make the left area just span 5 columns instead of 6, since the text is a bit too close to the donation form.
- We should update the actual embedded form itself to use the special Giving Tuesday version of the form. Here's the embed code:

```html
<iframe onload="window.parent.scrollTo(0,0)" height="1053" allowTransparency="true" frameborder="0" scrolling="no" style="width:100%;border:none"  src="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw"><a href="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw">Fill out my LGL Form!</a></iframe>
```

Note that I made a few manual tweaks to the last one, such as increasing the height slightly and enabling scrolling.

src/components/grid-aware/DonationBlock/DonationBlock.stories.js
davidagustin: Other storybook components do not just have "Default" but its component name appended with "Default" at the end.

src/components/grid-aware/DonationBlock/DonationBlock.module.css
davidagustin: ok

src/components/grid-aware/DonationBlock/DonationBlock.module.css
davidagustin: I used 10px margin-bottom

src/pages/donate/index.js
davidagustin: ok

src/pages/donate/index.js
davidagustin: ok

src/components/grid-aware/DonationBlock/DonationBlock.js
davidagustin: > A few last comments that I got from talking to Derek about this just now:
> 
> * He suggested that we make the left area just span 5 columns instead of 6, since the text is a bit too close to the donation form.
> * We should update the actual embedded form itself to use the special Giving Tuesday version of the form. Here's the embed code:
> 
> ```
> <iframe onload="window.parent.scrollTo(0,0)" height="1053" allowTransparency="true" frameborder="0" scrolling="no" style="width:100%;border:none"  src="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw"><a href="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw">Fill out my LGL Form!</a></iframe>
> ```
> 
> Note that I made a few manual tweaks to the last one, such as increasing the height slightly and enabling scrolling.

Made changes

davidagustin: "Could you just directly render these paragraphs, possibly wrapped by a single div with a class name? See @cw323's One-Paragraph Block component for an example of this: #95" I used a parent class and element selector as specified below

src/components/grid-aware/DonationBlock/DonationBlock.js
davidagustin: ok

src/pages/donate/index.js
davidagustin: ok

src/pages/donate/index.js
davidagustin: ok

davidagustin:
Screen Shot 2020-11-23 at 8 30 35 PM
Screen Shot 2020-11-23 at 8 30 48 PM

src/components/grid-aware/DonationBlock/DonationBlock.js
```diff
@@ -11,7 +11,7 @@ const rawLGLForm = {
   // - Scrolling has been enabled, since in a worst-case situation, we don't
   //   want the submit button to be cut off and unable to be scrolled to.
   __html: `
-<iframe onload="window.parent.scrollTo(0,0)" height="1040" allowTransparency="true" frameborder="0" scrolling="yes" style="width:100%;border:none"  src="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw"><a href="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw">Fill out my LGL Form!</a></iframe>
+<iframe onload="window.parent.scrollTo(0,0)" height="1053" allowTransparency="true" frameborder="0" scrolling="no" style="width:100%;border:none"  src="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw"><a href="https://secure.lglforms.com/form_engine/s/0wDrXsUXwzxrT32_RHQGrw">Fill out my LGL Form!</a></iframe>
```
richardxia: Could you please enable scrolling and set the height so that the submit button is visible in Firefox? See my comment a few lines higher about the alterations that I made.

src/pages/donate/index.js
```diff
@@ -13,24 +14,46 @@ export default () => (
...
+            shows and movies. People experiencing homelessness often do not have
+            internet access, making it harder to do all of these things.
+          </p>
+          <p>
+            At ShelterTech, we believe connectivity is a right, not a privilege,
+            and through our ShelterConnect program, we have installed free WiFi
+            at a number of San Francisco-based shelters to help bridge this
+            digital divide. You can learn more about ShelterConnect by watching
+            our video&nbsp;
+            <a href="https://www.youtube.com/embed/2aLyGwaRufY">here</a>.
```
richardxia: This looks like it's the embed URL. Can you just link to the normal YouTube page?

src/pages/donate/index.js
```diff
@@ -13,24 +14,46 @@ export default () => (
...
+            purchase deliverable groceries and take-out, and stream our favorite
+            shows and movies. People experiencing homelessness often do not have
+            internet access, making it harder to do all of these things.
+          </p>
+          <p>
+            At ShelterTech, we believe connectivity is a right, not a privilege,
+            and through our ShelterConnect program, we have installed free WiFi
+            at a number of San Francisco-based shelters to help bridge this
+            digital divide. You can learn more about ShelterConnect by watching
+            our video&nbsp;
```
richardxia: Please don't use a non-breaking space here. These should be breakable spaces. To appease Prettier, you'll have to use a JSX literal to place a space here (`{ ' ' }`). This goes for all the non-breaking spaces you had added.

src/pages/donate/index.js
```diff
@@ -13,24 +14,46 @@ export default () => (
...
+            <a href="https://www.youtube.com/embed/2aLyGwaRufY">here</a>.
+          </p>
+          <p>
+            Last year,&nbsp;<em>you</em>&nbsp; and other supporters gave
```
richardxia: I noticed two things about these `<em>`:

1. It looks like our CSS reset is causing all styling to be stripped from this. Can you style this to make italic?
2. It looks like we don't actually load the italic version of this font from Google Fonts. Can you update our global CSS file to get the italic version as well?

richardxia (commented):
This is looking better. I think most of the JSX is looking pretty good now, and I think the remaining things are smaller, more style-related things!

src/components/grid-aware/DonationBlock/DonationBlock.js
davidagustin: ok

src/pages/donate/index.js
davidagustin: ok

src/pages/donate/index.js
```diff
@@ -13,24 +14,47 @@ export default () => (
...
+          <p>
+            At ShelterTech, we believe connectivity is a right, not a privilege,
+            and through our ShelterConnect program, we have installed free WiFi
+            at a number of San Francisco-based shelters to help bridge this
+            digital divide. You can learn more about ShelterConnect by watching
+            our video{" "}
+            <a href="https://www.youtube.com/watch?v=2aLyGwaRufY">here</a>.
+          </p>
+          <p>
+            Last year, <span style={{ fontStyle: "italic" }}>you</span> and
```
richardxia: Sorry, I think I was unclear with what I was asking before. Could you:

1. change this back to `<em>`, and
2. write some CSS in DonationBlock.module.css to style it to be italic?

richardxia (commented):
Hopefully one last set of changes. Everything else is looking good to me, though!

richardxia (approved):
LGTM!

PR #130, opened 2020-11-23
Add flex to all statCard child components and add 0 margin-right to t…

ivanleeswe:
Add shorthand flex to the child components statCard to evenly distribute the basis(default size of an element before the remaining space is distributed), shrink(the ability for flex item to shrink if necessary), and grow(the ability for flex item to grow if necessary).

Add 0 to margin-right for the last child statCard to take away the spacing.

### DIFF
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -73,8 +73,13 @@
 }
 
 .statCard {
+  flex: 1 1 0;
   margin-right: 110px;
   width: 250px;
+}
+
+.statCard:last-child {
+  margin-right: 0;
 }
 
 @media (--tablet-and-down) {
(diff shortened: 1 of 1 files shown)


src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -92,6 +97,7 @@
   font-size: 72px;
   font-weight: bold;
   line-height: 1.3;
+  min-width: 100px;
```
richardxia: Is this still needed with the other two changes?

richardxia (approved):
LGTM, besides one comment about whether something is still needed.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
ivanleeswe: Good catch. It's not needed. I will delete it.

davidagustin (approved):
LGTM

candywang (approved):
LGTM!

PR #133, opened 2020-11-24
Touch up CSS to improve ThreeParagraphBlock at 1024px.

richardxia:
This tries to improve the ThreeParagraphBlock when viewed at 1024px.

There's really just two changes I made:

- I set a margin-right on the image on the left side to ensure that there is always some distance separating it from the other paragraphs
- I updated the Button component to center-align its text. I couldn't actually figure out a way to make the button narrower when it wraps. It just looks like when an inline-block element wraps, then it just takes up the full width of its parent. The best I think I can do here is just center-align the button text.

## Before

Screen Shot 2020-11-23 at 7 30 57 PM

## After

Screen Shot 2020-11-23 at 7 30 30 PM


davidagustin (approved):
LGTM!

PR #135, opened 2020-11-24
Update Article Spotlight on Volunteer Page with final link.

richardxia:
We got the Volunteer Spotlight blog post up, so now we can update this component at the bottom of the Volunteer page with the real link.


richardxia:
I'm just going to merge without review, since it's a pretty small change, and this is the last thing blocking the deploy to production.

PR #136, opened 2020-11-24
Revert back to old donation page

richardxia:
After talking to Jason, it looks like we're going to hold off from launching the Giving Tuesday version of the Donate page until Giving Tuesday itself, so I'm reverting this back to what it used to be.

Next Tuesday we can simply revert this PR to go back to the Giving Tuesday version of the donate page.

I also made a couple of small copy changes that were requested in https://github.com/ShelterTechSF/sheltertech.org/commit/204d0a828e8e7a1c9a42003bf4d9e8d69944d9f9.


davidagustin (approved):
LGTM

richardxia:
Pushed out one more commit, since we modified the original form to have more fields, so I had to resize it again to fit.

PR #139, opened 2020-11-25
Giving tuesday donation page update

davidagustin:


davidagustin:
mistake push do not use this

PR #148, opened 2020-11-30
Update Partners and Sponsors logos

### LINKED ISSUE #147: Update Partners and Sponsors logos
We should update the set of logos in the Partners and Sponsors component to include everything in the latest Figma design: https://www.figma.com/file/HLVHC58KwkcVFXkKAuY0ej/ShelterTech.org-Design-System?node-id=669%3A0

We should also double check that the component renders fine even though there are four rows of logos now.

candywang:
I updated the set of logos in the Partners and Sponsors component and also created a separate module to contain the array of all the logos. This makes the Home page - `src/pages/index.js` import block look cleaner with the reduction of 17 lines of code. If everything looks good, we can go ahead and close #147.


richardxia (approved):
Overall, I like this approach of factoring out that data into a common file. My one request here is maybe we could take out out of the `stories/` directory, and instead have it in some place where we place common data until we actually do the CMS integration?

As a straw proposal, how about a new `src/data/` directory?

davidagustin (approved):
With Richards request in, LGTM

ivanleeswe (approved):

PR #153, opened 2020-12-14
Install and save Classnames library

### LINKED ISSUE #143: Investigate whether we should use the classnames library
We have a few components where we conditionally compute the CSS class names depending on prop values. The [classnames](https://github.com/JedWatson/classnames) NPM library is specifically designed to help make this cleaner.

An example of this in our codebase is https://github.com/ShelterTechSF/sheltertech.org/pull/110#discussion_r527021090.

cc @davidagustin

candywang:
The [Classnames](https://github.com/JedWatson/classnames) library allows the expression of conditional classes as an object.

```
// Previous conditional statement
const modalClassName = `${s.content} ${noBezel ? s.noBezel : s.withBezel}`;

// New conditional statement with Classnames library
const classStyles = classNames.bind(s);
const modalClassName = classStyles({
  content: true,
  noBezel,
  withBezel: !noBezel,
});
```

In places where **CSS Modules** are being used, the **`bind`** variant should be applied:
```
import classNames from "classnames/bind";
import s from "./component.module.css";

const classStyles = classNames.bind(s);
const programClassName = classStyles({
  container: true,
  light: theme === "light",
  dark: theme === "dark",
});

// If light is true => programClassName = "container light"
// If dark is true => programClassName = "container dark"
```
vs
```
import classNames from "classnames";
import "./component.css";

const className = classNames({
  button: true,
  background: isDark,
});

// If isDark is true => className = "button background"
// If isDark is false => className = "button"
```

In this PR,
- The [Classnames](https://www.npmjs.com/package/classnames) package was added to `dependencies` in `package.json`.
- Conditional classes were refactored to use the library, making code cleaner and more readable.
- Resolves #143.

### DIFF
--- a/src/components/block/ProgramCard/ProgramCard.js
+++ b/src/components/block/ProgramCard/ProgramCard.js
@@ -1,15 +1,13 @@
-import classNames from "classnames/bind";
+import classNames from "classnames";
 import PropTypes from "prop-types";
 import React from "react";
 
 import s from "./ProgramCard.module.css";
 
 const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
-  const classStyles = classNames.bind(s);
-  const backgroundTheme = classStyles({
-    programCardContainer: true,
-    light: theme === "light",
-    dark: theme === "dark",
+  const backgroundTheme = classNames(s.programCardContainer, {
+    [s.light]: theme === "light",
+    [s.dark]: theme === "dark",
   });
 
   return (
--- a/src/components/block/ProgramCard/ProgramCard.js
+++ b/src/components/block/ProgramCard/ProgramCard.js
@@ -5,8 +5,8 @@ import React from "react";
 import s from "./ProgramCard.module.css";
 
 const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
-  const classStyle = classNames.bind(s);
-  const backgroundTheme = classStyle({
+  const classStyles = classNames.bind(s);
+  const backgroundTheme = classStyles({
     programCardContainer: true,
     light: theme === "light",
     dark: theme === "dark",
--- a/src/components/block/ProgramCard/ProgramCard.js
+++ b/src/components/block/ProgramCard/ProgramCard.js
@@ -1,19 +1,27 @@
+import classNames from "classnames/bind";
 import PropTypes from "prop-types";
 import React from "react";
+
 import s from "./ProgramCard.module.css";
 
-const ProgramCard = ({ theme, image, imageAlt, title, description }) => (
-  <div
-    className={`${s.programCardContainer} 
-    ${theme === "light" ? s.light : s.dark}`}
-  >
-    <div>
-      <img className={s.img} src={image} alt={imageAlt} />
+const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
+  const classStyle = classNames.bind(s);
+  const backgroundTheme = classStyle({
+    programCardContainer: true,
+    light: theme === "light",
+    dark: theme === "dark",
+  });
+
+  return (
+    <div className={backgroundTheme}>
+      <div>
+        <img className={s.img} src={image} alt={imageAlt} />
+      </div>
+      <div className={s.title}>{title}</div>
+      <div className={s.description}>{description}</div>
     </div>
-    <div className={s.title}>{title}</div>
-    <div className={s.description}>{description}</div>
-  </div>
-);
+  );
+};
 
 ProgramCard.propTypes = {
   theme: PropTypes.string.isRequired,
--- a/src/components/block/ProgramCard/ProgramCard.js
+++ b/src/components/block/ProgramCard/ProgramCard.js
@@ -5,10 +5,7 @@ import React from "react";
 import s from "./ProgramCard.module.css";
 
 const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
-  const backgroundTheme = classNames(s.programCardContainer, {
-    [s.light]: theme === "light",
-    [s.dark]: theme === "dark",
-  });
+  const backgroundTheme = classNames(s.programCardContainer, s[theme]);
 
   return (
     <div className={backgroundTheme}>
--- a/src/components/grid-aware/Modal/Modal.js
+++ b/src/components/grid-aware/Modal/Modal.js
@@ -1,4 +1,4 @@
-import classNames from "classnames/bind";
+import classNames from "classnames";
 import PropTypes from "prop-types";
 import React from "react";
 import ReactModal from "react-modal";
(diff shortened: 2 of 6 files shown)


candywang:
If anyone has a suggestion for a better variable name to replace `classStyles`, please let me know and I'll make an update. Thank you! 🙂

src/components/grid-aware/Modal/Modal.js
```diff
@@ -13,10 +14,18 @@ const Modal = ({
   contentLabel,
   noBezel,
 }) => {
-  const modalClassName = `${s.content} ${noBezel ? s.noBezel : s.withBezel}`;
-  const buttonClassName = `${s.closeButton} ${
-    noBezel ? s.noBezel : s.withBezel
-  }`;
+  const classStyles = classNames.bind(s);
+  const modalClassName = classStyles({
+    content: true,
```
richardxia: According to the `classnames`' documentation, you can mix both the string argument method and the object argument method in the same call, where passing in a string is equivalent to setting a key of that name to "true" in the object method. The example from their docs is:

```js
classNames('foo', { bar: true }); // => 'foo bar'
```

What do you think about

```js
classStyles('content', { noBezel, withBezel: !noBezel })
```
?

I think I could go either way, but whichever style we pick, I think we should exclusively go with that style and make it a repo-wide convention. This would be the perfect kind of thing to put in a style guide, which I've been meaning to write down for this application.

@davidagustin, @ivanleeswe, and @jessypeck, what do you think about the two different approaches?

src/components/block/ProgramCard/ProgramCard.js
```diff
@@ -1,19 +1,27 @@
...
 
-const ProgramCard = ({ theme, image, imageAlt, title, description }) => (
-  <div
-    className={`${s.programCardContainer} 
-    ${theme === "light" ? s.light : s.dark}`}
-  >
-    <div>
-      <img className={s.img} src={image} alt={imageAlt} />
+const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
+  const classStyles = classNames.bind(s);
```
richardxia: Regarding the `bind` way of doing this, it looks like in the [README](https://github.com/JedWatson/classnames/tree/v2.2.6#alternate-bind-version-for-css-modules) they recommend a different method if you're in an ES2015 (ES6) environment:

> _Note that in ES2015 environments, it may be better to use the "dynamic class names" approach documented above._

If you look at [the other section that they refer to](https://github.com/JedWatson/classnames/tree/v2.2.6#dynamic-class-names-with-es2015), then it suggests using the special bracket notation for using a variable's value as a key in an object literal, e.g. `{ [myKey]: "foo" }`. What this does is instead of creating an object with a key of `"myValue"`, it instead looks up the value that a variable named `myKey` has and uses _that_ as the actual key.

I think applied to `classnames` and CSS modules, they may be suggesting something like this:

```js
import s from "./Foo.module.css";

const className = classnames({
  [s.card]: true,
  [s.light]: theme === "light",
});
```

richardxia (commented):
> If anyone has a suggestion for a better variable name to replace `classStyles`, please let me know and I'll make an update. Thank you! 🙂

See my other comment, but I think we can get around using the special `bind` thing by using the special object key syntax. If we did, however, want to use the `bind` thing, then I'd suggest using the `cx` short name that you find in their README, since even if it's not very descriptive, it's 1) Googleable, since it appears in their README and 2) it would appear often enough that it would become something that any new reader of the code would learn quickly.

Thanks for taking this ticket! I think it looks mostly good, but since we're not in any rush, I'd like to spend some extra time on deciding on the canonical style we want to use for this entire project.

src/components/grid-aware/Modal/Modal.js
jessica-px: I think I'm fine with using this pattern:
```javascript
classStyles('content', { noBezel, withBezel: !noBezel })
```
Although I do want to say that the two different uses of implicit assignment (`content` via this library, and `noBezel` via the object property value shorthand) initially threw me off for a second. It took me a little extra time to understand why they had to be handled in different ways, and what the final result would look like. That said, I'm fine now that I've looked over the docs. But I do think it's worth observing that by doing more things implicitly, we're asking people to do a little bit more brain gymnastics to understand what's happening.

But looking at the docs, I don't find this example at all difficult to parse:
```javascript
classNames('foo', { bar: true, duck: false }, 'baz', { quux: true });
```
And if anything, I like the way it helps separate the static, always present class names from the conditional ones. So if we as a group think these implicit assignments won't throw us off, I think I'm leaning in this direction.

But I also have no objection to just explicitly declaring everything, even if it might look a little clunkier.

candywang: Thanks for the feedback! I am also fine with using this pattern.

candywang:
Sorry for the delay in getting this updated! I changed the class names to use the dynamic approach.

src/components/block/ProgramCard/ProgramCard.js
```diff
@@ -1,15 +1,13 @@
...
 
 const ProgramCard = ({ theme, image, imageAlt, title, description }) => {
-  const classStyles = classNames.bind(s);
-  const backgroundTheme = classStyles({
-    programCardContainer: true,
-    light: theme === "light",
-    dark: theme === "dark",
+  const backgroundTheme = classNames(s.programCardContainer, {
+    [s.light]: theme === "light",
+    [s.dark]: theme === "dark",
```
davidagustin: I don't think there should be two strict equalities for light and dark since only one can exist.  However, classnames does not allow terneries. There is also another component in this file that has a light and dark activation at src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.js on line 39 which is  [s.blackBackground]: blackBackground . Perhaps this can go in right now but there should be a standard for activation of light and dark.

richardxia: Actually, I think you can address @davidagustin's comment by just using the `theme` prop directly to access the associated CSS class as a variable property on the `s` object. Does it work if you write it as:

```javascript
classNames(s.programCardContainer, s[theme])
```
?

src/components/grid-aware/Modal/Modal.js
```diff
@@ -13,10 +14,15 @@ const Modal = ({
...
   noBezel,
 }) => {
-  const modalClassName = `${s.content} ${noBezel ? s.noBezel : s.withBezel}`;
-  const buttonClassName = `${s.closeButton} ${
-    noBezel ? s.noBezel : s.withBezel
-  }`;
+  const modalClassName = classNames(s.content, {
+    [s.noBezel]: noBezel,
+    [s.withBezel]: !noBezel,
+  });
```
richardxia: Similar to slightly different from my other comment about the light/dark theme, I think this you may be able to do by passing in just a single value for the noBezel/withBezel thing, but where you conditionally select the appropriate class name beforehand. Could you try the following?

```suggestion
  const modalClassName = classNames(s.content, noBezel ? s.noBezel : s.withBezel);
```

If you feel it's more readable, you can also take that ternary expression and assign it to a new variable just above this line.

richardxia (approved):
LGTM! I have a couple of suggestions to make the cases where we pick between two different options be more concise, but otherwise this all looks good to me!

richardxia:
Whoops, sorry for forgetting about this PR! I think we may want to revive this, since the classnames library could be pretty handy for some of the white v.s. gray background theming that we're about to do. It might be easier, however, to start over from a new branch, since this PR predates the TypeScript migration we did earlier this year.

richardxia:
Sorry for letting this get out of date. I'm going to close it, since it predates our TypeScript migration, so we'll probably just need to do it from scratch, but the stuff in this PR may still be useful as a reference.

PR #155, opened 2020-12-18
Add workflow for pushing Storybook to gh-pages

### LINKED ISSUE #152: Publish Storybook site to GitHub Pages
Set up a GitHub Actions workflow that builds and publishes the Storybook site to GitHub Pages. This can be done similar to the staging site builds, which run on every merge to the `main` branch and push the static HTML files up to AWS S3.

jessica-px:
## Description

This PR adds a [Github Actions workflow](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions) that should do the following:

- Whenever we merge to `main`, this workflow gets triggered
- The workflow builds Storybook and pushes the resulting `storybook-static` directory to the `gh-pages` branch
- Once pushed, we should be able to view Storybook live at a url with the pattern `http(s)://<organization>.github.io/<repository>/storybook-static` (so in our case, I think it would be `https://ShelterTechSF.github.io/sheltertech.org/storybook-static` ?)

Resolves #152.
## Testing
Due to the nature of GitHub Action workflows, it's impossible for me to completely test this prior to merging. However, I did decide to try setting up a similar workflow on one of my own repos. The total steps were slightly different (it has a more annoying npm setup stage), but the steps for building Storybook and pushing to GitHub Pages are the same, and they worked just fine. 

Once this is live, we'll want to make sure the following cases are true:
- The workflow triggers whenever we merge into `main`
- The workflow doesn't fail (unless it's because we clearly broke something unrelated)
- The `gh-pages` branch gets an automated commit containing the contents of `storybook-static`
- The `gh-pages` branch is otherwise up to date with `main`
- The GitHub Pages url shows Storybook the same way it does when we view it locally

## Other Notes

- When I did this on my personal repo, I had to enable GitHub Pages in my repository settings. I don't have access to the settings for this repo, so that part is up to @richardxia. The settings page is also where we can definitively see what the url for GitHub Pages is.

- The workflow checks out the `gh-pages` branch using `git checkout -B`, which isn't an operation that I've used before this, but here's my understanding of it:

    - `git checkout -B` does the same operation as `git branch --force`, which effectively resets the branch as if it's newly branched from `main`. Also, if the branch doesn't already exist, it creates a new one much like the `-b` flag. If this description sounds wrong, feel free to school me on this!
 
    - For our purposes this means it always uses the latest `main` without doing any merging or rebasing (a plus in my eyes), but it also means we won't keep old commits performed by this workflow (I am neutral about this, since they're auto-generated anyway, and we still have revision history for the actual source files)
 
    - I'm sure it's possible to set up this workflow such that we could avoid `git checkout -B`, but from my brief tinkering with it, it would add some extra complexity that I'm not sure we need right now. But if we ever want to do other things with the `gh-pages` branch, we will probably need to change this behavior.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -1,0 +1,37 @@
+name: Deploy Storybook
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  deploy:
+    name: Deploy Storybook
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Checkout GitHub Pages branch
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "github-actions@github.com"
+          git checkout -B gh-pages
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Build Storybook
+        run: npm run build-storybook
+
+      - name: Push Storybook to GitHub Pages branch
+        run: |
+          git add storybook-static
+          git commit -m "Build Storybook (auto-generated commit)"
+          git push --set-upstream origin gh-pages --force
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -15,6 +15,13 @@ jobs:
         with:
           node-version: '14'
 
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
       - name: Checkout GitHub Pages branch
         run: |
           git config user.name "GitHub Actions"
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -11,7 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v2-beta
+      - uses: actions/setup-node@v2
         with:
           node-version: '14'
 
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -38,4 +38,4 @@ jobs:
         run: |
           git add storybook-static
           git commit -m "Build Storybook (auto-generated commit)"
-          git push --set-upstream origin gh-pages --force
+          git push origin gh-pages --force
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -15,9 +15,6 @@ jobs:
         with:
           node-version: '14'
 
-      - name: Install dependencies
-        run: npm install
-
       - name: Checkout GitHub Pages branch
         run: |
           git config user.name "GitHub Actions"


richardxia:
> When I did this on my personal repo, I had to enable GitHub Pages in my repository settings. I don't have access to the settings for this repo, so that part is up to @richardxia.

It looks like in order to enable GitHub Pages, you have to provide it a branch to target, and the `gh-pages` branch didn't exist yet, so I created a blank one and pushed it up to GitHub. However, when I pushed it, it looks like it automatically enabled it, so GitHub must have something to automatically detect that branch name, which is how it used to work. In any case, this has been created now: http://sheltertechsf.github.io/sheltertech.org/

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,37 @@
...
+      - main
+  workflow_dispatch:
+jobs:
+  deploy:
+    name: Deploy Storybook
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2-beta
```
richardxia: Do we need to track the `-beta` release branch of this action? I'm not seeing you use anything special here, so I think we can change this to track `@v2` instead of `@v2-beta`.

If you aren't familiar with GitHub Actions, these are just references to Git repos on GitHub itself, with the part after the `@` being any commit hash, branch, or tag name. The `actions` "organization" on GitHub is just where GitHub themselves put official GitHub Actions, and [`setup-node`](https://github.com/actions/setup-node) is just a Git repo. Most of GitHub's actions repos follow a convention where they use semver-ish tag/branch names and where they keep pushing over the tag/branch names that don't specify the full version string. For example, when v2.1.4 was released, they not only pushed a `v2.1.4` tag, but they also pushed over the `v2` tag.

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: npm install
+
+      - name: Checkout GitHub Pages branch
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "github-actions@github.com"
+          git checkout -B gh-pages
+
+      - name: Install dependencies
+        run: npm install
```
richardxia, line 32: It looks like you have this install step twice. If the second copy isn't actually doing something useful, could you delete it?

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2-beta
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: npm install
```
richardxia: It's not super important, since this won't run very often, but you might want to add some caching of the NPM dependencies so that this runs faster. There's a examples of this in the other workflows, but all you need to do is copy the following lines into this workflow file:

https://github.com/ShelterTechSF/sheltertech.org/blob/4fef794442e27e55770c5f2fb871f52925abf399/.github/workflows/deploy-staging.yml#L18-L23

And for what it's worth, these lines were copied verbatim from the documentation of [actions/cache](https://github.com/actions/cache), so we're not doing anything too fancy here: https://github.com/actions/cache/blob/v2.1.3/examples.md#macos-and-ubuntu

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Build Storybook
+        run: npm run build-storybook
+
+      - name: Push Storybook to GitHub Pages branch
+        run: |
+          git add storybook-static
+          git commit -m "Build Storybook (auto-generated commit)"
+          git push --set-upstream origin gh-pages --force
```
richardxia: Nit: The `--set-upstream` is not useful in an automated job like this, since this entire local checkout will get deleted when the GH Actions job finishes running. It's not harmful, but it might be simpler to just omit it.

richardxia (approved):
Thanks for submitting this and for the very detailed writeup in the pull request description! I agree that this will be difficult to test without just merging and running it, so I don't think we need to do that much more testing here.

I have a few, hopefully minor suggestions for improvement here, but I think they can all be made pretty safely, so feel free to merge this without another review once those changes have been made.

I'd also like to suggest we add some description of this to the README, but I think that can wait for this to be fully deployed first so that we make sure we get the final URLs for the site.

> For our purposes this means it always uses the latest main without doing any merging or rebasing (a plus in my eyes), but it also means we won't keep old commits performed by this workflow (I am neutral about this, since they're auto-generated anyway, and we still have revision history for the actual source files)

Yes, I think it's fine to throw out the old commits here. I view this as being a convenience for internal developers at ShelterTech, and we don't need to worry keeping old versions around.

> I'm sure it's possible to set up this workflow such that we could avoid git checkout -B, but from my brief tinkering with it, it would add some extra complexity that I'm not sure we need right now. But if we ever want to do other things with the gh-pages branch, we will probably need to change this behavior.

I think that in an automated workflow that gets a fresh clone each time, `git checkout -B` will have the same behavior as `git checkout -b`, but `-B` will ensure that it works properly even in the event that you are trying to build it from an existing checkout.

.github/workflows/deploy-storybook.yml
jessica-px: `actions/setup-node@v2-beta` is what was being used in the other ymls, so I just left it as-is, but it's an easy change. I changed this to one `v2`, but left the others alone since I'm less familiar with those workflows.

.github/workflows/deploy-storybook.yml
jessica-px: Ahaha woops, that was the result of sloppy copy/pasting on my part.

.github/workflows/deploy-storybook.yml
jessica-px: Thanks for the suggestion -- added!

.github/workflows/deploy-storybook.yml
jessica-px: Done

PR #159, opened 2021-01-04
WIP Typescript Integration

davidagustin:
Still getting this error in storybook but good progress so far.
Screen Shot 2021-01-03 at 7 29 12 PM


src/components/block/ProgramCard/ProgramCard.module.css.d.ts
```diff
@@ -0,0 +1,6 @@
+export const programCardContainer: string;
```
richardxia, line 1: Are we supposed to commit all these files to version control? I feel like this is probably not a good idea because it's just redundant bookkeeping whenever we make a change to the actual CSS files. Do you know how these .d.ts files generated from CSS modules are meant to be used?

src/components/inline/InputText/InputText.js
```diff
@@ -1,8 +1,30 @@
-import PropTypes from "prop-types";
-import React from "react";
-
-import s from "./InputText.module.css";
-
+"use strict";
```
richardxia, line 1: I think we definitely should not be committing the JavaScript file being generated from TypeScript to git. The generated .js code should just be considered a build artifact, not a source file.

src/components/inline/InputText/InputText.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
```
richardxia, line 1: Since the rest of this codebase chooses to use `.js` as the file extension for JSX files, I think we should probably follow that convention (i.e. `.ts`) for TSX files. I personally prefer .jsx and .tsx for JSX/TSX files, but the Gatsby tutorial seems to encourage just using `.js`, 🤷.

src/components/inline/InputText/InputText.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {CSSProperties, useState} from "react";
```
richardxia, line 2: We should make sure that Prettier/ESLint is set up right for TypeScript files. They normally prefer to put a space on the insides of `{` and `}`.

src/components/inline/InputText/InputText.tsx
```diff
@@ -0,0 +1,56 @@
...
+  );
+};
+
+InputText.propTypes = {
+  name: PropTypes.string,
+  placeholderText: PropTypes.string,
+  onChange: PropTypes.func,
+  value: PropTypes.string,
+  type: PropTypes.string.isRequired,
+};
```
richardxia, line 46: We can probably get rid of the prop types when we convert things over to TypeScript, since they're redundant.

storybook-static/iframe.html
```diff
@@ -0,0 +1,74 @@
+<!doctype html><html lang="en"><head><meta charset="utf-8"><title>Storybook</title><meta name="viewport" content="width=device-width,initial-scale=1"><base target="_parent"><style>:not(.sb-show-main) > .sb-main,
```
richardxia, line 1: You accidentally committed all of the generated `storybook-static/` files. I've been meaning to add this directory to the .gitignore, and this reminded me, so I'm going to go ahead and add that now in a separate PR.

package.json
```diff
@@ -44,6 +45,9 @@
     "@storybook/addon-storyshots": "^6.0.26",
     "@storybook/addon-viewport": "^6.0.21",
     "@storybook/react": "^6.0.21",
+    "@types/node": "^14.14.17",
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
```
richardxia, line 49: There may be a few more type packages we should add, for the other third party dependencies that we use.

richardxia (commented):
Thanks for working on this! I think my biggest question right now is how the .d.ts files corresponding to the module.css files are supposed to be handled.

storybook-static/iframe.html
davidagustin: ok

src/components/inline/InputText/InputText.stories.tsx
davidagustin: ok

davidagustin:
> Thanks for working on this! I think my biggest question right now is how the .d.ts files corresponding to the module.css files are supposed to be handled.

This is something to research. I hope to find a good practice that is relevant to our codebase.

src/components/block/ProgramCard/ProgramCard.module.css.d.ts
davidagustin: I've never seen anything like this before. This feels off, but I'm glad I'm presenting it here for further discussion.

src/components/inline/InputText/InputText.js
davidagustin: I guess until all the javascript files turns into typescript. Careful staging of files must be stressed. Then we can put a regular expression in the .gitignore right?

src/components/inline/InputText/InputText.stories.tsx
davidagustin: ok

src/components/inline/InputText/InputText.tsx
davidagustin: ok

package.json
davidagustin: I'll need your help in careful evaluation of this.

storybook-static/iframe.html
richardxia: I lied in my original comment and didn't submit a PR until just now: https://github.com/ShelterTechSF/sheltertech.org/pull/160

But please make sure you delete these files from your branch, even if you merge/rebase the `main` branch after that PR merges, since the `.gitignore` changes won't retroactively remove things that have been committed.

davidagustin: ok

src/components/inline/InputText/InputText.js
richardxia: It still feels a bit too error prone to me to have to carefully add only the .tsx files and not the generated .js files, since the .js files end up being generated in the same directory as the other files, so I'm almost certain all of us will accidentally commit them as we're working on stuff.

I'll try to do some research on different ways people do this for TypeScript codebases, since the way this plugin works right now seems pretty weird to me.

jessica-px: Off the top of my head I think there are two ways we can resolve this:
1. We can use the `outDir` config option to output all compiled files into one directory, which we can then add to gitignore. Check out [this stackoverflow conversation](https://stackoverflow.com/questions/24454371/how-can-i-get-the-typescript-compiler-to-output-the-compiled-js-to-a-different-d) about it.
2. We can use the `noEmit` config option, which stops it from outputting compiled .js files in the first place. I think this would work fine for us, since we're not relying on Typescript itself for final compilation anyway (Gatsby handles that, with webpack/babel under the hood).

But I don't think I see a `tsconfig.json` file in this PR? We're going to need one to configure things like this for Typescript. Adding one is really easy -- Gatsby even seems to provide [their own example file](https://github.com/gatsbyjs/gatsby/blob/master/examples/using-typescript/tsconfig.json), so copy/pasting that would probably make a good starting point. And for understanding the different options, I like [this resource](https://medium.com/javascript-in-plain-english/typescript-configuration-options-tsconfig-json-561d4a2ad4b).

jessica-px: Hah -- sorry to overlap you on this one, Richard!

richardxia: Hah, no problem, looks like we posted it at right about the same time! I think you have more experience than I do on this, @jessypeck, but I think these sound like good paths to try out.

Note that I think we may still want the CSS modules files to create `.d.ts` files, since they'll be needed in order for TypeScript to understand the types of those CSS files when import, but if there's a way for the tools to understand the CSS files as TypeScript files without needing to generate the files, that'd be great.

davidagustin: Yeah, I think the .d.ts files can get too messy with our code. I'm researching some alternative for it.

davidagustin:
After spending a considerable amount of time (with @richardxia) doing the typescript integration. I have found these truths self evident, that the plugin "gatsby-plugin-typescript" does not do type checking ( See: https://www.gatsbyjs.com/plugins/gatsby-plugin-typescript/ ). The plugin "gatsby-plugin-typescript" runs another plugin which is "babel-plugin-transform-typescript" which transforms typescript code to javascript but does not do type checking.

davidagustin:
Too messy and trying to make a cleaner WIP push

PR #160, opened 2021-01-07
Add storybook-static dir to .gitignore.

richardxia:
This adds the `storybook-static/` directory to .gitignore so that we don't accidentally commit it during normal development.

Note that I had to make a small modification (https://github.com/ShelterTechSF/sheltertech.org/commit/5fc4c0038a830dae2b6fcd3af211c49df5fda3d4) to @jessypeck's GitHub Action for building and deploying Storybook, since the `git add` command now needs the `--force` flag to tell it to add the directory, even though would normally be ignored. I didn't test that, but I tested the `git add --force` command locally to check that it will add the `storybook-static/` files.

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -36,6 +36,8 @@ jobs:
 
       - name: Push Storybook to GitHub Pages branch
         run: |
-          git add storybook-static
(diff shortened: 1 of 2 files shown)


jessica-px (approved):
Didn't test, but should be fine to merge

richardxia:
Thanks for reviewing!

PR #161, opened 2021-01-18
Typescript Integration

### LINKED ISSUE #151: Set up TypeScript compiler with our build system
This should be configured for both the Gatsby build and the Storybook build.

The way we can test that this has been done correctly is to convert the inline [`<InputField>`](https://github.com/ShelterTechSF/sheltertech.org/tree/4fef794442e27e55770c5f2fb871f52925abf399/src/components/inline/InputText) field component to TypeScript and check that it builds both in Gatsby and Storybook.

davidagustin:
Continuation from https://github.com/ShelterTechSF/sheltertech.org/pull/159 alot of d.ts files removed and ignored here


davidagustin:
A better plugin to replace "gatsby-transformer-typescript-css-modules" may be needed because a modules loader within the plugin generates a .d.ts file alongside every css file. See https://www.gatsbyjs.com/plugins/gatsby-transformer-typescript-css-modules/ 

 An alternative npm package I found is https://github.com/Quramy/typed-css-modules

gatsby-config.js
```diff
@@ -9,6 +9,8 @@ module.exports = {
   },
   plugins: [
     "gatsby-plugin-react-helmet",
+    "gatsby-plugin-ts-checker",
+    "gatsby-plugin-typescript",
```
davidagustin: "gatsby-plugin-typescript" does not do typecheck but "gatsby-plugin-ts-checker" does and is more current for type checking for gatsby .

.gitignore
```diff
@@ -22,6 +22,7 @@ jspm_packages/
 # gatsby files
 .cache/
 public
+*.css.d.ts
```
davidagustin: Workaround but not a good solution for the current problem of every .css.d.ts file generated in the same folder of a css file

.eslintrc.js
```diff
@@ -54,7 +57,7 @@ module.exports = {
         "no-restricted-syntax": "off",
       },
       env: {
-        // Jest injects itself as a varibable in test files
+        // Jest injects itself as a variable in test files
```
davidagustin, line 89: Typo correction

.eslintrc.js
```diff
@@ -32,7 +35,7 @@ module.exports = {
     },
     {
       // Storybook stories
-      files: ["src/**/*.stories.js"],
+      files: ["src/**/*.stories.js", "src/**/*.stories.tsx"],
```
davidagustin: For storybook but needs to be check this implementation is actually needed

package.json
```diff
@@ -5,8 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "jest",
-    "start": "gatsby develop",
-    "build": "gatsby build",
+    "start": "npx tsc && gatsby develop",
+    "build": "npx tsc && gatsby build",
```
davidagustin: I added this here so please be aware

package.json
```diff
@@ -72,6 +78,7 @@
     "react-test-renderer": "^16.14.0",
     "stylelint": "^13.6.1",
     "stylelint-config-sass-guidelines": "^7.0.0",
-    "stylelint-config-standard": "^20.0.0"
+    "stylelint-config-standard": "^20.0.0",
+    "typescript": "^4.1.3"
   }
 }
```
davidagustin, line 85: An npm audit fix was done to work around some errors

src/components/inline/InputText/InputText.tsx
```diff
@@ -0,0 +1,55 @@
...
+  type: PropTypes.string.isRequired,
+};
+
+InputText.defaultProps = {
+  name: undefined,
+  placeholderText: undefined,
+  onChange: undefined,
+  value: undefined,
+};
+
```
davidagustin: Prettier for typescript and possibly other tools may need to be added still

tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "esModuleInterop": true,
+    // Types should go into this directory.
+    // Removing this would place the .d.ts files
+    // next to the .js files
+    "outDir": "dist"
+  },
+  "files": ["./src/typings/declarations.d.ts"],
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "src/legacy"]
+}
```
davidagustin: These are all experimentations and possibly all or none of these options can be used.

davidagustin (commented):
npm audit fix was made to successfully build

package.json
```diff
@@ -27,7 +27,7 @@
   },
   "homepage": "https://github.com/ShelterTechSF/sheltertech.org#readme",
   "dependencies": {
-    "gatsby": "^2.24.47",
+    "gatsby": "^2.30.3",
```
davidagustin: Gatsby was also updated to workaround some errors

package.json
richardxia: Could you split that out into a different PR? Due to the number of new dependencies being added, it's difficult to tell which locked dependencies changed due to `npm audit fix` and which locked dependencies changed due to adding brand new dependencies.

This separate PR can be merged earlier if we test that the changes don't cause any regressions.

gatsby-config.js
richardxia: Do we still need both plugins, or is `ts-checker` sufficient? Regardless of the answer to that question, it might be nice to add a comment here describing the role and purpose of the plugins, since it wasn't trivial even finding the correct plugins in the Gatsby ecosystem.

package.json
richardxia: Could we upgrade Gatsby in a separate PR, possibly in the same separate PR as the `npm audit fix`? Upgrading a core framework like this is a higher-risk activity, especially since Gatsby has an extremely large dependency tree. It'll help mitigate risk to do this separately from adding TypeScript to the codebase so that we don't mix issues caused by upgrading existing libraries with issues caused by integrating TypeScript.

package.json
richardxia: I don't think this is going to do what we want it to do, at least for `gatsby develop`. The issue is that `gatsby develop` spins up a server that runs until it is manually stopped, so the `npx tsc` command will only run once on initial startup. This probably is not an issue for `gatsby build`, since that's a command that doesn't start a server.

It's probably a good idea to have a standalone npm script that just runs the TypeScript type checking by itself, which we can add to CI to run alongside the lint checks.

In addition, if it's possible, I think it'd be nice to more tightly integrate the TypeScript type checking directly into the Gatsby/webpack development server, since that is already watching for changes to files and rebuilding them. That said, I'm not personally familiar with how people normally set this stuff up, and I don't know if people generally rely on their IDE to get live feedback on TypeScript errors (as a vim user, I'm not accustomed to using my editor like that).

---

On a separate note, can you drop the `npx` from the command? It's unnecessary because `tsc` is already installed to `node_modules/.bin`, and all of the stuff in that directory is already available to npm scripts.

src/components/inline/InputText/InputText.tsx
richardxia: Yes, I agree with this, and I think we should make sure all of our existing tools have been properly configured for TypeScript before we merge.

src/components/inline/InputText/InputText.tsx
```diff
@@ -0,0 +1,55 @@
+import * as PropTypes from "prop-types";
```
richardxia: This is a comment about the file, but not this particular line.

1. Could we drop the `x` in `.tsx`? The default Gatsby setup for non-TypeScript projects omits the `x` in `.jsx`, so we probably should just follow the same convention for TypeScript.
2. Can you actually delete `InputText.js` and make sure that the existing site code is importing this new TypeScript version of the file?
3. Can we actually start rewriting this component to use types? I think it's important that we actually have a real, type-annotated file as part of this first PR to integrate the TypeScript into the codebase. I think it's appropriate to drop PropTypes as we do this, since we won't need PropTypes after the whole codebase is updated.

src/typings/declarations.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*css' {
```
richardxia: What's the purpose of this file, and how does it interact with the `gatsby-transformer-typescript-css-modules` that you added to `gatsby-config.js`? If the purpose of that plugin is to generate `.d.ts` files for all of our CSS modules, then I'm not sure why we need this module declaration as well.

tsconfig.json
richardxia: I think most of these settings look reasonable to me. I think it'd be a good idea, however, for us to go through the available options and decide which ones we want to enable/disable. Since I haven't really worked in a TypeScript project before, I'm not really familiar with all the options, but I do know about the [strict null checks option](https://www.typescriptlang.org/tsconfig#strictNullChecks) and I absolutely think we should enable that one.

@jessypeck, do you have any thoughts on a good set of options to enable? My preference is to err on the side of being overly strict, since it's easier to relax them over time than to increase strictness.

richardxia (commented):
> A better plugin to replace "gatsby-transformer-typescript-css-modules" may be needed because a modules loader within the plugin generates a .d.ts file alongside every css file. See https://www.gatsbyjs.com/plugins/gatsby-transformer-typescript-css-modules/

Is there a way to configure where that plugin outputs the `.d.ts` files, similar to the `tsconfig.json` `outDir` configuration parameter?

---

Thanks for figuring all this stuff out and getting us closer! I do think it's important to get a better solution for the CSS modules `.d.ts` file thing. I also think that making sure we get all the other tooling such as ESLint and Prettier set up for TypeScript is important.

src/components/inline/InputText/InputText.tsx
richardxia: It looks like it's non-trivial to drop the `x` on `.tsx`, since the TypeScript compiler relies on it in order to decide whether to use TSX parsing mode or normal TS parsing mode. Apparently the angle brackets subtly conflict with the generic type parameters (e.g. `MouseEvent<HTMLInputElement>`), so it makes the parsing harder when you don't know which syntax type is used in the file.

I'm fine with standardizing on .tsx for this whole codebase, despite Gatsby's convention of trying to omit the `x`.

src/components/inline/InputText/InputText.tsx
```diff
@@ -24,7 +23,20 @@ import s from "./InputText.module.css";
  *
  */
 
-const InputText = ({ name, placeholderText, onChange, value, type }) => {
+type InputTextProps = {
```
jessica-px, line 26: For the purposes of defining React props, type aliases and interfaces are functionally interchangeable, so although I don't think we have a style guide yet, I think when we do create it, "we prefer to use type aliases over interfaces when possible" would be the kind of thing I'd want to see there. It doesn't make a difference to me, I'd just like us to pick one and stick with it.

richardxia:
I decided to push up one more commit to update the Command Line Reference doc: https://github.com/ShelterTechSF/sheltertech.org/pull/161/commits/f7350f9cf019af027dcd069efdc5cc63d498685f.

Aside from that, this is now ready for a final review, as TypeScript has now been fully integrated into the following technologies that we use:

- Gatsby
- Storybook
- ESLint
- Prettier
- Jest

This was a lot more challenging to work on than expected, due to the large number of other tools/frameworks we're using and due to the complex interoperability between them, but I think this is finally now all working.

I encourage everyone to nitpick about anything and everything, since it's best that we do it now than before we migrate the rest of the codebase to TypeScript!

jessica-px (approved):
Checked it out and was able to get both the dev server and storybook to work, LGTM 👍

src/components/inline/InputText/InputText.tsx
richardxia: Yeah, I was wondering about that when I typed this line. Since I haven't had much real world experience here, I don't feel that I have a good sense of which to use. In other languages, I would normally prefer interfaces for anything that is actually used by another file, and aliases only for things where I want to save typing, but I think TypeScript's aliases are more powerful than type aliases in other languages.

I've found https://github.com/typescript-cheatsheets/react to be a pretty useful resource for best practices, and they have a specific section on [types vs. interfaces](https://github.com/typescript-cheatsheets/react/tree/cfa1209cdf926ab5894f0ef150c2fe48cce06a6b#types-or-interfaces). Looks like their rule of thumb is to use `type` for React props and state, but it sounds like they aren't dogmatic about it because there are causes where interfaces have advantages over types.

If that sounds reasonable to you, I think we can go with `type`s by default for props/state, but we can feel free to use `interface`s when it's useful.

jessica-px: Yeah I'm happy using `type` by default -- I don't have strong opinions here, but I lean more towards `type` than I do against it.

I do think it can be confusing for people coming in from other languages, and older React/Typescript documentation and guides from before `type` was added still use `interface`. But as long as we write down somewhere that this was a style choice we made, I don't foresee any problems.

package.json
```diff
@@ -72,6 +79,7 @@
     "react-test-renderer": "^16.14.0",
     "stylelint": "^13.6.1",
     "stylelint-config-sass-guidelines": "^7.0.0",
-    "stylelint-config-standard": "^20.0.0"
+    "stylelint-config-standard": "^20.0.0",
+    "typescript": "^4.1.3"
   }
 }
```
davidagustin, line 85: I remember I said I did an npm audit fix in a previous push and may need to be reverted for a later PR, was that addressed?

src/components/inline/InputText/InputText.tsx
richardxia: Sounds good! Let me leave this PR open for another day, in case if anyone one else wanted to do a review on it, but if there are no other comments, I can merge it in tomorrow.

davidagustin:
Other than that,  approve this PR

package.json
richardxia: I didn't do anything specific to try to undo it, since it's not easy to undo an `npm audit fix` when it's mixed in with actual changes to package.json. `npm audit fix` will generally only change the package-lock.json file to bump any packages to newer versions with security fixes, since the vast majority of packages will only be listed in package-lock.json.

Since everything seems to be building fine, I'm inclined to not touch the package-lock.json file any further and just merge this as is. Thanks for reminding me, though, since I had forgotten!

PR #190, opened 2021-02-05
Convert Button component to TypeScript

### LINKED ISSUE #164: Convert Button component to TypeScript

ivanleeswe:
Convert button component to TypeScript

### DIFF
--- a/src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
+++ b/src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
@@ -61,7 +61,13 @@ const ParagraphBlock = ({ title, description, button }) => {
 ParagraphBlock.propTypes = {
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
-  button: PropTypes.shape(Button.propTypes),
+  button: PropTypes.shape({
+    text: PropTypes.string.isRequired,
+    noHover: PropTypes.bool,
+    externalLink: PropTypes.string.isRequired,
+    internalLink: PropTypes.string.isRequired,
+    onClick: PropTypes.func.isRequired,
+  }),
 };
 
 ParagraphBlock.defaultProps = {
@@ -85,7 +91,13 @@ const CTABlock = ({ title, buttons }) => (
 
 CTABlock.propTypes = {
   title: PropTypes.string.isRequired,
-  buttons: PropTypes.arrayOf(Button.propTypes).isRequired,
+  buttons: PropTypes.arrayOf({
+    text: PropTypes.string.isRequired,
+    noHover: PropTypes.boolean,
+    externalLink: PropTypes.string.isRequired,
+    internalLink: PropTypes.string.isRequired,
+    onClick: PropTypes.func.isRequired,
+  }).isRequired,
 };
 
 /* Main component */
@@ -207,7 +219,13 @@ ThreeParagraphBlock.propTypes = {
     alt: PropTypes.string,
   }),
   ctaTitle: PropTypes.string,
-  ctaButtons: PropTypes.arrayOf(Button.propTypes),
+  ctaButtons: PropTypes.arrayOf({
+    text: PropTypes.string.isRequired,
+    noHover: PropTypes.boolean,
+    externalLink: PropTypes.string.isRequired,
+    internalLink: PropTypes.string.isRequired,
+    onClick: PropTypes.func.isRequired,
+  }),
 };
 
 ThreeParagraphBlock.defaultProps = {
--- a/src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
+++ b/src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
@@ -61,7 +61,7 @@ const ParagraphBlock = ({ title, description, button }) => {
 ParagraphBlock.propTypes = {
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
-  button: PropTypes.arrayOf({
+  button: PropTypes.shape({
     text: PropTypes.string.isRequired,
     noHover: PropTypes.bool,
     externalLink: PropTypes.string.isRequired,
@@ -219,13 +219,7 @@ ThreeParagraphBlock.propTypes = {
     alt: PropTypes.string,
   }),
   ctaTitle: PropTypes.string,
-  ctaButtons: PropTypes.arrayOf({
-    text: PropTypes.string.isRequired,
-    noHover: PropTypes.bool,
-    externalLink: PropTypes.string.isRequired,
-    internalLink: PropTypes.string.isRequired,
-    onClick: PropTypes.func.isRequired,
-  }),
+  ctaButtons: PropTypes.arrayOf(Button.propTypes),
 };
 
 ThreeParagraphBlock.defaultProps = {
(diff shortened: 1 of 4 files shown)


src/components/inline/Button/Button.tsx
```diff
@@ -1,10 +1,23 @@
...
 import s from "./Button.module.css";
 
-const Button = ({ text, externalLink, internalLink, onClick, noHover }) => {
+type ButtonProps = {
+  text: string;
+  externalLink: string;
+  internalLink: string;
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  noHover: boolean;
+};
```
richardxia: Ah, so here you can actually express this with a more advanced type declaration in TypeScript. Even before throwing TypeScript into the mix, if you look at the original PropTypes that @davidagustin added on this component, you can see that it's doing something more complicated with `PropTypes.oneOfType()` and `PropTypes.exact()`. This basically allows us to say that either a Button can be passed in `externalLink` or an `internalLink` or an `onClick` function, but you can't have any of them simultaneously. That is, it's an error to pass in both `externalLink` and `internalLink` at the same time to any one Button component.

TypeScript has support for describing constraints like this in an even more powerful way than PropTypes with its [union and intersection types](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html).

Without going too much into the theory of how this works, this basically allows you to write "and" (&) and "or" (|) expressions to combine two existing types. For this component, I think you'll want something that looks like this:

```suggestion
type ButtonProps = { text: string, noHover: boolean } & (
  { externalLink: string }
  | { internalLink: string }
  | { onClick: (e: React.MouseEvent<HTMLButtonElement>) => void }
);
```

You'll also want to change the actual body of the `Button` component a bit, since TypeScript will allow you to eliminate the `throw new Error()` case at the end, since TypeScript can guarantee that the existing three `if` statements will all possible things you can pass into the component, and therefore guarantee that you will never hit the `else` case at the end.

I'm not exactly certain how this will interact with the destructuring we're doing in the function argument definition, so you might need to replace the arguments with an non-destrictured `props` argument and do the destructuring later, after checking for which of the three cases we have.

Let me know if you wanted me to walk you through this in a call, since it's a bit complicated.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -61,7 +61,13 @@ const ParagraphBlock = ({ title, description, button }) => {
...
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
-  button: PropTypes.shape(Button.propTypes),
+  button: PropTypes.shape({
+    text: PropTypes.string.isRequired,
+    noHover: PropTypes.bool,
+    externalLink: PropTypes.string.isRequired,
+    internalLink: PropTypes.string.isRequired,
+    onClick: PropTypes.func.isRequired,
+  }),
```
richardxia: You'll actually have to inline the original definition that had the `PropTypes.oneOf()` expression in there, or else everything will fail this because you've marked all the fields as required.

If you want a quicker way to run the tests without having to push code up to GitHub, you should be able to just run them with `npm run lint` locally.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
```diff
@@ -61,7 +61,13 @@ const ParagraphBlock = ({ title, description, button }) => {
 ParagraphBlock.propTypes = {
   title: PropTypes.string.isRequired,
   description: PropTypes.string.isRequired,
-  button: PropTypes.shape(Button.propTypes),
+  button: PropTypes.arrayOf({
```
richardxia: I think you accidentally changed this from `PropTypes.shape()` to `PropTypes.arrayOf()`. You'll need to be careful and continue to use `.shape()` in the places that had `.shape()` before and `.arrayOf()` in the places that had `.arrayOf()` before.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.js
ivanleeswe: `const ParagraphPropType = PropTypes.shape({
  button: PropTypes.oneOfType([
    PropTypes.exact({
      text: PropTypes.string,
      externalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      internalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      onClick: PropTypes.func,
    }),
  ]),
});
#`

ivanleeswe: const ParagraphPropType = PropTypes.shape({
  button: PropTypes.oneOfType([
    PropTypes.exact({
      text: PropTypes.string,
      externalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      internalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      onClick: PropTypes.func,
    }),
  ]),
});

ivanleeswe: > const ParagraphPropType = PropTypes.shape({
  button: PropTypes.oneOfType([
    PropTypes.exact({
      text: PropTypes.string,
      externalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      internalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      onClick: PropTypes.func,
    }),
  ]),
});

ivanleeswe: > const ParagraphPropType = PropTypes.shape({
  button: PropTypes.oneOfType([
    PropTypes.exact({
      text: PropTypes.string,
      externalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      internalLink: PropTypes.string,
    }),
    PropTypes.exact({
      text: PropTypes.string,
      onClick: PropTypes.func,
    }),
  ]),
});

ivanleeswe: So above is the original definition with PropTypes.oneOf() expression, do I add the whole expression or just add the properties text, internalLink, externalLink, and onClick?

> ParagraphBlock.propTypes = {
  button: PropTypes.shape({
    text: PropTypes.string,
    externalLink: PropTypes.string,
    internalLink: PropTypes.string,
    onClick: PropTypes.func,
  }),
};

ivanleeswe: @richardxia thank you in advance

richardxia: It's a little bit hard to read, but I think you'll want to do what you had in your first version.

BTW, when you want to paste a snippet of code in a comment, you should use triple backticks (```) so that the indentation is preserved and so that the text gets formatted in a monospaced font. See https://docs.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks#fenced-code-blocks for the documentation about this.

richardxia (approved):
Awesome, thank you for fixing up all the prop types in the other components, and sorry again for not realizing ahead of time that this would happen. Let's get this in, and then we can move on to migrating these other components to TypeScript as well!

candywang (approved):
LGTM!

richardxia:
Hey @ivanleeswe, a bit of bad news, it looks like this change made all the buttons stop functioning when viewed in the Gatsby site. You can see this on the staging site at https://staging.sheltertech.org/.

I'll create a ticket to track fixing it, but could you look into why the buttons aren't working anymore? And sorry for not thinking to check the Gatsby site when I was reviewing this PR. Since this PR contained substantial enough changes, it probably warranted more thorough testing.

PR #192, opened 2021-02-22
Convert Program card component to TypeScript

### LINKED ISSUE #167: Convert ProgramCard component to TypeScript

candywang:
In this PR,

- Converted component to TypeScript.
- In the `tsconfig` file, added  `"esModuleInterop": true` in `compilerOptions` to ["import CommonJS modules in compliance with es6 modules spec"](https://stackoverflow.com/questions/56238356/understanding-esmoduleinterop-in-tsconfig-file).
  - **Before**: `import * as React from "react";`
  - **After**: `import React from "react";`

- This resolves #167.


tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "allowJs": true, // Allow a mix of JS and TS until we fully migrate to TS.
     "jsx": "react",
     "noEmit": true, // No need to emit .d.ts files, since we're not publishing a package.
+    "esModuleInterop": true,
```
jessica-px: Can you add a comment explaining what this does?

jessica-px (approved):
LGTM!

tsconfig.json
richardxia: Actually, could we not turn this on (at least not yet)? It looks like moving forward, with React 17 and onwards, [the official stance is to not have a default export from the React module anymore](https://github.com/facebook/react/pull/18102). See the announcement blog post for some more context for it: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

Note that if we upgrade to React 17, we can actually remove the `import React from 'react'` statement from most of our files, since we'll only need an import from React if we use some specific function from React, such as `useState`.

richardxia (approved):
Left a comment about the `esModuleInterop` thing, but this otherwise LGTM!

ivanleeswe (approved):

PR #193, opened 2021-03-01
Convert Title Block component to TypeScript

### LINKED ISSUE #183: Convert TitleBlock component to TypeScript

ivanleeswe:
Convert Title Block properties to TypeScript

### DIFF
--- a/src/components/grid-aware/TitleBlock/TitleBlock.tsx
+++ b/src/components/grid-aware/TitleBlock/TitleBlock.tsx
@@ -1,9 +1,12 @@
-import PropTypes from "prop-types";
-import React from "react";
+import * as React from "react";
 
 import s from "./TitleBlock.module.css";
 
-const TitleBlock = ({ title }) => {
+type TitleBlockProps = {
+  title: string;
+};
+
+const TitleBlock = ({ title }: TitleBlockProps) => {
   return (
     <div className={s.bleedWrapper}>
       <div className={s.bleedMainContent}>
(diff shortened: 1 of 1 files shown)


richardxia (approved):
LGTM!

src/components/grid-aware/TitleBlock/TitleBlock.tsx
```diff
@@ -1,9 +1,12 @@
-import PropTypes from "prop-types";
-import React from "react";
+import * as React from "react";
 
 import s from "./TitleBlock.module.css";
 
-const TitleBlock = ({ title }) => {
+type TitleBlockProps = {
```
davidagustin, line 5: @richardxia Do we care about grammar? There is only one property in here yet the name is plural

jessica-px: I feel like `props` is generally used to mean "the name of the object passed to a React component", regardless of how many properties it actually contains. Singular `prop` would, to me, suggest a specific key/value pair within the `props` object. I'd be thrown off if I saw it used to refer to the object itself, so I vote for keeping this plural by convention.

(It also makes it so we don't have to make trivial grammar updates when we add/remove props).

jessica-px (approved):

src/components/grid-aware/TitleBlock/TitleBlock.tsx
richardxia: Yeah, I agree with @jessypeck on both points. It's kind of like how the `children` prop on components is always `children` and not `child` even when there's only one child element passed in.

davidagustin: LGTM then!

davidagustin (approved):

candywang (approved):
LGTM!

PR #195, opened 2021-03-01
Convert Spacer component to TypeScript

### LINKED ISSUE #181: Convert Spacer component to TypeScript

ivanleeswe:
Convert Spacer component to TypeScript


src/components/grid-aware/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,28 @@
...
+  heightMobile,
+  color = "transparent",
+}: SpacerProps) => (
+  <div
+    style={
+      {
+        "--spacer-height-desktop": heightDesktop,
+        "--spacer-height-mobile": heightMobile,
+        "--spacer-color": color,
+      } as React.CSSProperties
```
richardxia, line 22: Ah, this is interesting. I hadn't thought about how the CSS custom properties would interact with TypeScript. I think this approach is OK for CSS custom properties that are local to just that component, but if we ever find ourselves using custom properties across multiple components, then we may want to try out an approach closer to https://stackoverflow.com/a/65959390, where they defined a custom interface that extends the `CSSProperties` interface.

richardxia (approved):
LGTM!

src/components/grid-aware/Spacer/Spacer.tsx
davidagustin: Interesting, forgot about this mechanic also.

davidagustin (approved):

davidagustin (approved):

jessica-px (approved):

candywang (approved):
LGTM!

PR #201, opened 2021-03-07
Consolidate button property types

### LINKED ISSUE #200: Simplify Button component's props
We're starting to hit a number of issues due to the complexity of the Button component's prop's type definition, so to simplify things, let's collapse all of the union types into a single type but with optional properties.

Specifically, we should take the type definitions for the props in https://github.com/ShelterTechSF/sheltertech.org/blob/0956cede535dda961b57c544b1a4d2dcc7b50b3a/src/components/inline/Button/Button.tsx#L6-L24 and combined them into a single `ButtonProps` type that has all the properties in it.

ivanleeswe:
Fixes #200 by changing properties externalLink, internalLink, and onClick to be optional. This change allows our 3 if statements to check if any one of them is true and if it is, it will return the body of the if statement.

### DIFF
--- a/src/components/inline/Button/Button.tsx
+++ b/src/components/inline/Button/Button.tsx
@@ -3,32 +3,24 @@ import * as React from "react";
 
 import s from "./Button.module.css";
 
-type CommonProps = {
+type ButtonProps = {
   text: string;
   noHover?: boolean;
+  externalLink?: string;
+  internalLink?: string;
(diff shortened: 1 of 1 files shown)


richardxia (approved):
Great, thanks, @ivanleeswe, for doing this! Hopefully this is the last time we have to touch this file before we convert the rest of the site over to TypeScript!

@candywang, once this merges in, this should hopefully allow you to get past the TypeScript errors that you were hitting on your component.

PR #205, opened 2021-03-28
Convert PartnersAndSponsorsBlock component to TypeScript

### LINKED ISSUE #179: Convert PartnersAndSponsorsBlock component to TypeScript

candywang:
Closes #179


jessica-px (approved):
LGTM!

richardxia (approved):
LGTM!

BTW, this reminds me that the image prop type, consisting of a `url` and an `alt`, is something we could probably define as a TypeScript type now, which is something we weren't able to do previously in https://github.com/ShelterTechSF/sheltertech.org/pull/100/files since the PropTypes linter rule was unable to resolve props defined in other files.

This is something we can try to resurrect after we wrap up the TypeScript conversion.

candywang:
@richardxia Definitely was thinking the same! 🤓

PR #207, opened 2021-03-29
Convert Navigation component to TypeScript

### LINKED ISSUE #177: Convert Navigation component to TypeScript

richardxia:
Closes #177.

Note: This is soft-blocked on a [PR I submitted upstream to DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52053) to update the type definitions for @types/react-burger-menu. The latest version of those types don't include a definition for [`itemListElement`](https://github.com/negomi/react-burger-menu/blob/v2.8.0/src/menuFactory.js#L374), which we are using to control the HTML element used for the list component. Assuming that my PR is accepted, once they publish a new version of the types, I can add @types/react-burger-menu to our app's dependency list. Until then, this PR will fail our CI checks, since it won't pass the type checker. If this takes too long, then I can find some other way to work around the issue until my upstream PR is accepted.

As for this PR, aside from the usual PropTypes -> TypeScript conversions, there are a couple things of note:

1. I was running into type errors with my SVG imports, since it is really Webpack's file-loader/url-loader that allows us to import those kinds of files into JavaScript. Fortunately, this is specifically documented in one of [Webpack's guides to integrating with TypeScript](https://webpack.js.org/guides/typescript/#importing-other-assets), so I added that shim to our existing shims file in https://github.com/ShelterTechSF/sheltertech.org/commit/b3dc2f3f40b36cad8f50deb06524a285da67a45c. Also, I renamed our shims file from `styles.d.ts` to `shims.d.ts` to better reflect that there are shims for more than just the CSS modules stuff in there.
2. The way we are handling the "themes" for the navigation components doesn't make TypeScript happy. Here's a short explanation of what we're doing: We have a `<NavList>` component that is shared between our desktop navigation (`<Navigation>`) and our mobile navigation (`<BurgerMenu>`). The styles are slightly different between the two, but we managed to reuse the same HTML structure by defining two separate stylesheets and switching between them by passing the imported CSS module as a `theme` prop on the `<NavList>` component.

    TypeScript is unhappy about this because TypeScript doesn't actually know about the CSS classes we are defining in our CSS files. We currently have a shim that just says that the imported module is a key-value pair of string-strings. This works in most files, but here the extra level of indirection between when we import the CSS modules and when we pass them into the `<NavList>` component as a prop causes ESLint errors of the following form:

    ```
    22:24  error  Invalid type "any" of template literal expression  @typescript-eslint/restrict-template-expressions
    22:24  error  Unsafe member access .navLink on an any value      @typescript-eslint/no-unsafe-member-access
    ```

    When I ran into the errors above, I had assigned the `theme` prop a type of `any`, and the errors complain about trying to access a property on an `any`, which is indeed something one should avoid. Instead of just silencing those ESLint rules, I figured that actually fully specifying the CSS class names we expect to find in the `<NavList>` would serve as better documentation. This however only pushed the typing problem up a level into the `Navigation.tsx` and `BurgerMenu.tsx` file, since TypeScript still doesn't like passing in a `{ [key: string]: string }` to an argument of a more specific type. At this point, I just put type assertions of the form `as ThemeType` in the call sites.

    Any thoughts on whether this is a good idea? Is this overcomplicating things, and should I just disable the ESLint errors for the scope of the `NavList.tsx` file?


jessica-px:
> Any thoughts on whether this is a good idea? Is this overcomplicating things, and should I just disable the ESLint errors for the scope of the `NavList.tsx` file?

🤔 I do think it's a little more complex than I'd like, especially since these are pretty simple components. Honestly I'm wondering if it's even worth the trade-off of passing around a `theme` to make `<NavList>` reusable. We might end up with a little bit of redundancy, but I think if we just wrote separate components for the desktop/mobile views, it'd be easier to reason about and maintain.

davidagustin (dismissed):
LGTM

package.json
```diff
@@ -46,6 +46,7 @@
     "@storybook/addon-viewport": "^6.0.21",
     "@storybook/react": "^6.0.21",
     "@types/react": "^17.0.0",
+    "@types/react-burger-menu": "^2.8.0",
```
davidagustin, line 49: How is this package needed when a type was created for the burger menu already?

richardxia:
> 🤔 I do think it's a little more complex than I'd like, especially since these are pretty simple components. Honestly I'm wondering if it's even worth the trade-off of passing around a `theme` to make `<NavList>` reusable. We might end up with a little bit of redundancy, but I think if we just wrote separate components for the desktop/mobile views, it'd be easier to reason about and maintain.

Yeah, that sounds reasonable to me, and with TypeScript, we're able to make the input data the same between both components, which is probably the more important thing.

Let me "unfactor" the NavList component to see what that ends up looking like.

package.json
richardxia: I'm not exactly sure what you're asking here, but just to be clear, we're using the `react-burger-menu` library to implement our own `<BurgerMenu>` component, which just wraps the `react-burger-menu`'s `push` component (renamed to `Menu` in the import statement). We need type declaration on both the library we're using as well as our own component, since both are instantiated in our code.

richardxia:
> Let me "unfactor" the NavList component to see what that ends up looking like.

Just pushed up a commit where I copied the NavList components into the BurgerMenu and Navigation components, and where I collapsed the "theme" CSS files into the existing BurgerMenu and Navigation CSS modules. LMK what you think!

jessica-px (approved):
I like this refactor, LGTM! 👏

PR #208, opened 2021-03-29
Convert FooterCTACard component to TypeScript

### LINKED ISSUE #166: Convert FooterCTACard component to TypeScript

candywang:
Closes #166.


richardxia (commented):
@candywang, it looks like you're currently failing the CI tests. My guess is that the previous `ButtonPropType` didn't actually conform to our usual "Button" interface, since it had a `link` property instead of one of the `internalLink`, `externalLink`, `onClick` properties.

I think it's reasonable to change this to match our common Button interface, but this means you'll need to update this both in the Storybook tests as well as on the actual Footer component and any pages that set that property. In the future, once we migrate everything to TypeScript, this would be caught by a TypeScript type error rather than our automated tests.

candywang:
> Footer component and any pages that set that property

Thanks for pointing this out @richardxia ! I made the change and it's passing the tests now. ~~Would you mind elaborating on the changes needed in the `Footer` component?~~ Noted for the `Footer` component.

richardxia (approved):
LGTM!

> Thanks for pointing this out @richardxia ! I made the change and it's passing the tests now. ~Would you mind elaborating on the changes needed in the Footer component?~ Noted for the Footer component.

Hmm, I just realized/remembered that we had actually taken this component out of the Footer before we launched the site, so it's not actually being used on the live site, so I think it's just updating the Storybook story now.

PR #212, opened 2021-04-02
Convert videospotlightblock component to TypeScript

### LINKED ISSUE #187: Convert VideoSpotlightBlock component to TypeScript

ivanleeswe:
Convert Videospotlightblock component to TypeScript


src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.tsx
```diff
@@ -31,8 +40,9 @@ const VideoSpotlightBlock = ({
   button,
   imageURL,
   playButtonOnClick,
-  blackBackground,
-}) => {
+  blackBackground = false,
+  playIcon,
```
richardxia: I don't think this is the right approach, to add the `playIcon` to the props of the component like this. I think it's unlikely that we'd want to change this for each instance of the Video Spotlight Block, so it's better to just bake it into the component definition itself.

If you were hitting issues with TypeScript when importing SVG files, you might need a little snippet of code to add to our TypeScript module declaration file, which I had added in https://github.com/ShelterTechSF/sheltertech.org/pull/207 (which has not yet merged). I made that change in its own commit, so it should be easy to cherry-pick into this PR if you want.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.tsx
```diff
@@ -42,7 +52,7 @@ const VideoSpotlightBlock = ({
       <div className={s.bleedImageWrapper}>
         <div
           className={s.bleedImage}
-          style={{ "--background-image": `url(${imageURL})` }}
+          style={{ backgroundImage: `url(${imageURL})` }}
```
richardxia: What was the rationale behind this change? It's unlikely that this is the change you want to make, since we had specifically created and [use this CSS custom property](https://github.com/ShelterTechSF/sheltertech.org/blob/780ef84c827225633c8a1dc975faf4dc4f7eeae3/src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.module.css#L46) in the associated .scss file.

You can see that our SCSS uses it as part of a full [`background:` shorthand property](https://developer.mozilla.org/en-US/docs/Web/CSS/background), so if you stop setting it, then that CSS property may not function correctly.

richardxia (changes requested):
Thanks for working on this! I'm assuming that you ran into a little bit of trouble while working on the conversion, which is why you made a few some of the unusual changes you had here. Could you explain what the issues you were hitting?

Also, in the future, if you could describe these kinds of things in your PR description, that would be helpful for reviewers to understand the decisions you made and the context in which you made them, since otherwise, we have to take a guess based on the end result.

src/components/grid-aware/VideoSpotlightBlock/VideoSpotlightBlock.tsx
ivanleeswe: I added as React.CSSProperties instead.

richardxia:
Regarding the two errors you're getting in CI:

- The `playIcon` one will probably be resolved as soon as you merge the `main` branch into this one, since I added a TypeScript module declaration for all SVG files in https://github.com/ShelterTechSF/sheltertech.org/pull/207, which merged recently
- The CSSProperties one, we may just have to write a type assertion (the `as` syntax for forcing TypeScript to treat a value as a specific type), since the builtin TypeScript definitions don't support CSS custom properties (the special properties that we define that start with a `--`)

richardxia (approved):
Great, thanks for making the changes. LGTM!

PR #213, opened 2021-04-02
Convert Videoheader component to TypeScript

### LINKED ISSUE #186: Convert VideoHeader component to TypeScript

ivanleeswe:
Convert Videoheader component to TypeScript


src/components/grid-aware/VideoHeader/VideoHeader.tsx
```diff
@@ -0,0 +1,95 @@
+import * as React from "react";
+import Button, { ButtonProps } from "../../inline/Button";
+
+import s from "./VideoHeader.module.css";
+
```
richardxia, line 6: Super nit: It looks like you deleted the `/* Subcomponents */` comment that was in the original .js file. Was there a reason you did that, but not the `/* Main component */` lower in the file?

It's not super important in this case, but in general I avoid making any unnecessary changes when working on a task, since 1) it's hard for a reviewer to tell whether something was done intentionally or accidentally when it's not related to the task at hand, and 2) making small unrelated changes will increase the risk of conflicting with something else or running into an error for unnecessary reasons.

src/components/grid-aware/VideoHeader/VideoHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+  <div className={s.videoHeaderText}>
+    <div className={s.title}>{title}</div>
+    <div className={s.description}>{description}</div>
+  </div>
+);
+
+type VideoHeaderTextProps = {
+  title: string;
+  description: string;
+};
```
richardxia: Nit: Can we put the type declarations for the props of a component immediately above that component? This matches our style everywhere else, and it puts the definitions right next to the arguments of the components.

src/components/grid-aware/VideoHeader/VideoHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+/* Main component */
+
+const VideoHeader = ({
+  ctaButtons,
+  image,
+  playButtonOnClick,
+  title,
+  description,
+  playIcon,
```
richardxia: Same comment as https://github.com/ShelterTechSF/sheltertech.org/pull/212/files/8ebdab1de7be2b22fd4c1544cc1ca88ba3ca9c62#diff-b3b892c87e1a324c028b4a574ed9ea0f2d7974f9d5d25d3d905c5a91eeeae0c0 in a different PR, but let's not expose the `playIcon` as a prop, since it's not something that we really need to make configurable.

src/components/grid-aware/VideoHeader/VideoHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+type VideoHeaderProps = {
+  ctaButtons: ButtonProps[];
+  image: string;
+  playButtonOnClick: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  title: string;
+  description: string;
+  playIcon: ImageProps;
+};
```
richardxia: Same comment about putting the prop type declaration for a component immediately above the definition of that component.

richardxia (changes requested):

src/components/grid-aware/VideoHeader/VideoHeader.tsx
```diff
@@ -0,0 +1,90 @@
...
+  playButtonOnClick,
+  title,
+  description,
+}: VideoHeaderProps) => {
+  return (
+    <div className={s.bleedWrapper}>
+      <div
+        className={s.bleedBackground}
+        style={{
+          backgroundImage: `url(${image})`,
```
richardxia: Looks like you didn't fully revert this back to what it used to be, where it used a CSS custom property. Could you change this back and add the appropriate TypeScript type assertion?

richardxia (approved):
LGTM! Thanks!

PR #214, opened 2021-04-02
Convert OneParagraphBlock component to Typescript

### LINKED ISSUE #178: Convert OneParagraphBlock component to TypeScript

davidagustin:


src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.tsx
```diff
@@ -1,11 +1,22 @@
...
-import React from "react";
+import * as React from "react";
 
-import Button from "../../inline/Button";
+import Button, { ButtonProps } from "../../inline/Button";
 
 import s from "./OneParagraphBlock.module.css";
 
-const OneParagraphBlock = ({ title, descriptions, ctaButtons }) => {
+/* Subcomponents */
```
richardxia: This comment doesn't look right; there's only one component in this file, so if we want to have a comment to demarcate this part of the file, it really ought to say something more like "Main component". However, since there's nothing else in this file but the one and only component, maybe we should just remove this comment, which didn't exist in the original non-TypeScript version?

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.tsx
```diff
@@ -1,11 +1,22 @@
...
+
+type OneParagraphBlockProps = {
+  title: string;
+  descriptions: string;
```
richardxia: This doesn't match the original PropTypes type that was assigned to it, which was `PropTypes.node`. The PropTypes version is the correct type, because in the actual page where we instantiate this component, we're passing in a React fragment, not a string:

https://github.com/ShelterTechSF/sheltertech.org/blob/780ef84c827225633c8a1dc975faf4dc4f7eeae3/src/pages/volunteer/index.js#L150-L179

This isn't triggering a TypeScript error because we haven't converted the page components yet, but it will once we do that. Could you replace this with `React.ReactNode`, which is the equivalent to `PropTypes.node`?

richardxia (approved):
A couple of minor comments, but this otherwise LGTM! Feel free to merge once you address my comments.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.tsx
ivanleeswe: Removed.

src/components/grid-aware/OneParagraphBlock/OneParagraphBlock.tsx
ivanleeswe: Changed description prop type to React.ReactNode.

richardxia (approved):

PR #215, opened 2021-04-02
Convert ProgramsBlock component to TypeScript

### LINKED ISSUE #180: Convert ProgramsBlock component to TypeScript

ivanleeswe:
Convert ProgramsBlock component to TypeScript

### DIFF
--- a/src/components/grid-aware/ProgramsBlock/ProgramsBlock.tsx
+++ b/src/components/grid-aware/ProgramsBlock/ProgramsBlock.tsx
@@ -1,19 +1,23 @@
-import PropTypes from "prop-types";
-import React from "react";
+import * as React from "react";
 
 import ProgramCard from "../../block/ProgramCard";
 
 import s from "./ProgramsBlock.module.css";
 
-const ProgramsPropType = PropTypes.shape({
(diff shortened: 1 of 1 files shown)


src/components/grid-aware/ProgramsBlock/ProgramsBlock.tsx
```diff
@@ -1,19 +1,23 @@
...
 import s from "./ProgramsBlock.module.css";
 
-const ProgramsPropType = PropTypes.shape({
-  theme: PropTypes.string.isRequired,
-  image: PropTypes.string.isRequired,
-  imageAlt: PropTypes.string.isRequired,
-  title: PropTypes.string.isRequired,
-  description: PropTypes.string.isRequired,
-});
+type ProgramsPropTypeProps = {
```
richardxia: This name is a little bit awkward since it's redundant. How about we just name this `Program` or `ProgramType`? It's not actually a "prop", since it's not directly the prop of a React component (it's really just a single item of an array, which is the actual prop), so we don't need to try to include the word "Prop" in the name of this type alias.

richardxia (approved):
LGTM, once you make a minor name change!

PR #217, opened 2021-04-02
Convert BlockQuoteBlock component to TypeScript

### LINKED ISSUE #170: Convert BlockQuoteBlock component to TypeScript

ivanleeswe:
Convert BlockQuoteBlock component to TypeScript

### DIFF
--- a/src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.tsx
+++ b/src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.tsx
@@ -28,8 +28,8 @@ const BlockQuoteBlock = ({ quotes }: BlockQuoteBlockProps) => (
         isIntrinsicHeight
         isPlaying
         interval={10000}
-        naturalSlideHeight={0}
-        naturalSlideWidth={0}
+        naturalSlideHeight={100}
+        naturalSlideWidth={100}
       >
         <Slider>
           {quotes.map(({ quote, attribution }, index) => (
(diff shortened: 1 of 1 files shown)


src/components/grid-aware/BlockQuoteBlock/BlockQuoteBlock.tsx
```diff
@@ -1,30 +1,35 @@
...
   <div className={s.bleedWrapper}>
     <div className={s.carouselWrapper}>
       <CarouselProvider
         totalSlides={quotes.length}
         isIntrinsicHeight
         isPlaying
-        interval="10000"
+        interval={10000}
+        naturalSlideHeight={0}
+        naturalSlideWidth={0}
```
ivanleeswe: Previously, this component did not require these properties to be passed down as props, but in this TS integration, the linter forced these values to be required. @richardxia

richardxia: Ah, interesting, well it's nice that [`pure-react-carousel`](https://github.com/express-labs/pure-react-carousel) comes with its own official TypeScript types, and that it caught this mistake we made with not passing in required properties. However, I don't think these values of 0 are the correct ones to pass in.

According to their README:

> The only required properties are the orientation, naturalSlideWidth, and naturalSlideHeight. The naturalSlideWidth and naturalSlideHeight are used to create an aspect ratio for each slide. Since the carousel is responsive by default, it will stretch to fill in the width of it's parent container.

That sounds like we should be passing in real values, but the values are not the absolute height and width, but rather relative numbers that allow us to construct the aspect ratio it should be scaled up and down with. Could you figure out the actual aspect ratio we currently have on the site and pick some numbers here that roughly conform to that aspect ratio?

ivanleeswe: Changed width to 45 and height to 10.

richardxia (approved):
Awesome, thank you, LGTM!

PR #222, opened 2021-04-05
176 convert modal to typescript

### LINKED ISSUE #176: Convert Modal component to TypeScript

jessica-px:
Closes #176 

Converts the `<Modal>` component to typescript. Only three comments about this:

1. `ariaHideApp` is an optional prop with a default of `true`, but TypeScript doesn't directly have a way of handling optional props. I investigated making it required, but we only ever need to set this to `false` for Storybook. It should generally always be `true` for accessibility reasons, so manually passing it around everywhere would feel a bit tedious. 
So I instead slapped together this "default" logic: `const ariaHide = ariaHideApp === undefined ? true : ariaHideApp;`, but if someone has oppositions to this kind of ternary operator, I'm open to alternatives.

2. TypeScript didn't want me to do a default import of react-modal, so I tried importing it like so: `import * as ReactModal from "react-modal";`. But our storybook/jest configuration absolutely hated this and kept failing tests for no good reason. I could not figure out how to placate it, so I reconfigured TypeScript instead -- to get TypeScript to agree with the default import, I had to set `esModuleInterop` to `true` in our `tsconfig.json`.

3. Discovered a weird bug 🐛 I tested the modals in storybook and locally to ensure the behavior didn't change. We have 4 modals in total, and 3 of them worked as expected. But the modal toggled in `<VideoSpotlightBlock>` doesn't open in Firefox (my default browser). It does, however, open in Chrome! I checked the live site and, surprisingly, it's the same situation there. Anyway, because it's in production, it should be unrelated to this PR. I can open a separate issue for investigating that.

### DIFF
--- a/src/components/grid-aware/Modal/Modal.tsx
+++ b/src/components/grid-aware/Modal/Modal.tsx
@@ -7,7 +7,7 @@ import closeIcon from "./close-icon.svg";
 type ModalProps = {
   children: React.ReactNode;
   isOpen: boolean;
-  setIsOpen: (x: boolean) => null;
+  setIsOpen: (x: boolean) => void;
   ariaHideApp?: boolean; // we only ever set this to false for storybook
   contentLabel: string;
   noBezel?: boolean; // If true, then eliminate the padding and close button
--- a/src/components/grid-aware/Modal/Modal.tsx
+++ b/src/components/grid-aware/Modal/Modal.tsx
@@ -1,9 +1,17 @@
-import PropTypes from "prop-types";
-import React from "react";
+import * as React from "react";
 import ReactModal from "react-modal";
 
 import s from "./Modal.module.css";
 import closeIcon from "./close-icon.svg";
+
+type ModalProps = {
+  children: React.ReactNode;
+  isOpen: boolean;
+  setIsOpen: (x: boolean) => null;
+  ariaHideApp?: boolean; // we only ever set this to false for storybook
+  contentLabel: string;
+  noBezel?: boolean; // If true, then eliminate the padding and close button
+};
 
 const Modal = ({
   children,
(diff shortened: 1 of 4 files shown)


src/components/grid-aware/Modal/Modal.tsx
```diff
@@ -1,29 +1,38 @@
...
+  ariaHideApp?: boolean; // we only ever set this to false for storybook
+  contentLabel: string;
+  noBezel?: boolean; // If true, then eliminate the padding and close button
+};
+
 const Modal = ({
   children,
   isOpen,
   setIsOpen,
   ariaHideApp,
```
richardxia: Regarding the issue with the default true prop, shouldn't you be able to assign an ES6 default argument value of `true`, like this?

```suggestion
  ariaHideApp = true,
```

This is basically what I did with the `isButton` prop in the Navigation component, which is optional: https://github.com/ShelterTechSF/sheltertech.org/blob/3cf755d7e289b5c9734e6ac3551eb3958d6f1e65/src/components/grid-aware/Navigation/BurgerMenu.tsx#L13

src/components/grid-aware/Modal/Modal.tsx
```diff
@@ -1,29 +1,38 @@
...
+import * as React from "react";
 import ReactModal from "react-modal";
 
 import s from "./Modal.module.css";
 import closeIcon from "./close-icon.svg";
 
+type ModalProps = {
+  children: React.ReactNode;
+  isOpen: boolean;
+  setIsOpen: (x: boolean) => null;
```
richardxia: You should actually specify a return type of [`void`](https://www.typescriptlang.org/docs/handbook/2/functions.html#void) here. The difference between `void` and `null` is that `null` is a real JavaScript value that can be returned, while `void` is a TypeScript only type indicates that a function has no real return value. With a function that returns `null`, it is legal to assign that return value to a variable that accepts `null` as a potential value, whereas with a function that has a `void` return type, it is illegal to try to do anything with the return value.

richardxia (approved):
Left a couple of small comments, but this mostly LGTM!

I'm a bit sad about the `esModuleInterop` thing, but I agree that it's probably the right thing to do for now. Maybe in the future, when enough libraries make use of ES-style modules, we'll be able to turn that off again.

And good catch on the Firefox bug for the VideoSpotlightComponent! I agree that it's probably something that we've shipped for a long time, probably due to us not testing enough on Firefox.

src/components/grid-aware/Modal/Modal.tsx
jessica-px: Forgot about that! Very exciting feature 🎉

PR #236, opened 2021-04-16
Initial Prismic integration

### LINKED ISSUE #225: Set up initial Prismic–Gatsby integration
This covers getting Gatsby set up with our Prismic account and doing some basic integration on some simple component just to prove that the integration is functional.

richardxia:
Resolves #225. Note that this PR targets a new branch named `prismic`, not `main`. More on that later in the PR description.

This PR includes some small changes to have this Gatsby application talk to [Prismic](https://prismic.io/), the content management system (CMS) we're planning to integrate with. I've installed [gatsby-source-prismic](https://github.com/angeloashmore/gatsby-source-prismic), which is the Gatsby plugin that [is officially supported by Prismic](https://prismic.io/docs/technologies/plugin-configuration-gatsby), and connected it to our Prismic account.

As the most basic demonstration of an integration with Prismic, we're now fetching the `address` field from the `Footer` object from Prismic, using that data to populate the address at the very bottom right of the footer. In even doing this, we've already encountered several issues that we've broken up into GitHub tickets to tackle separately.

@jessica-px and I pair programmed on most of this, and although it's far from complete, we think it's a useful, PR-able first step. We're planning to use a new branch named `prismic` as a longer-running integration branch so that we keep the production site isolated from these changes. This will allow us to explore and experiment more without impacting the production site and give us the freedom to change our minds any part of the setup. This also gives us the freedom to completely wipe the data and data schemas defined in Prismic, which is not something we'd be able to easily do after launching this in production.

Anyone should be able to check out this branch, `npm install` the new dependencies, and run the Gatsby development server to see data being fetched from Prismic. You don't even need the Prismic account credentials because the API endpoint is public. You can tell if it's working if you see "268 Bush Street2" as the address in the footer, since I had intentionally altered the address to test the integration.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,6 +48,8 @@ module.exports = {
       // Gatsby configuration files
       files: ["gatsby-config.js"],
       rules: {
+        // global-require is fine because this is a Node.js script.
+        "global-require": "off",
         "import/no-extraneous-dependencies": [
           "error",
           { devDependencies: true },
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -27,8 +27,35 @@ module.exports = {
         postCssPlugins: [postcssCustomMedia],
       },
     },
-    // this (optional) plugin enables Progressive Web App + Offline functionality
-    // To learn more, visit: https://gatsby.dev/offline
-    // 'gatsby-plugin-offline',
+    {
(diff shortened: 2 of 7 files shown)


jessica-px (approved):
Checked it out and confirmed the address was correctly reading from Prismic -- LGTM 👍

PR #243, opened 2021-04-22
Generate TypeScript types from GraphQL schemas

### LINKED ISSUE #229: See if TypeScript types can be generated from Prismic/Gatsby/GraphQL schemas

richardxia:
Resolves #229. Once again, this targets the `prismic` branch instead of `main`.

## Overview

This adds the [gatsby-plugin-typegen](https://github.com/cometkim/gatsby-plugin-typegen) plugin to our project, which will automatically generate TypeScript type definitions for the GraphQL schemas.

## Research

I did some research online, and it looked like the two main plugins that could be used for this were https://github.com/cometkim/gatsby-plugin-typegen and https://github.com/d4rekanguok/gatsby-typescript. However, the top of the README of the latter specifically recommends that if you're only interested in TypeScript GraphQL types, you should actually go use the other library instead, since `gatsby-typescript` is less maintained when it comes to the GraphQL type generation.

## How this works

The actual type generation just happens automatically when you run the Gatsby build, and it places the type definitions int `src/__generated__/gatsby-types.ts`. All of the types will be located in a TypeScript namepace named `GatsbyTypes`, which is globally defined, so you do not need to import before using it.

The type names appear to be derived from the name we assign the GraphQL query within the `graphql`` ` construct. For example, if we have:

```js
graphql`
  query Foo {
    data
  }
`;
```

then the TypeScript type will be named `GatsbyTypes.FooQuery`. This means that we will need to adopt a consistent naming convention for GraphQL queries in to avoid name conflicts and because we won't be able to change the name of the query without having to change all references to the TypeScript type. I think for all the "pages" (i.e. files under `src/pages`), we should name the query `[Foo]Page` so that the type name becomes `[Foo]PageQuery` (replace `[Foo]` with the name of the page). For non-page queries, we'll probably have to evaluate it on a case by case basis.

With the types automatically available, then the only thing to do to make use of them is to assign the appropriate type to the props of each page's React component. For Gatsby pages, you can use the `PageProps` generic type, which takes a generic type for the data sources within it, so the home page's `props` parameter should have the type `PageProps<GatsbyTypes.HomePageQuery>`. You can take a look at the actual diff of my PR to see concrete examples.

## What I changed in this PR

Because it looks like page queries and (non-page) static queries [have different behaviors](https://www.gatsbyjs.com/docs/how-to/querying-data/static-query/#how-staticquery-differs-from-page-query) in Gatsby, I thought it would be good to have a demonstration of both types of queries. Within Prismic itself, I added a new Home Page custom type with a single field for the video header title. We can change the custom type later, when we think more about the design of the schemas, but I just wanted available to query now.

Within this PR, I added the schema file for the home page type, and then I updated the home page to start querying the Video Header title from Prismic. I also moved the footer query into the Layout component, which holds both the header and the footer, and this query is done with the `useStaticQuery()` hook from Prismic.

Most of the rest of the changes are just plumbing work, which I think are either self-explanatory, or that you can see from the individual commit messages I gave my commits.

## About optional props
One thing that I quickly noticed is that although we now have static type information from the API queries, it looks like all of the fields are marked as optional, including nested fields. The README of [gatsby-typescript](https://github.com/d4rekanguok/gatsby-typescript) (which, BTW, is the plugin that we're _not_ using) has this in their FAQ:

> **What is up with all the Maybe<T>?**
> It's due to Gatsby internal. There's an effort to [make typing more strict here](https://github.com/gatsbyjs/gatsby/issues/20069).

This basically means that we have to do a lot of [option chaining in TypeScript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining), which, I'm glad is at least available to us (writing this code without the `?.` operator would be a nightmare). This ends up looking like `data.prismicFooter?.data?.address?.text`.

You might be wondering whether automatically generating TypeScript definitions for the GraphQL queries is even worth it if there are optional values everywhere. My answer is that I still think they're useful, since they catch issues with trying to access the wrong attribute on an object, and it still provides useful information to editors such as autocomplete options.

Even if Gatsby improves their types to remove unnecessary null types, we still actually have to handle the case where the API query to Prismic returns null (empty?) values due to not all fields being filled out for an object in Prismic. Prismic actually has a blog post titled "[Unpopular opinion: why required fields lead to terrible UX](https://prismic.io/blog/required-fields)" explaining why they made this decision.

I'm thinking that we should gradually update our components to gracefully handle missing fields by conditionally rendering pieces only if their values are defined. I've already gone through and made the VideoHeader component handle the case where the title is missing, but we may need to do a lot more of this as we cover more of the content.

On the plus side, TypeScript will make it easy for us to make sure we've handled optional values, since it'll yell at us if we try to use a potentially null/undefined value before checking that it is not null/undefined.

## Other remarks

Overall, I am pretty happy about this integration, since it just kind of worked out of the box. There's less documentation than I would have liked in `gatsby-plugin-typegen`, but if you're not afraid to dive into the code or inspect the generated files, it all seems to work together.

I have noticed the whole build process be a little bit flaky, especially when you add a new custom type. I'm not sure if there's much we can really do here, since we've definitely glued together a done of heavy libraries in this app, so we may just need to get used to restarting Gatsby often. That said, once the schemas are already added, it seems to be pretty accurate and reliable.

## Request for comments

I'm interested in getting feedback about:

- Whether this looks like the right thing to do, despite some of the drawbacks
- Whether you have any thoughts about the optional props discussion
- Actual code review of the changes I've made here

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,11 +21,14 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      # This has to come before linting because we need to run the Gatsby build
+      # at least once in order to generate the GraphQL TypeScript definitions.
+      # The TypeScript type check lint check will otherwise fail.
+      - name: Check build
+        run: npm run build
+
       - name: Run lint checks
         run: npm run lint
 
       - name: Run tests
         run: npm run test
-
-      - name: Check build
-        run: npm run build
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ jspm_packages/
 # gatsby files
 .cache/
 public
+src/__generated__/
 
 # Storybook files
 /storybook-static
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -49,6 +49,7 @@ module.exports = {
         schemas: {
           // Your custom types mapped to schemas
           footer: require("./src/schemas/footer.json"),
+          home_page: require("./src/schemas/home_page.json"),
         },
 
         // Add the Prismic Toolbar script to the site. Defaults to false.
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -57,5 +57,6 @@ module.exports = {
         prismicToolbar: true,
       },
     },
+    "gatsby-plugin-typegen",
   ],
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1478,6 +1478,23 @@
         "minimist": "^1.2.0"
       }
     },
+    "@cometjs/core": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@cometjs/core/-/core-0.2.0.tgz",
+      "integrity": "sha512-H5G7/xYoHxmR4L4Pd77ScRCok2ttz7sG0BZ2B+pzsPammoIE63UiN2bCHjDOkXqbQplWU8xCiJBbHs3Cut2k/g==",
+      "dev": true,
+      "requires": {
+        "core-js": "^3.6.5"
+      },
+      "dependencies": {
+        "core-js": {
+          "version": "3.10.1",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-3.10.1.tgz",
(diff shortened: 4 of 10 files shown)


jessica-px (approved):
Looks good to me -- reviewing this hasn't changed my opinions since we last talked 👍 I do think I'll examine all of this with more scrutiny once we're finalizing the process and fully documenting it, but I don't have any concerns or anything right now.

PR #244, opened 2021-04-23
Convert infoblock and subscriptionblock to typescript

### LINKED ISSUE #240: Convert Footer SubscriptionBlock and Footer InfoBlock to TypeScript
We had missed these when we converted the other parts of the Footer component to TypeScript.

ivanleeswe:
Fixes #240


richardxia (approved):
LGTM, although this only converts the InfoBlock to TypeScript, so we'll also want to convert the SubscriptionBlock before considering this ticket "closed".

ivanleeswe:
@richardxia Please review before I merge. I'm not sure whether making onChange optional is the right choice, but there was an error stating as follow:

`Property 'onChange' is missing in type '{ name: string; placeholderText: string; type: string; }' but required in type 'InputTextProps'.ts(2741) InputText.tsx(29, 3): 'onChange' is declared here.`

richardxia (approved):
> @richardxia Please review before I merge. I'm not sure whether making onChange optional is the right choice, but there was an error stating as follow:
> 
> `Property 'onChange' is missing in type '{ name: string; placeholderText: string; type: string; }' but required in type 'InputTextProps'.ts(2741) InputText.tsx(29, 3): 'onChange' is declared here.`

Ah yeah, in this case, we do want to make this change, since we're using that component as an [uncontrolled form component](https://reactjs.org/docs/uncontrolled-components.html), so we don't want to pass in an onChange handler.

These changes LGTM!

PR #228, opened 2021-04-11
Convert Donate Page to TypeScript

candywang:
Closes #221.

PR #245, opened 2021-04-23
Change js files to tsx

### LINKED ISSUE #242: Convert all non-legacy index.js files to TypeScript
To be clear: non-legacy means everything in `src/components` as well as `src/pages/donate/index.js` and `src/pages/volunteer/index.js`.

ivanleeswe:
Fix #242


src/pages/donate/index.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
```
richardxia, line 1: Note that this file is already in the process of being converted in @candywang's PR, https://github.com/ShelterTechSF/sheltertech.org/pull/228.

richardxia (commented):
Hmm, I noticed you changed all the file extensions to .tsx instead of .ts. My inclination would be to only use .tsx for files containing JSX syntax (the HTML-y looking stuff embedded in JavaScript) and .ts for file that don't require JSX.

Note that this is not just an aesthetic preference. TypeScript actually has to use a [slightly different syntax and parser for TSX files](https://www.typescriptlang.org/docs/handbook/jsx.html#the-as-operator) because there is a conflict between the JSX language syntax and the TypeScript language in how angle brackets (`<>`) are handled. Although the TypeScript team decided to introduce a new syntax for type assertions in order to handle JSX (the `as` operator), there is still at least one case where a TSX file is just [incapable of expressing some things](https://stackoverflow.com/a/54614279) that would normally possible in plain TypeScript.

In terms of aesthetics, I also prefer .ts for non-JSX files since it's an indication to the reader of what kind of code is present in the file. Seeing .tsx indicates to the reader that there is definitely React code in there, whereas seeing .ts indicates that there is no React code, and that it is likely to be purely JavaScript logic.

That said, I could see that this distinction could be difficult to maintain and explain to new people on the project, and maybe we should just go full .tsx to keep things simpler. Even [Microsoft went with that approach for their TypeScript-React-Starter](https://stackoverflow.com/a/54614279) repo because "It was just simpler to stick with TSX files across the board instead of involving users with this explanation."

@jessica-px, do you have any thoughts on this?

jessica-px:
> @jessica-px, do you have any thoughts on this?

@richardxia My default inclination is also to use `.ts` for plain TypeScript files and `.tsx` for files with JSX, just like how the common convention is to distinguish `.js` from `.jsx` files. I agree that it's helpful to be able to, at a glance, discern what contains components and what's just logic. And since it's such a common convention, I don't think it's a big ask to just make it a standard part of our code style.

But if we do find ourselves having to give this explanation over and over, then I suppose a decision like Microsoft's makes sense. I don't expect we'd lose anything important by using `.tsx` across the board. Although it might be confusing to those of us used to the dual extension convention, it's probably less effort for us to learn to use one thing than it is for others to learn to use two things.

So I guess I'd base the decision around who we expect to be using this codebase more -- people already familiar with the convention, or people still new to it? And, for the people who are new to it, do we want to help them learn the common conventions? Or do we just want to make this particular codebase easy to use?

richardxia:
> And since it's such a common convention, I don't think it's a big ask to just make it a standard part of our code style.

That sounds reasonable to me. I think we may also be able to configure ESLint to enforce this using the [`react/jsx-filename-extension`](https://github.com/yannickcr/eslint-plugin-react/blob/0d999ef2095dc0f5d0944a41f0a23c22cbe84562/docs/rules/jsx-filename-extension.md), and the only trouble we may run into are the legacy pages, which have followed Gatsby's convention of naming all JSX and JS files with the `.js` extension, but we may just be able to exclude those pages individually since we're almost done with the TypeScript conversion.

@ivanleeswe, do you think you could go back and change the names of all the `.tsx` files that don't actually contain any JSX code to just be `.ts` files? This should be almost all the `index` files, although a few of them may actually contain JSX code.

richardxia:
And I just opened up a ticket for updating the ESLint configuration to enforce the style: https://github.com/ShelterTechSF/sheltertech.org/issues/246

ivanleeswe:
Thanks for helping me understand the difference between .ts and .tsx files. I will change them now.

richardxia (approved):
LGTM, although you have a merge conflict with `main`, so you'll need to resolve that before this can be merged.

ivanleeswe:
@richardxia Hi Richard, I made a lot of changes to fix this issue. Will you please look through this one more time?

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.tsx
```diff
@@ -7,7 +7,7 @@ import s from "./TwoParagraphBlock.module.css";
 /* Main component */
 type TwoParagraphBlockProps = {
   title: string;
-  paragraph1: React.ReactNode;
+  paragraph1: string;
```
richardxia: Is there a reason you needed to change this? `ReactNode` is a type alias that includes `string` as one of its valid subtypes, so you should be able to pass a string into a ReactNode prop. I think we'd prefer the `ReactNode` here for symmetry with `paragraph2`, where we're currently passed in something more general than a string, JSX code.

tsconfig.json
```diff
@@ -17,6 +17,6 @@
     // Advanced
     "skipLibCheck": true // Don't type check .d.ts files, which mostly come from third party libraries.
   },
-  "include": ["**/*", ".storybook/*", ".eslintrc.js"],
+  "include": ["**/*", ".storybook/*", ".eslintrc.js", "src/components/grid-aware/Footer/SubscriptionBlock/.js"],
```
richardxia: What is this change for? `**/*` should already include basically everything, and because `src/components/grid-aware/Footer/SubscriptionBlock/.js` is just a simple string, not a glob pattern, this won't match any actual files.

richardxia (commented):
> @richardxia Hi Richard, I made a lot of changes to fix this issue. Will you please look through this one more time?

Yep, sorry for the delay! Just looked through it again, and I left a couple questions about some of the changes you made.

PR #248, opened 2021-04-25
Convert data/partnersAndSponsorsLogo.js to TypeScript

### LINKED ISSUE #241: Convert data/partnersAndSponsorsLogos.js to TypeScript

candywang:
Closes #241.


richardxia:
Ah, I was actually hoping that we could explicitly define a type for the items in the array. Even though TypeScript seems to be able to infer it here, I think it'd be better to have an explicit type since 1) it helps readability, and 2) it makes us less reliant on TypeScript's type inference, which might not do the expected thing if we accidentally mistyped one of the keys in the array.

richardxia (approved):
Awesome, LGTM!

PR #255, opened 2021-06-07
Create stacking context to allow play button to be clickable on Firefox

### LINKED ISSUE #223: VideoSpotlightBlock modal doesn't open in Firefox
The modal toggled by the `<VideoSpotlightBlock>` component doesn't seem to open in Firefox. You can reproduce this by going to the live site in Firefox and trying to click the play button on this video:

Screen Shot 2021-04-04 at 6 30 49 PM

In Chrome, it will open a modal with an embedded YouTube video, which is the expected behavior. In Firefox, nothing happens. Potentially the issue might lie in the CSS somewhere, but I didn't investigate that closely.

ivanleeswe:
Fix #223


richardxia (approved):
LGTM! I'm glad that we were able to solve this while constraining the z-index to a stacking context.

ivanleeswe:
> LGTM! I'm glad that we were able to solve this while constraining the z-index to a stacking context.

Thanks for pairing up with me. Learned a lot about positioning, z-index, and stacking context!

PR #257, opened 2021-06-16
Configure ESLint rule react/jsx-filename-extension to restrict .tsx extension to only files containing JSX

### LINKED ISSUE #246: Configure ESLint rule react/jsx-filename-extension to restrict .tsx extension to only files containing JSX
We may need to exclude the legacy pages, since the Gatsby default style is to name all JS and JSX files with a .js extension.

candywang:
Closes #246.


.eslintrc.js
```diff
@@ -20,10 +20,7 @@ module.exports = {
   plugins: ["@typescript-eslint", "react", "react-hooks"],
   rules: {
     "import/order": ["error", { alphabetize: { order: "asc" } }],
-    "react/jsx-filename-extension": [
-      "error",
-      { extensions: [".js", ".jsx", ".tsx"] },
-    ],
+    "react/jsx-filename-extension": [1, { extensions: [".tsx"] }],
```
richardxia: Setting a value of 1 only turns violations into warnings, not errors. This is why I prefer to use the string versions (e.g. "off", "warn", and "error") of the settings instead of the numbers (e.g. 0, 1, and 2). See https://eslint.org/docs/user-guide/configuring/rules#configuring-rules for documentation on this.

Also, it looks like you are triggering warnings in our codebase, which you can see with the GitHub diff view, since they show the warnings and errors that lint detects. It looks like you need to allow for a `.jsx` extension for the Storybook stories, which are written in JavaScript, not TypeScript.

candywang: @richardxia Do we plan to convert Storybook stories to have a `.jsx` extension?

richardxia: Just writing down what we said in the meeting: Yes, let's rename all the Storybook stories to `.jsx`.

richardxia:
The ESLint configuration looks good to me now. All that's left is to actually rename the .stories.js files to stories.jsx, which should be only thing you need to do to get this to pass CI!

richardxia (approved):
LGTM!

Issue #260, opened 2021-06-20
Upgrade Gatsby from v2 to v3


richardxia:

richardxia:
cc @jessica-px

It looks like some of the issues we have with upgrading are due to Gatsby moving to Webpack 5, whereas Storybook still relies on Webpack 4. Coincidentally, Storybook 6.3.0 will support Webpack 5 and is [slated to come out tomorrow](https://github.com/storybookjs/storybook/issues/14397). I might try upgrading Storybook by itself, since I think it's more flexible about its dependencies than Gatsby is.

PR #261, opened 2021-06-22
Upgrade Storybook to 6.3.0

richardxia:
Relates to https://github.com/ShelterTechSF/sheltertech.org/issues/260, but does not resolve it.

## Background
It's kind of a nightmare trying to upgrade to Gatsby v3 because of how many things break, so I'd like to take an easier step by upgrading smaller sets of packages at a time. The Gatsby and Storybook upgrades are unfortunately pretty linked because in our current setup, a lot of the resolved versions of common dependencies end up getting intertwined. NPM allows multiple versions of the same package to be used within the same project, but there are some complicated heuristics for determining whether versions are "hoisted" and which version becomes the project's "default" (i.e. located in the top-level node_modules/ directory as opposed to being nested under some other package).

One package in particular that is causing a lot of issues is Webpack, since Webpack 5 made some significant, backwards incompatible changes, and it looks like it's somewhat fragmented the frontend ecosystem, especially when packages completely drop support for Webpack 4.

Gatsby v3 moves from Webpack 4 to Webpack 5, and this appears to be a major source of issues in our project. Through a lot of trial and error as well as Googling error messages, I've determined and confirmed that most of these issues seem to be caused by both Webpack 4 and Webpack 5 operating on the same internal data structures, where Gatsby is pulling in Webpack 5 but Storybook is relying on Webpack 4, and in certain places the dynamic importing mechanisms end up trying to pass data structures from Webpack 5 to Webpack 4.

Storybook 6.2.0 added experimental support for Webpack 5, and [6.3.0 appears to add full support](https://github.com/storybookjs/storybook/blob/7d81f46f947769ae3db920bc553c2f9b13f3f9b8/CHANGELOG.md#630-rc0-june-9-2021).

## Changes made in this PR
This PR _only_ bumps Storybook from 6.0 to 6.3 without enabling Webpack 5 just yet. This at least allows me to address some of the warnings that come from upgrading Storybook, and it also confirms that upgrading Storybook by itself is safe.

~~Note that Storybook 6.3 has not been officially released yet, so this actually just upgrades to the 11th release candidate. I would normally avoid pointing to release candidate versions of libraries, but I think it's justified here because:~~

1. ~~The Gatsby v3 upgrade is so challenging that I want to do anything to make it easier, and~~
2. ~~Storybook is only a developer tool, not a critical feature of our site. In the worst case scenario, we could completely remove Storybook and we could still do development on the site.~~

Edit: Storybook 6.3 has been officially released!

## How this was tested
Since I want to be very thorough about this, I've tested all four of the following:

- Gatsby development server (`npm run start`)
- Gatsby production build, including viewing the built files in a browser (`npm run build`)
- Storybook development server (`npm run storybook`)
- Storybook static build (`npm run build-storybook`)

## Next steps
I've made a lot of progress upgrading Gatsby in a local branch, but I'm still hitting issues. I'm trying to pull out the smaller, safer changes into separate PRs to help me rule out the causes of the remaining issues I'm seeing in my local branch.

After this PR is merged, I'll attempt to rebase my local branch on top of this and see if I can work through the last remaining issues by configuring Storybook to use Webpack 5 as well. If that doesn't go well, I might break out more PRs that upgrade individual libraries so that I can minimize the number of upgraded libraries in my Gatsby upgrade branch.


richardxia:
Just updated this PR to point to the released version of Storybook 6.3.0, not the RC version. Ran through the same set of manual tests I described above, and it looks like things are still working fine.

jessica-px (approved):
Ran through the tests you described -- works fine on my end 👍

PR #262, opened 2021-06-24
Upgrade Gatsby to v3

### LINKED ISSUE #260: Upgrade Gatsby from v2 to v3

richardxia:
Fixes #260.

Note: This currently depends on https://github.com/ShelterTechSF/sheltertech.org/pull/261 being merged first.

This was a lot of trial, error, and Googling around, but I finally managed to get the codebase upgraded to Gatsby v3. The magic to getting most of this working was to completely delete `package-lock.json` _and_ `node_modules/` and then regenerate `package-lock.json` from scratch. This appears to fix some issues with transitive dependencies getting pinned to older versions of other transitive dependencies, and completely blowing away `package-lock.json` essentially upgrades every package, both direct and transitive, to the latest versions.

The other major, difficult-to-make change was upgrading Storybook to use webpack 5, which I did in https://github.com/ShelterTechSF/sheltertech.org/commit/21b22ebd556ee6ae00b9871089c1152d869540f1. It took a _lot_ of futzing around with the webpack configuration in `.storybook/main.js` to get things to work, since initially, it just wasn't understanding our CSS modules and rendering every component with any styling whatsoever.

Other, smaller changes included changing all of our CSS modules imports from `import s from "./Foo.module.css"` to `import * as s from "./Foo.module.css"` (thanks @jessica-px, for pointing me to your branch where you had already done this, since I just cherry-picked it in!) and minor changes to code to either pass the newer ESLint version or to waive more rules.

Note: If you have issues with getting the app running locally after these changes, I would recommend nuking your `.cache/` and `node_modules/` directories and reinstalling all dependencies from scratch.


jessica-px (approved):
I deleted the files/directories you named and it builds and runs just fine! 🎉 In fact, it looks even better than before, because the YouTube thumbnails weren't actually working for me locally before...but they work fine now! Thanks for figuring all of this out, Richard!!

richardxia:
Thanks for reviewing! I'm glad that this stuff is now in so that we can start revisiting the newer versions of the Gatsby-Prismic plugins.

PR #263, opened 2021-07-01
Upgrade gatsby-source-prismic to v4

### LINKED ISSUE #256: Upgrade gatsby-source-prismic
My current progress trying to upgrade gatsby following the steps in these guides:
- https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3/
- https://github.com/angeloashmore/gatsby-source-prismic/blob/alpha/packages/gatsby-source-prismic/docs/migrating-from-v3-to-v4.md

I'm currently getting this error when I try to run `npm run start`:

```
ERROR #98124  WEBPACK

Generating development JavaScript bundle failed

undefined

If you're trying to use a package make sure that 'undefined' is installed. If you're trying to use a
 local file make sure that the path is correct.

failed Building development bundle - 25.595s
ERROR in commons
Module not found: TypeError: Cannot read property 'indexOf' of undefined
```

It's hard for me to tell exactly what's causing the error. Because the old version of `gatsby-source-prismic` won't work with the new version of `gatsby`, I had to (try to) upgrade them at the same time. But this makes debugging harder. My next step was going to be the branch off of `main` and see if I could get gatsby upgraded on its own, without prismic.

jessica-px:
## Description

Upgrades us to gatsby-source-prismic v4! Since @richardxia already did the hard part upgrading Gatsby itself, and we weren't really doing much with Prismic yet, this actually took a pretty minimal amount of Googling to figure out. One thing to note is that we now require a schema for every Prismic type, even if we're not using them (and also, we'll need to keep the definitions even if we later delete the types!). For now I passed empty objects to the schemas we're not yet using.

I also deleted some things that it yelled at me about, including `fetchLinks` (which v4 has a new replacement for) and `prismicToolbar`, which has been combined into the Prismic preview plugin that I'll be implementing in a different PR.

## How to Test

If you run this locally, the site should look exactly as you expect. If the title text in the frontpage video ("Less than half of nearly 28,000 people...") shows correctly, then we know Prismic is working!

Closes #256


richardxia (approved):
> One thing to note is that we now require a schema for every Prismic type, even if we're not using them (and also, we'll need to keep the definitions even if we later delete the types!).

Hmm, that's pretty weird, and it's probably worth filing a ticket against gatsby-source-prismic, since that will make it difficult to gradually roll out changes to a live Prismic instance. But what you have with the empty objects LGTM!

jessica-px:
> > One thing to note is that we now require a schema for every Prismic type, even if we're not using them (and also, we'll need to keep the definitions even if we later delete the types!).
> 
> Hmm, that's pretty weird, and it's probably worth filing a ticket against gatsby-source-prismic, since that will make it difficult to gradually roll out changes to a live Prismic instance. But what you have with the empty objects LGTM!

It seems like intended behavior, from how they describe it [in the upgrade guide here](https://github.com/angeloashmore/gatsby-source-prismic/blob/alpha/packages/gatsby-source-prismic/docs/migrating-from-v3-to-v4.md#provide-all-custom-type-schemas) and in [this discussion response here](https://github.com/angeloashmore/gatsby-source-prismic/issues/377#issuecomment-850644839) (the 4th bullet point).

richardxia:
> It seems like intended behavior, from how they describe it [in the upgrade guide here](https://github.com/angeloashmore/gatsby-source-prismic/blob/alpha/packages/gatsby-source-prismic/docs/migrating-from-v3-to-v4.md#provide-all-custom-type-schemas) and in [this discussion response here](https://github.com/angeloashmore/gatsby-source-prismic/issues/377#issuecomment-850644839) (the 4th bullet point).

Ah, thanks for finding that issue. Seems like they might move to using the Prismic Custom Types API, which would allow us to always dynamically fetch the custom type schemas from Prismic itself, which would be nice. I guess for now, we'll have to carefully coordinate with each other if we want to add custom types, since doing so will immediately break all other development branches.

PR #264, opened 2021-07-05
Build About Us page

candywang:
In this PR,

The **About Us** page has been implemented with the basic layout and components. However, overall CSS and the `ImageHeader` component is a WIP.


src/components/grid-aware/Footer/Footer.stories.jsx
```diff
@@ -49,6 +49,7 @@ DefaultFooter.args = {
   formAntiBotInputName: "b_c47829732a0bea5c8e8a94604_08f60e42ef",
   formDescription: "Subscribe to get updates",
   footerNavigation: [
+    { text: "About", internalLink: "/new/about" },
```
richardxia: No need to update this, since this is just part of the Storybook story, which is just to provide a demonstration of what the component looks like, especially across multiple scenarios. Same goes for Navigation story.

src/components/grid-aware/ImageHeader/ImageHeader.tsx
```diff
@@ -10,16 +10,18 @@ type ImageProps = {
 type ImageHeaderProps = {
   title: string;
   subtitle: string;
-  description: string;
+  description?: string;
+  descriptionNode?: React.ReactNode;
```
richardxia: Let's just combine these two into one prop, since a `string` is a valid type to pass into a `ReactNode` prop. See https://github.com/facebook/react/blob/v17.0.2/packages/shared/ReactTypes.js#L10-L24 for the ReactNode type definition.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -90,6 +90,23 @@
   margin-top: 20px;
 }
 
+.descriptionWrapper :first-child {
+  margin: 0;
+}
```
richardxia: I don't think this does anything, since our CSS normalization stylesheet resets most basic elements' margins to 0. Could you remove this?

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -90,6 +90,23 @@
...
+
+.descriptionWrapper :not(:first-child):not(li) {
```
richardxia: Instead of using the `:not(li)` selector, I think what you really want is to use a direct child selector, `>`:

```suggestion
.descriptionWrapper > :not(:first-child) {
```

By default, nested selectors will select items that are arbitrarily deep, but here, I assume you want to select the first direct child of `.descriptionWrapper`, but not the first child of any level of the tree.

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -90,6 +90,23 @@
...
+  margin-top: 15px;
+}
+
+.descriptionWrapper ul {
+  margin-left: 2em;
+}
+
+.descriptionWrapper ul li {
```
richardxia: You may want to scope the selector so that we only select `li` elements that are the direct children of `ul` elements; otherwise, you may end up selecting things that are arbitrarily deep, such as in an `ol` that's under a `ul`.

```suggestion
.descriptionWrapper ul > li {
```

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -110,6 +127,11 @@
     margin-bottom: 20px;
     margin-right: 0;
   }
+
+  .descriptionWrapper {
+    font: var(--font-body-medium);
```
richardxia: This is the same font size used in the desktop view. Recall that we have the font variables themselves change based on the screen size, so this will automatically be adjusted for smaller screen sizes. Could you remove this line?

src/components/grid-aware/ImageHeader/ImageHeader.module.css
```diff
@@ -110,6 +127,11 @@
     margin-bottom: 20px;
     margin-right: 0;
   }
+
+  .descriptionWrapper {
```
richardxia: Nit: Could you keep the selectors within the media query in the same order as they appear outside the media query by moving this above the `.ctaButtonRow`? It makes it easier to scan the file if the selectors are in a consistent order.

src/components/grid-aware/ImageHeader/ImageHeader.tsx
```diff
@@ -28,9 +30,13 @@ const ImageHeader = ({
     <div className={s.gridAreaText}>
       <h1 className={s.title}>{title}</h1>
       <div className={s.subtitle}>{subtitle}</div>
-      <div className={s.description}>{description}</div>
+      <div className={s.description}>
+        {description || (
+          <div className={s.descriptionWrapper}>{descriptionNode}</div>
```
richardxia: I think we should be able to collapse `s.descriptionWrapper` into `s.description`, especially if you merge the `description` and `descriptionNode` props into one. Were there any significant styling differences between the string-only version and the non-string version?

src/pages/about/index.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from "react";
+import { Helmet } from "react-helmet";
+
+import ArticleSpotlightCard from "../../components/grid-aware/ArticleSpotlightCard";
+import ImageHeader from "../../components/grid-aware/ImageHeader";
+import Spacer from "../../components/grid-aware/Spacer";
+import teamDesignWorkshopImage from "../../components/grid-aware/ThreeParagraphBlock/stories/image2.png";
```
richardxia: Let's not import images from the Storybook story. We can make a copy of the image for now, since this will all go away soon with the CMS integration.

richardxia (commented):
Overall, this is looking pretty good! I wasn't sure how much to review of the CSS and ImageHeader changes, since you said that it was still a WIP, so feel free to disregard any comments I have if you're still making other changes to it.

candywang:
Hi @richardxia ! Thank you for taking the time to review everything. I've completed the CSS changes and have resolved all the comments.

richardxia (approved):
LGTM!

Like we discussed on the call today, this is OK for now, but we should work with Derek to figure out how to factor the components such that we don't have to create a separate set of CSS styles to handle the slight variations between pages.

PR #265, opened 2021-07-11
Implement programs page

### LINKED ISSUE #259: Build Programs Page
https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4216%3A0

This will also include building the new component on that page with the individual program listings.

Due date: 7/15

ivanleeswe:
Fix #259


src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.module.css
```diff
@@ -3,6 +3,7 @@
   grid-template-columns: var(--bleed-grid-template-columns);
   grid-template-rows: 600px;
   margin: auto;
+  margin-bottom: 120px;
```
richardxia: I think this change shouldn't be here, right? The Article Spotlight card doesn't even appear on the new Programs page, and adding this extra margin makes the bottom of the Home Page look like it has too much space.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+import Layout from "../../components/layout";
+import articleSpotlightImage from "./laura-barerra-vera.png";
+export default () => (
+  <Layout>
+    <ProgramsPageLargeParagraph
+      title="Our Programs"
+      description="ShelterTech's programs are designed to suppport the underserved community experiencing homelessness and housing insecurity through WiFi connectivity and acccess to online resources."
+    />
+    <ProgramsSubTitleBlock title="ShelterTech offers the following programs:" />
+    <SFServiceGuideBlock
```
richardxia: Ah, all three of these blocks (SF Service Guide, ShelterConnect, and Community Rep) should just use the same, single component definition but with different props, since they all have an identical visual structure. Although the main description may have a different number of paragraphs or bulleted lists, we can actually just represent that as a single `description` prop that accepts a JSX fragment, which would allow us to pass in different sets of `<p>`, `<ul>`, and other elements.

You can take a look at how we do this for the ImageHeader component as instantiated in the new [About Us page](https://github.com/ShelterTechSF/sheltertech.org/blob/dfd799813a656f4fd3be125e704a75b943fcbe4b/src/pages/about/index.tsx#L23-L68) and on the [Volunteer page](https://github.com/ShelterTechSF/sheltertech.org/blob/dfd799813a656f4fd3be125e704a75b943fcbe4b/src/pages/volunteer/index.tsx#L40).

Do you think you could go consolidate these into a single component and just provide them different props on the programs page? We could call the component the `ProgramBlock` or something like that.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      bullet3="Ongoing ISP charges"
+      description3="In 2019 we established a partnership with the City of San Francisco Digital Equity team to deliver 1GB fiber/ISP capability to future ShelterConnect locations at no cost.  Working with partners, ShelterConnect has implemented free WiFi solutions at over 20 locations including the Star Hotel, Mission Hotel, MSC-South, and ten locations managed by Larkin Street Youth Services."
+      image={{
+        url: classroom,
+        alt: "Classroom of volunteer members collaborating with one another.",
+      }}
+      ctaButtons={[
+        {
+          text: "DONATE",
+          onClick: () => console.log(true),
```
richardxia: Let's have all the donate buttons on the page take you to the Donate page.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        },
+        {
+          text: "PARTNER WITH US",
+          onClick: () => console.log(true),
```
richardxia: Can we have this pop up the same partnership modal as on the Home Page? It's the one in the "Get Involved" section, under the "Partnerships" subsection.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+      description="The SF Service Guide is an online directory of human services in San Francisco that launched in November 2019 and currently serves 1,500 unique monthly users. The directory's focus is on homelessness and housing services but also covers a variety of other services, from education and legal aid to senior services and re-entry programs. Our goal is to help anyone with access to smartphone, tablet, or computer to find services they need."
+      description2="SF Service Guide is supported by a grant from SF Mayor's Office of Housing and Community Development and verified for accuracy by our data partner, the Homeless Advocacy Project of the JDC."
+      image={{
+        url: classroom,
+        alt: "Classroom of volunteer members collaborating with one another.",
+      }}
+      ctaButtons={[
+        {
+          text: "Visit Site",
+          onClick: () => console.log(true),
```
richardxia: Let's have this be an external link that takes you to https://sfserviceguide.org/.

richardxia (commented):
Thanks for working on this! I think my main comment is that we should be defining a single React component that can be used with all three of the programs listed on the page, so I'm going to hold off from reviewing the three separate subcomponents that you have here for now. That said, I think you've done a great job faithfully implementing the designs, so I don't expect to have too many follow up comments about that once the components have been merged into one.

A few other comments I'd like to make are:

- Could you merge in the latest `main` branch? You currently have a conflict due to @candywang's About Us page being merged in and also touching the same part of the Navigation component
- Could you also update the footer navigation such that the Programs link is renamed to Our Programs and links to this new page?

src/components/grid-aware/ArticleSpotlightCard/ArticleSpotlightCard.module.css
ivanleeswe: Strange, don't know how that got there. Deleted.

src/components/layout.tsx
```diff
@@ -21,7 +21,7 @@ type NavItem = {
 
 const navigationItems: NavItem[] = [
   { text: "About Us", internalLink: "/about" },
-  { text: "Programs", internalLink: "/" },
+  { text: "Programs", internalLink: "/programs" },
```
richardxia: Could you update the link text to say "Our Programs"? You made the change for the footer, but not the top navigation.

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from "react";
+
+import * as s from "./ProgramsPageLargeParagraph.module.css";
+
+type ProgramsPageLargeParagraphProps = {
+  title: string;
+  description: string;
+};
+
+const ProgramsPageLargeParagraph = ({
```
richardxia: I think we could probably give this a better name, since this is more than just a paragraph, as it also has a heading. Since this plays a similar role as the `ImageHeader` and `VideoHeader` components — being the top-level heading of the page — what if we called this the `TextHeader`?

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.tsx
```diff
@@ -0,0 +1,28 @@
...
+  title,
+  description,
+}: ProgramsPageLargeParagraphProps) => {
+  return (
+    <div className={s.bleedWrapper}>
+      <div className={s.bleedMainContent}>
+        <div className={s.gridParent}>
+          <div className={s.textWrapper}>
+            <div className={s.title}>{title}</div>
```
richardxia: Since this definitely plays the role of being the top-level heading of the page, let's use the `<h1>` tag for this element. Note that you may need to update the CSS to override some default styling that comes from the `<h1>`.

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.stories.jsx
```diff
@@ -0,0 +1,19 @@
...
+  component: ProgramsPageLargeParagraph,
+};
+
+const Template = ({ title, description }) => (
+  <ProgramsPageLargeParagraph title={title} description={description} />
+);
+
+export const DefaultProgramsPageLargeParagraph = Template.bind({});
+DefaultProgramsPageLargeParagraph.args = {
+  title: "We believe connectivity is a human right.",
```
richardxia: Nit: This is much longer than the actual title of the actual Programs page, so much that I think it's a bit misleading because the rendered component in Storybook doesn't resemble the actual component on the Programs page. Could we shrink this down?

Here's what it looks like between the Storybook example and the actual Our Programs page, where the ratio between the area of the title and the area of the description is completely inverted:

## Storybook
Screen Shot 2021-08-02 at 9 15 16 PM

## Actual page
Screen Shot 2021-08-02 at 9 15 11 PM

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.module.css
```diff
@@ -0,0 +1,67 @@
...
+    grid-template-columns: 40px 1fr 40px;
+    padding-bottom: 40px;
+    padding-top: 30px;
+  }
+}
+
+.bleedMainContent {
+  grid-column: 2;
+  grid-row: 1;
+  text-align: center;
```
richardxia: Is there a reason this is here? It looks like you only override this in both the title and the description to set it back to the normal left-alignment.

If this doesn't have a purpose, can you delete this and the left-alignment overrides below?

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.module.css
```diff
@@ -0,0 +1,67 @@
...
+}
+
+@media (--tablet-and-down) {
+  .gridParent {
+    display: block;
+  }
+}
+
+.textWrapper {
+  grid-column: 2/ span 6;
```
richardxia: Nit: Put a space between the `2` and the `/` so that there's consistent spacing around the `/`.

```suggestion
  grid-column: 2 / span 6;
```

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.module.css
```diff
@@ -0,0 +1,67 @@
...
+  grid-column: 2;
+  grid-row: 1;
+  text-align: center;
+}
+
+.gridParent {
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  grid-template-rows: repeat(2, auto);
```
richardxia: You only have one grid row (both the title and the description are within the same grid cell), so you can remove this property that declares that there are two rows.

src/components/grid-aware/ProgramsPageLargeParagraph/ProgramsPageLargeParagraph.module.css
```diff
@@ -0,0 +1,67 @@
...
+}
+/* Override grid item default of auto so that item contents cannot stretch the grid itself */
```
richardxia: Nit: Could you add a blank line between CSS blocks?

```suggestion
}

/* Override grid item default of auto so that item contents cannot stretch the grid itself */
```

src/components/grid-aware/ProgramsSubTitleBlock/ProgramsSubTitleBlock.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+
+import * as s from "./ProgramsSubTitleBlock.module.css";
+
+type ProgramsSubTitleBlockProps = {
+  title: string;
+};
+
+const ProgramsSubTitleBlock = ({ title }: ProgramsSubTitleBlockProps) => {
```
richardxia: Nit: "Subtitle" is a single English word, so let's leave the "t" in title lowercase in the name of this component (and all the file names).

src/components/grid-aware/ProgramsSubTitleBlock/ProgramsSubTitleBlock.module.css
```diff
@@ -0,0 +1,66 @@
...
+
+/* Main grid */
+
+.gridParent {
+  color: var(--color-black);
+  column-gap: var(--grid-column-gap);
+  display: grid;
+  grid-column-start: 2;
+  grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
+  padding-bottom: 60px;
```
richardxia: Instead of making this padding, let's make it margin. Padding is the spacing that's "inside" of an element, while margin is the spacing that's "outside" of the element. When we don't have a border or background color, it's not as apparent visually, but it's still important to think about whether spacing is internal or external to a component.

src/components/grid-aware/ProgramsSubTitleBlock/ProgramsSubTitleBlock.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const GridAreaLeft = () => (
+    <div className={s.gridAreaLeft}>
+      <h1 className={s.title}>{title}</h1>
```
richardxia: Since this is found lower on the page, this shouldn't be an `<h1>` tag, which should be reserved for the page title, of which there should only be one on the page. Could you change this to be an `<h2>`?

src/components/grid-aware/ProgramBlock/ProgramBlock.tsx
```diff
@@ -0,0 +1,59 @@
...
+type ServiceGuideProps = {
+  url: string;
+  alt: string;
+};
+
+type ProgramBlockProps = {
+  title: string;
+  subtitle: string;
+  description: React.ReactNode;
+  image: ServiceGuideProps;
```
richardxia: It's a bit weird to name this type `ServiceGuideProps`, given that it has nothing to do with the Service Guide. Did you mean to name this `ImageProps`?

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+}
+
+.gridParent > * {
+  min-width: 0;
+}
+
+@media (--tablet-and-down) {
+  .gridParent {
+    display: flex;
+    flex-direction: column-reverse;
```
richardxia: Instead of reversing the flex direction, I think it'd be better if you just put the image before the text in the HTML. We should avoid doing unusual things like using `column-reverse` since it changes the layout order to not match the HTML order, which makes things a bit more tricky to read.

In addition, you may not even need to use flexbox at all, since if you're using a column flex direction and not using other flexbox-specific features, then this isn't really any different than `display: block`.

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+  }
+}
+
+
+/* Main grid areas */
+
+.gridAreaText {
+  grid-column: 6/ span 6;
```
richardxia: Nit: Let's put a space between the `6` and the `/`.

```suggestion
  grid-column: 6 / span 6;
```

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+.gridAreaImage {
+  grid-column: 2 / span 4;
+  grid-row: 1;
+}
+
+/* Left grid area */
+
+.title {
+  font: var(--font-body-large);
+  margin-bottom: 5px;
```
richardxia, line 69: I see that for several of the elements here, in order to space them apart you apply bottom margin, but for others you apply top margin. There's tradeoffs between the two approaches, but I generally find that it's a good practice to pick one direction and apply it consistently; that is, either primarily use bottom margin to separate components or primarily use top margin to separate components.

Could you pick a direction and make it consistent?

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+  margin-top: 0;
+}
+
+@media (--tablet-and-down) {
+  .title {
+    margin-top: 0;
+  }
+}
```
richardxia: This style doesn't seem to be necessary, since the non-media-query version of it also sets a top margin of 0. Could you delete this?

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+.subtitle {
+  font: var(--font-subtitle);
+}
+
+.description {
+  font: var(--font-body-medium);
+  margin-top: 20px;
+}
+
+.description :first-child {
```
richardxia: You should probably use the direct child selector, `>`, here so that you don't select all first children in the whole tree. For example, not only does this select the first child of `.description`, but if you have a `<ul><li>Item 1</li><li>Item 2</li></ul>`, then it'll also select that first `<li>` because it's the first child of `<ul>`.

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+  margin: 0;
```
richardxia: Setting a margin of 0 might be too extreme here. I think you are just looking to eliminate the top margin of the first element. Since any arbitrary HTML element can appear here, you don't want to overly override styles for it.

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+  margin-top: 20px;
+}
+
+.description ul {
+  margin-left: 1.5em;
+}
+
+.description ul > li {
+  list-style-position: outside;
+  list-style-type: disc;
```
richardxia: Three comments:

1. The default list-style-position is outside, so you don't need to explicitly set it here
2. You can just use the [`list-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style) shorthand property to set of the three specific `list-style-*` subproperties
3. You can define these styles on the `ul` element, which is the more common way of doing it unless if you really needed to set a different style on some, but not all `li` elements (which would be pretty unusual).

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+}
+
+.ctaButtonRow {
+  display: flex;
+  flex-flow: row wrap;
+  margin-top: 30px;
+}
+
+.ctaButtonRowItem:not(:last-child) {
+  margin-right: 30px;
```
richardxia: I _really_ want to say that we should just use the new `column-gap` property for flexbox here, since that'll allow us to avoid having to do this annoying `:not(:last-child)` selector. Originally, the [`gap`](https://developer.mozilla.org/en-US/docs/Web/CSS/gap) properties were only for CSS Grid, but it was recently updated so that they could apply to flexbox as well. Firefox started supporting gap for flexbox in 2018, and Chrome and Edge added it in 2020, but Safari only added support a few months ago. Overall, it looks like only 82% of users are on browser versions that support it: https://caniuse.com/flexbox-gap

How about we use `column-gap`, and we get bug reports from people on older browsers, then we go back to this `:not(:last-child)` approach?

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  .ctaButtonRowItem:not(:last-child) {
+    margin-bottom: 20px;
+    margin-right: 0;
+  }
+}
+
+
+/* Right grid area*/
```
richardxia: This comment doesn't seem right. The image is on the left side.

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+
+.image {
+  height: 250px;
+  width: 100%;
```
richardxia: These two properties by themselves are going to stretch and distort the images. You'll want to add [`object-fit: cover`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) so that the images are cropped to fit instead of stretched.

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,139 @@
...
+}
+
+@media (--tablet-and-down) {
+  .image {
+    height: 179.88px;
```
richardxia: I'm assuming that you got this from the Figma design, but I think that was not meant to be precise, so let's just round this up to 180px.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          alt: "SF Service Guide",
+        }}
+        ctaButtons={[
+          {
+            text: "Visit Site",
+            externalLink: "https://sfserviceguide.org/",
+          },
+          {
+            text: "Donate",
+            externalLink: "https://www.sheltertech.org/donate/",
```
richardxia: For this and all the other CTA links to the Donate page, could you use an internalLink instead of an externalLink? Making it an external link causes it to open in a new tab instead of directly taking you there.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+import ArticleSpotlightCard from "../../components/grid-aware/ArticleSpotlightCard";
+import Modal from "../../components/grid-aware/Modal";
+import ProgramBlock from "../../components/grid-aware/ProgramBlock";
+import ProgramsPageLargeParagraph from "../../components/grid-aware/ProgramsPageLargeParagraph";
+import ProgramsSubTitleBlock from "../../components/grid-aware/ProgramsSubTitleBlock";
+import Spacer from "../../components/grid-aware/Spacer";
+import Layout from "../../components/layout";
+import PartnershipSignupForm from "../../components/thirdparty/mailchimp/PartnershipSignupForm";
+import communityDevelopmentImg from "./community-development.jpg";
+import articleSpotlightImage from "./laura-barerra-vera.png";
```
richardxia: I think you have an older version of this image, where the image is cropped in a way where the Volunteer Spotlight box obscures her face. @candywang's very recent PR, https://github.com/ShelterTechSF/sheltertech.org/pull/267, updated the image such that Laura's face is further to the right, so could you pull in her updates and make sure you use this updated image?

richardxia (commented):
Thanks for making the changes to this PR! The consolidated ProgramBlock component is exactly what I was looking for, so thanks for making that change.

I've now gone through and reviewed the individual components in a lot more detail. I have a lot of little comments, but there's nothing major, so I think this is in pretty good shape to get merged in soon. Thanks again for working on this!

src/components/grid-aware/ProgramBlock/ProgramBlock.module.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+.description {
+  font: var(--font-body-medium);
+  margin-top: 20px;
+}
+
+/* .description :first-child {
+  margin-top: 0px;
+} */
```
richardxia: Let's delete this commented out style if it's no longer being used at all.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          alt: "SF Service Guide",
+        }}
+        ctaButtons={[
+          {
+            text: "Visit Site",
+            externalLink: "https://sfserviceguide.org/",
+          },
+          {
+            text: "Donate",
+            internalLink: "https://www.sheltertech.org/donate/",
```
richardxia: Can you leave off the `https://www.sheltertech.org` from all these internal links? Having the full domain here makes the link behave strangely in local development environments, since clicking on the link causes it to open the real, production site instead of navigating there in the local environment. This would also affect staging, as well as any version of the site that's not the production one.

There's two more `internalLink`s lower on the page that should be fixed up as well.

src/pages/programs/index.tsx
richardxia: I just wanted to bump this comment, since it still looks like you're using the older image. You can tell the difference because in the older one, the black box containing the "Volunteer Spotlight" text overlaps her face, while the newer one doesn't.

src/pages/programs/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+          alt: "Community Development",
+        }}
+        ctaButtons={[
+          {
+            text: "DONATE",
+            internalLink: "https://www.sheltertech.org/donate/",
+          },
+        ]}
+      />
+      <Spacer heightDesktop="20px" heightMobile=".37px" />
```
richardxia: What's up with this .37px spacing on mobile? If it shows up in the Figma design, then I think we can assume that it wasn't intentional, and that we can just round this to 0.

richardxia (commented):
Nice, this is getting close to being ready to merge. I have a few more comments for the latest batch of changes, and I wanted to bring up an older comment I made about using the newer photo of Laura for the Volunteer Spotlight component, since it looks like we're still using the older one.

Also, I just saw this image that doesn't appear to be used: src/pages/programs/mission-hotel.jpeg. Could you remove it if it's not in use?

richardxia (approved):
LGTM!

PR #268, opened 2021-07-16
Add support for Prismic link resolving

### LINKED ISSUE #235: Learn how to set up links to different pages in Prismic

richardxia:
Resolves #235. This targets the `prismic` branch, not `main`.

## Overview

This adds some support for Prismic link resolving, which allows us to take Prismic's representation of a link and map it onto our internal page routes and React components. There are a few layers of preexisting things in Prismic and the associated Gatsby plugin, so let me first explain them before going into what I did in this PR.

## Background

Prismic has its own concept of a [Link](https://prismic.io/docs/core-concepts/link-content-relationship), which content creators can use create links to other internal pages, external sites, or file uploads. These can live either in Link fields or in Rich Text fields, the latter of which provides a very high amount of flexibility on the content creator's side, and therefore means our site would needs to support that flexibility. These two scenarios are handled separately in Gatsby, since the use cases are different enough.

The `gatsby-source-prismic` has direct support for link resolving, and the v4 version of the plugin aligns more closely with Prismic's representation. The [current docs on Prismic's](https://prismic.io/docs/technologies/link-resolver-gatsby) website still reference the v3 version of the `gatsby-source-prismic` plugin, but I actually found Prismic's non-Gatsby, pure JavaScript docs on [link resolving](https://prismic.io/docs/technologies/link-resolver-javascript) has better documentation on the properties you can expect to get from the Prismic API. Some of the more important fields are the `type`, which tells you which custom type on Prismic it corresponds to, the `uid` and the user-friendly ID for resource being linked to.

Finally, there's one last property that they return that I couldn't find much documentation for, but it's the `link_type`, which can be one of `Document` (internal Prismic document), `Web` (external link), or `Media` (file upload).

## Changes made in this PR

There were two utilities that I created: a `linkResolver()` function and a `<PrismicLink>` React component.

The `linkResolver()` just takes in a Prismic link object and returns a string URL, and it's used in the `gatsby-config.js` for configuring `gatsby-source-prismic` itself. It currently only returns the URL for the home page, but as we add new pages, we'll need to add them to the `linkResolver()` function so that they are routed properly.

The `<PrismicLink>` React component I created is meant to take any arbitrary Prismic link and return either a `<GatsbyLink>`, for internal pages, or a regular `<a>` tag, for external pages or media files. So far I've only used it manually in one of our components, but I think this is something we can also pass into the rendering of a Rich Text field, which the `gatsby-source-prismic` plugin can handle. I haven't tested the Rich Text field integration yet, but I think that can happen in a followup PR.

I made a minor change to the footer so that the ShelterTech logo is now a link to the Home Page, which is configured as a Prismic Link on our Footer type in Prismic.

I think the last thing to note is that you have to query quite a few fields within the Prismic field type to so that we have all the fields necessary for link resolver. As far as I can tell, we just have to explicitly write these fields down in the GraphQL query, but is there some way for us to specify the same set of fields for all link types?

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -45,6 +45,13 @@ const Layout = ({ children }: LayoutProps) => {
           address {
             text
           }
+          sheltertech_logo_link {
+            isBroken
+            link_type
+            target
+            type
+            uid
+          }
         }
       }
     }
@@ -112,6 +119,7 @@ const Layout = ({ children }: LayoutProps) => {
           shelterTechLogo={{
             url: shelterTechLogoWhite,
             alt: "ShelterTech Logo",
+            link: data.prismicFooter?.data?.sheltertech_logo_link,
           }}
           socialMediaLinks={[
             {
--- a/src/utils/linkResolver.js
+++ b/src/utils/linkResolver.js
@@ -1,0 +1,27 @@
+// Note: This file cannot be written in TypeScript because it is imported by
+// gatsby-config.js.
+
+/** A link resolver function for Prismic.
+ *
+ * Accepts a description of an arbitrary Prismic document of any type and
+ * returns a string URL to the document.
+ *
+ * See https://prismic.io/docs/technologies/link-resolver-javascript for the
+ * available properties on the `doc` argument.
+ */
+const linkResolver = (doc) => {
+  switch (doc.type) {
+    case "home_page":
+      return "/";
+    default:
+      // eslint-disable-next-line no-console
+      console.warn(
+        `Could not resolve the following document to a URL: ${JSON.stringify(
+          doc
+        )}`
+      );
+      return "/";
+  }
+};
+
+module.exports = linkResolver;
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,5 +1,7 @@
 const postcssCustomMedia = require("postcss-custom-media");
 
+const linkResolver = require("./src/utils/linkResolver");
+
 module.exports = {
   siteMetadata: {
     title: "ShelterTech - Technology for the underserved",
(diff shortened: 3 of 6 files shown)


src/components/layout.tsx
```diff
@@ -45,6 +45,13 @@ const Layout = ({ children }: LayoutProps) => {
           address {
             text
           }
+          sheltertech_logo_link {
```
jessica-px, line 48: It isn't at all a priority, but I do agree it would be nice if we could encapsulate this query somehow. Maybe we can throw a ticket in the backlog for investigating that?

src/utils/linkResolver.js
```diff
@@ -0,0 +1,27 @@
...
+ * Accepts a description of an arbitrary Prismic document of any type and
+ * returns a string URL to the document.
+ *
+ * See https://prismic.io/docs/technologies/link-resolver-javascript for the
+ * available properties on the `doc` argument.
+ */
+const linkResolver = (doc) => {
+  switch (doc.type) {
+    case "home_page":
+      return "/";
```
jessica-px, line 15: What are you thinking this will look like when we have dynamic urls, like for blog posts?

jessica-px (approved):
Thanks for adding so much context in the PR description! This looks like a really reasonable approach to me -- I just left a couple comments, but this looks good to merge 👍

src/components/layout.tsx
richardxia: Yep, good call. Created a ticket to capture that: https://github.com/ShelterTechSF/sheltertech.org/issues/275

src/utils/linkResolver.js
richardxia: For dynamic URLs, I think we'd want to use the `uid` field on the `doc` object, which Prismic's docs say hold the user-friendly unique ID. I think it'd look something like:

```js
switch (doc.type) {
  case "blog_post":
    return `/blog/${doc.uid}`;
}
```

Gatsby has its own [special support for dynamic pages](https://www.gatsbyjs.com/docs/tutorial/part-seven/) as well, which I haven't used before, but we'll want to make those match the Prismic link resolver.

richardxia:
Thanks for reviewing this! I'm going to merge this in, since there are no blocking comments.

PR #278, opened 2021-08-24
Update stats block theme

### LINKED ISSUE #269: Update StatsBlock to allow for light variant

Screen Shot 2021-07-24 at 10 37 22 AM

The new design for the Volunteer Page has a new version of the StatsBlock component with a lighter color scheme. The scope of this ticket is to add a new prop to the component to control light vs. dark, and to update the version of it on the Volunteer Page to use the light variant.

candywang:
Closes #269.

### DIFF
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -94,11 +94,11 @@
 }
 
 .number {
+  color: var(--color-sheltertech-blue);
   font-family: var(--font-family-lato);
   font-size: 72px;
   font-weight: bold;
   line-height: 1.3;
-  color: var(--color-sheltertech-blue);
 }
 
 .statement {
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -1,3 +1,16 @@
+.light {
+  color: var(--color-black);
+}
+
+.lightNumber {
+  color: var(--color-sheltertech-blue);
+}
+
+.dark {
+  background-color: var(--color-black);
+  color: var(--color-white);
+}
+
 .bleedWrapper {
   display: grid;
   grid-template-columns: var(--bleed-grid-template-columns);
@@ -11,7 +24,6 @@
 }
 
 .bleedBackground {
-  background-color: var(--color-black);
   grid-column: 2 / span 2;
   grid-row: 1;
 }
@@ -23,7 +35,6 @@
 
 @media (--tablet-and-down) {
   .bleedWrapper {
-    background-color: var(--color-black);
     grid-template-columns: 40px 1fr 40px;
   }
 }
@@ -31,7 +42,6 @@
 /* Main grid */
 
 .gridParent {
-  color: var(--color-white);
   column-gap: var(--grid-column-gap);
   display: grid;
   grid-template-columns: repeat(var(--grid-column-count), minmax(0, 1fr));
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -1,14 +1,5 @@
-.light {
-  color: var(--color-black);
-}
-
-.lightNumber {
-  color: var(--color-sheltertech-blue);
-}
-
-.dark {
-  background-color: var(--color-black);
-  color: var(--color-white);
+.gray {
+  background-color: var(--color-gray-300);
 }
 
 .bleedWrapper {
@@ -107,6 +98,7 @@
   font-size: 72px;
   font-weight: bold;
   line-height: 1.3;
+  color: var(--color-sheltertech-blue);
 }
 
 .statement {
--- a/src/components/grid-aware/StatsBlock/StatsBlock.tsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.tsx
@@ -5,11 +5,14 @@ import * as s from "./StatsBlock.module.css";
 type StatCardProps = {
   number: string;
   statement: string;
+  theme?: string;
 };
 
-const StatCard = ({ number, statement }: StatCardProps) => (
+const StatCard = ({ number, statement, theme }: StatCardProps) => (
   <div className={s.statCard}>
-    <div className={s.number}>{number}</div>
+    <div className={`${s.number} ${theme === "light" ? s.lightNumber : ""}`}>
+      {number}
+    </div>
     <div className={s.statement}>{statement}</div>
   </div>
 );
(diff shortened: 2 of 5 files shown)


src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -1,3 +1,16 @@
...
+}
+
+.lightNumber {
+  color: var(--color-sheltertech-blue);
+}
+
+.dark {
+  background-color: var(--color-black);
+  color: var(--color-white);
+}
```
richardxia: We actually don't have a dark theme anymore for components in the latest designs. We had made the conscious decision to replace the black background style with a much lighter gray style, so now the only two styles you'll see on the site are a white background or a light gray one. Although the StatsBlock component doesn't have an example of the lighter gray background, I figured it may still be useful to make the background parametrizable.

The plus side is that we then don't need to worry about making the text color parametrizable, since the black text works well on either white or light gray background colors.

Could you remove the variant with the black background, and instead replace it with a light gray variant?

In terms of the naming, since both the white background and the gray background are "light" colors, what do you think about just naming the themes "white" and "gray", respectively?

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -5,29 +5,44 @@ import * as s from "./StatsBlock.module.css";
...
+      {number}
+    </div>
     <div className={s.statement}>{statement}</div>
   </div>
 );
 
 type StatsBlockProps = {
   title: string;
   statCards: StatCardProps[];
+  theme: string;
```
richardxia: I was thinking that it might be a good idea to make the allowed theme colors stricter than just `string`. TypeScript has a neat feature called [literal types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types), which allows you to use individual string values as types. For example, `"white"` would be a valid type, and it would enforce that any variable with the type of `"white"` is only ever allowed to have the string value of `"white"` passed to it.

By themselves, literal types aren't super useful, but defining a _union_ of literal types allows you to specific a much smaller set of valid values, for example `"white" | "gray"`. This would cause a type error if I passed in a string such as `"black"` to it, and it would allow allow you to pass in `"white"` or `"gray"`.

What do you think?

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -5,29 +5,44 @@ import * as s from "./StatsBlock.module.css";
...
 };
 
-const StatsBlock = ({ title, statCards }: StatsBlockProps) => (
-  <div className={s.bleedWrapper}>
-    <div className={s.bleedBackground} />
+const StatsBlock = ({ title, statCards, theme }: StatsBlockProps) => (
+  <div
+    className={`${s.bleedWrapper} 
+      ${theme === "light" ? s.light : s.dark}`}
```
richardxia: Instead of doing this, if we're careful about having the theme names match the CSS class names, we could use `theme` to index into the `s` object, for example, `${s[theme]}`. Recall that in JavaScript, `s.light` is the same as `s["light"]`, so this allows us to more concisely choose an object property dynamically.

On a separate note, this is making we want to use the `classnames` library again, which is the work you had started in https://github.com/ShelterTechSF/sheltertech.org/pull/153. If you agree, do you want to revive that PR and get it merged in first?

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -5,29 +5,44 @@ import * as s from "./StatsBlock.module.css";
 type StatCardProps = {
   number: string;
   statement: string;
+  theme?: string;
```
richardxia: It shouldn't be necessary to make this optional, since the property isn't optional on the parent component, so we'll always end up passing it in.

richardxia (commented):
Thanks for getting this started! I think the biggest requested change I have is to replace the dark theme with a gray theme, since gray and white will be the background colors of all components moving forward.

Sorry for specifying light and dark in the original issue! I had written that before I confirmed with Derek that we'll only have white and gray backgrounds in the future, and I forgot to update the issue description after finding out.

richardxia:
Oh, one last request: Could you update the Storybook story to pass in the `theme` prop? That will make it easier to toggle between the different colored backgrounds in Storybook.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
jessica-px: If we wanted added consistency, we could also define a single type somewhere, like `type ThemeColorOptions = 'white' | 'black';` and import that across different components...it would be an extra import to keep track of, but would make it even harder to mistype things, and if for whatever reason we ever needed to change these options, it would only have to be in one place.

But I don't know if we have a good place to define a universal type like that? And it's not like this is a huge app, so maybe it's fine to type these things on the fly 🤔

richardxia: That sounds good to me. Maybe we can just create a new file at `src/types.ts`?

src/components/grid-aware/StatsBlock/StatsBlock.tsx
candywang: Without the `theme` prop being optional, it requires each stat card to pass in a `theme` prop. Is there a better way of passing in the theme from `<StatsBlock>` to `<statCards>`?

```
<StatsBlock
        title="See what our volunteers have accomplished together."
        statCards={[
          {
            number: "9,000",
            statement:
              "At-risk community members can now access services and amenities",
            theme="white",
          },
          {
            number: "560",
            statement:
              "Bay Area volunteers have contributed to our cause since our founding in 2016",
            theme="white",
          },
          {
            number: "$2M",
            statement:
              "Worth of work provided by volunteers has been put towards our programs",
            theme="white",
          },
        ]}
        theme="white"
      />
```

src/components/grid-aware/StatsBlock/StatsBlock.tsx
candywang: When the theme is not "white", do we still want the stat number color to be "sheltertech-blue"?

src/types.ts
```diff
@@ -0,0 +1 @@
+export type ThemeColorOptions = "white" | "gray";
```
richardxia: Very tiny nit, but could you rename the type to `ThemeColorOption` (without the "s")? Even though this contains more than one option, the name of the _type_ should be singular, since it'll be used to describe a single object. For example, `const theme: ThemeColorOption = "white"`.

One way of thinking about it is that types and interfaces should be using the same kind of naming convention as classes, and you would give a class a singular name (e.g. `class Card`) even if you instantiate many objects from that class.

richardxia (approved):
One very minor request, but otherwise this LGTM! Feel free to merge this after fixing that up.

src/types.ts
candywang: Thanks for pointing this out!

PR #280, opened 2021-09-04
Enable prismic previews

jessica-px:
## Description

This PR enables Prismic previews! Right now, only the Home Page is previewable, but future pages will be easy to enable -- we just have to add `_previewable` to the GraphQL query and wrap the exported page with `withPrismicPreview()`.

Note that non-page things cannot be previewed with this setup -- so previewing the "Footer", for example, won't work.

Note also that we need to add a schema for _every_ Prismic type in `gatsby-config.js`, or previews will break...even if we deleted the type, we at least need to assign it an empty object `{}`. If you start getting errors when trying to preview something in Prismic, this is the first place you should look -- make sure everything has a schema, and all of the schemas are up to date.

## Testing Steps

- Check out this branch and run it locally
- Visit the Prismic website
- Under "Documents" click on "Home Page"
- Confirm that it has visible unpublished changes (currently the header title has some exclamation points)
- Click the preview "eye" icon in the upper right and select "Local" ("Production" won't work until this is merged)
- A new tab should open with the preview, showing the unpublished changes


src/pages/index.tsx
```diff
@@ -215,3 +218,10 @@ export default ({ data }: PageProps<GatsbyTypes.HomePageQuery>) => {
     </Layout>
   );
 };
+
+export default withPrismicPreview(IndexPage, [
+  {
+    repositoryName: "sheltertech",
+    linkResolver,
+  },
+]);
```
richardxia, line 227: It's kind of annoying that we need to repeat these arguments for each page. Could we write our own HOC that wraps `withPrismicPreview` by providing the `repositoryName` and `linkResolver` options? It seems that the API they provide allows you to have one app connected to multiple Prismic sites, but since I don't see us needing to support mulitple Prismic sites, I think it'll be much simpler to have a HOC that provides the one Prismic site config options.

richardxia (approved):
I tested this out locally and can confirm that this works for me. Thanks for working on this!

Aside from my comment about possibly creating a custom HOC, I was wondering about a couple things that I see in the Prismic docs for setting up previews with Gatsby (https://prismic.io/docs/technologies/previews-gatsby):

1. It looks like they talk about adding a preview resolver page. Is this something that we need? I'm wondering if this is working for us only because we've only tested the home page, which is the default page.
2. It also looks like they talk about modifying the 404 page so that previews can still work for pages that haven't been published yet. I think this would be useful for things like blog posts, where you won't have published the post yet, and therefore won't have a page to navigate to yet. I think the preview stuff normally kicks in after the initial page load, so this probably is necessary for those kinds of Prismic types.

I don't think either of these things are blockers for this PR, but I think we'll want to track these as tickets to work on.

jessica-px:
@richardxia Regarding the resolver page and the 404 page, I do expect we'll need to add these at some point, but I wanted to keep this first implementation minimal. But yeah we'll especially need them for the blog posts.

I like the idea of a HOC! I can make a ticket for it

PR #282, opened 2021-09-05
Make subtitle prop optional and remove from Programs page

### LINKED ISSUE #281: Remove subtitle text from Program Cards in Programs page
On the new Programs page, the feedback we got from the Marketing team was to remove the subtitle text under each of the cards on the page today:

Screen Shot 2021-09-05 at 10 10 26 AM

This should be done for all of the program cards. In addition, we may need to adjust the margins/spacing for the cards to make sure there isn't too large of a gap between the program name and the main description bodies.

candywang:
Closes #281.


richardxia (approved):
LGTM, thanks for the quick turnaround on this!

PR #285, opened 2021-09-09
Reduce boilerplate for Prismic link queries

### LINKED ISSUE #275: Reduce boilerplate for querying Prismic Links
As @jessica-px commented on in https://github.com/ShelterTechSF/sheltertech.org/pull/268#discussion_r678821871, currently it requires a bit of boilerplate to query Prismic Links, since there are several fields that we'll always need to pull off the Link object for use in link resolving. The scope of this ticket is to see if there's some way to reduce that boilerplate.

richardxia:
Fixes #275.

This is actually two completely separate changes, where the second change isn't really related to the associated issue, so you may find reviewing the individual commits easier.

The main changes are in https://github.com/ShelterTechSF/sheltertech.org/commit/e47ff466dee714329dbb7d39481cdaa1107b6187, where I created a GraphQL fragment named `MinimalPrismicLinkData` to hold the minimum set of fields needed needed for rendering the `<PrismicLink>` component I had created in an earlier PR. It's a bit of Gatsby magic, but simply exporting a variable named `query` in any component that is eventually imported by a page is enough for the fragment to be registered, which is why it is legal to reference `MinimalPrismicLinkData` from the query in `layout.tsx`. See https://prismic.io/docs/technologies/fragments-gatsby for some docs on how this works in the Gatsby Prismic plugin we use.

The second commit (https://github.com/ShelterTechSF/sheltertech.org/commit/f11fca34a91aa9291828a559b01dce7fd1c71e6a) is technically unrelated, and I can split it out into a separate PR if preferred. Here I significantly trimmed down on the number of fields being queried, since I discovered that the Gatsby Prismic plugin adds a special GraphQL field named `url` to PrismicLinkType objects. This actually holds the URL after being resolved by the `linkResolver()` function. This is once again a bit of magic, this time [by the Gatsby Prismic plugin](https://prismic.io/docs/technologies/link-resolver-gatsby#3.-usage), but I figured out that if we use this, we don't need the `<PrismicLink>` component to import and call the `linkResolver()` function on its own. When we remove the dependency on the `linkResolver()`, we can drop most of the fields we were querying before.

Note that I was hitting some transient issues with my local development setup, so it's possible that you may need to run `npx gatsby clean` to clear your local cache in order to test this out locally.

### DIFF
--- a/src/components/inline/PrimsicLink.tsx
+++ b/src/components/inline/PrimsicLink.tsx
@@ -1,4 +1,4 @@
-import { Link as GatsbyLink } from "gatsby";
+import { Link as GatsbyLink, graphql } from "gatsby";
 import React from "react";
 
 import linkResolver from "../../utils/linkResolver";
@@ -45,5 +45,16 @@ const PrismicLink = ({ linkData, children }: PrismicLinkProps): JSX.Element => {
       );
   }
 };
+
+export const query = graphql`
+  # The minimum set of data required for the <PrismicLink> component.
+  fragment MinimalPrismicLinkData on PrismicLinkType {
+    isBroken
+    link_type
+    target
+    type
+    uid
+  }
+`;
 
 export default PrismicLink;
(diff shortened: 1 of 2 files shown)


jessica-px (approved):
Tested, LGTM! Thanks for tidying this stuff up, Richard!

richardxia:
Thanks for the review!

PR #288, opened 2021-09-27
Update volunteer page

### LINKED ISSUE #274: Update all copy on Volunteer Page

After we make the updates to the visual style of the components, let's update all the copy on the Volunteer Page to match the Figma design. The changes range from "completely rewritten" to "slightly reworded", so we should carefully examine the copy on Figma and make the code match.

ivanleeswe:
Fix #274


richardxia:
@ivanleeswe, it looks like there are a lot of unrelated changes in the PR diff, and you have merge conflicts as well. Could you fix it up so that this only contains the copy changes? When starting new work, always remember to pull the latest changes to the `main` branch before creating a feature branch so that you're always working from the latest changes.

PR #286, opened 2021-09-19
Update OneParagraphBlock to have a gray variant #272

davidagustin:

PR #289, opened 2021-09-27
Update programsblock and programcard's theme

### LINKED ISSUE #273: Update ProgramsBlock to have a light variant
https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=1673%3A1602

Screen Shot 2021-07-24 at 11 14 49 AM

The Home Page has a slightly updated design where the ProgramsBlock now has a white background. We should add a prop to control the theme of the block.

ivanleeswe:
Fix #273

### DIFF
--- a/src/components/block/ProgramCard/ProgramCard.module.css
+++ b/src/components/block/ProgramCard/ProgramCard.module.css
@@ -3,11 +3,11 @@
 }
 
 .light {
-  color: var(--color-black);
+  color: var(--color-white);
 }
 
 .dark {
-  color: var(--color-white);
+  color: var(--color-black);
 }
 
 .img {
--- a/src/components/block/ProgramCard/ProgramCard.module.css
+++ b/src/components/block/ProgramCard/ProgramCard.module.css
@@ -2,14 +2,6 @@
   max-width: 410px;
 }
 
-.light {
-  color: var(--color-white);
-}
-
-.dark {
-  color: var(--color-black);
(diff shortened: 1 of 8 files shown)


richardxia:
Hey @ivanleeswe, sorry if the ticket wasn't clear, but I think it'd be a good idea to make the the background color configurable via a React prop, similar to all the other similar tickets. For example, see @candywang's older PR https://github.com/ShelterTechSF/sheltertech.org/pull/278 or @davidagustin's open PR https://github.com/ShelterTechSF/sheltertech.org/pull/286.

Do you think you could add a prop to choose between a white background and a light gray background like those PRs?

richardxia:
I updated this PR by rebasing it on `main` and adding a `theme` prop to the `<ProgramsBlock>` component, similar to the other components we've been updating. I also ended up removing the `theme` prop from the `<ProgramCard>` component, since the only remaining thing it affected was the text color, which no longer needs to be configurable because the components always have a lighter background and therefore the text color can stay with the site's default darker color.

If anyone wants to take a quick look at this, that'd be great, but otherwise I'll probably merge it in a few days just so that we can do a push to prod by the end of the year, with all of the other style updates.

Thanks again, @ivanleeswe, for starting the work on this PR!

richardxia:
Going to merge this in, as to not block shipping the overall updates to the page.

PR #290, opened 2021-09-27
Update layout and styling

### LINKED ISSUE #270: Update ThreeParagraphBlock layout

Screen Shot 2021-07-24 at 10 49 20 AM

The new design for the ThreeParagraphBlock changes all the text to be laid out in a single column, instead of spread across two separate columns. This ticket should update that component to lay out the text this way.

Please consult the latest Figma designs for both the Home Page and the Volunteer Page, since they have slightly different variations of the component (mostly with respect to the CTA buttons), so we should make sure both styles are updated to the vertical layout for text.

candywang:
I've updated the Volunteer page with the latest Figma design. However, one last thing to change/point out is the Home page, specifically the "Get involved" text and the solo image on the left.

Questions:
1. Does the "Get involved" text span 3 or 4 columns? According to Figma, it looks like it spans 4 columns and if that's the case, I can add a new class to differentiate `.gridAreaLeft` between the Home and Volunteer page.
2. Do we want to keep all images to `max-width: 100%';`?
3. What's the best way to have the image on the Home page stretch to the end of the third span?
4. What's the best way to ensure all images have a min or max-width of 100% without knowing the size of the actual image?


richardxia:
> 1. Does the "Get involved" text span 3 or 4 columns? According to Figma, it looks like it spans 4 columns and if that's the case, I can add a new class to differentiate `.gridAreaLeft` between the Home and Volunteer page.

It does look like it's slightly wider than three columns, so it must be four. However, I don't think it's worth creating separate CSS classes for the two variations. If it looks all right, could we just use four columns for both pages? I'm assuming that this would look better than using three columns for both pages, since having a line break in "Get Involved" would look weirder than "How we work" all being on a single line.

> 2. Do we want to keep all images to `max-width: 100%';`?

We should definitely keep that around, since otherwise, if you upload a larger image, it'll overflow past the width of the parent element.

> 3. What's the best way to have the image on the Home page stretch to the end of the third span?

It's hard to say because it's an unusual layout, and it's not entirely obvious how the component should behave with other kinds of images. I think one approach would be to not worry about getting the image to be exactly aligned with the grid column, and just add some right margin to push it away from the right edge of the parent element.

> 4. What's the best way to ensure all images have a min or max-width of 100% without knowing the size of the actual image?

Could you elaborate on what you mean by this question? A min or max width of 100% means 100% of the parent element's width, which means both properties are already unaffected by the dimensions of the actual image.

src/pages/volunteer/index.tsx
```diff
@@ -79,34 +78,34 @@ export default () => {
         theme="white"
       />
       <ThreeParagraphBlock
-        title="Let’s work together"
+        title="How we work"
```
richardxia: Could you undo all the changes to the text here? We actually have a separate ticket for this, which @ivanleeswe is working on in https://github.com/ShelterTechSF/sheltertech.org/pull/288, where he'll be updating all of the copy on the Volunteer page.

src/pages/volunteer/index.tsx
```diff
@@ -79,34 +78,34 @@ export default () => {
...
           alt: "Team bonding activity night.",
         }}
-        rightImage={{
-          url: pairProgrammingImage,
-          alt: "Two volunteers pair programming.",
-        }}
+        // rightImage={{
+        //   url: teamInClassroom,
+        //   alt: "Team members in a classroom setting.",
+        // }}
```
richardxia: Since the right image no longer exists, and there's no room for any images on the right side anymore, could you just delete this commented out code?

richardxia (commented):
Thanks for making these changes!

I left a few small comments on the code, but because I think we may want to set the left column to a width of four columns, not three, I wanted to hold off from doing a deep review of the CSS changes.

However, one other thing I wanted to point out is that I think you're missing some spacing between the paragraphs. The Figma designs show a 50px vertical gap between the paragraphs, whereas when I checked out your branch locally, it looks like there's only a 25px gap.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -137,21 +137,18 @@
   position: relative;
 }
 
-.leftBottomImageWrapper.solo {
-  margin-top: 40px;
+.leftBottomImageWrapper.paired {
+  margin-top: -30px;
+  min-width: 100%;
```
richardxia: Is this min-width necessary? Setting a min-width of 100% is fairly unusual, since it means you're forcing the element to be as large as or larger than its parent (minus additional padding + margin).

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -166,22 +163,19 @@
...
-    margin-top: 40px;
     max-width: 100%;
   }
 
   .leftTopImageWrapper {
-    display: none;
+    margin-right: -40px;
+    margin-top: 30px;
+    position: relative;
+    right: 0;
```
richardxia, line 177: It looks like these two properties are being set to undo the effects of the same properties being set outside of the media query. The fact that these are undoing the other properties is not obvious from just reading the code inside the media query here, and I myself didn't even realize it until viewing the full set of CSS properties being applied to the element in the browser.

To make this a bit more maintainable, could we alter the `position` and `right` CSS properties being applied to `.leftTopImageWrapper` outside of the media query by putting them inside of a `--laptop-and-up` media query? This way, the effects of the relative positioning are more self-contained and we don't have to worry about undoing them in the smaller screen sizes.

src/components/grid-aware/ThreeParagraphBlock/ThreeParagraphBlock.module.css
```diff
@@ -118,7 +118,7 @@
 }
 
 .paragraph3Wrapper {
-  margin-top: 25px;
+  margin-top: 50px;
 }
 
 .ctaTitleBlock {
```
richardxia: On the line just below this, we're setting a max-width. Could we remove this altogether? I think the reason this was here before was that we were trying to micro-optimize the line breaking of the text, but in the new layout, this should be wider than it used to be. It looks like the natural line breaking looks fine for the current copy, so I think we can just remove the max-width completely.

richardxia (approved):
Just a few minor comments, but this otherwise LGTM! Feel free to merge as soon as you address them.

PR #291, opened 2021-10-03
Add document on adding a new Prismic type.

### LINKED ISSUE #251: Write up rough notes on adding custom Prismic types

richardxia:
Closes #251.

This adds some documentation on the steps required to add a new Prismic type. I'm not actually sure if this is 100% complete, but I think it at least provides a starting point, especially for things that are easy to forget when creating a new Prismic type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,13 @@ For more specific documentation, visit one of the pages below:
 - [Command-Line Quick Reference][]
 - [Component Design System][]
 - [How To Deploy][]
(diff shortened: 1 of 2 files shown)


jessica-px (approved):
We're probably going to want to expand on this with how to query/type content from Prismic, but this is a perfectly fine starting point 👍 Sorry I'm only just now seeing it 😅

richardxia:
No worries, and thanks for taking a look at it!

PR #296, opened 2021-11-13
Add dynamic page for blog posts

### LINKED ISSUE #293: Dynamic Routing for Prismic Blog Posts
Let's figure out how to get each blog post in Prismic associated with a unique route in Gatsby. [Relevant docs here.](https://prismic.io/docs/technologies/create-pages-and-urls-gatsby#dynamic-pages)

jessica-px:
## Description

Closes #293 

Adds the dynamic route `/blog/{slug}` for our blog posts (where `slug` is our label for the Prismic `uid` field). If you were to check this out locally, both http://localhost:8000/blog/mission-hotel-free-internet/ and http://localhost:8000/blog/cool-blog-post should render pages with their respective blog titles and authors.

This PR doesn't put all the blog content on the page, nor does it style it -- I only added enough JSX for a proof of concept. I did make the decision to break the JSX out into its own component (`<BlogPostTemplate />`), because the query in `src/pages/blog/{prismicBlogPost.uid}.tsx ` is very long, so I felt one or the other should live in a different file.

## Notes
- If you add a new blog post in Prismic, or change the slug on an existing blog post, you'll need to restart Gatsby to be able to see it locally
- [The Prismic docs](https://prismic.io/docs/core-concepts/uid) _say_ they keep a history of old uids/slugs, so old ones should still direct to the right page, but I couldn't get this to work. If you want to test for yourself, http://localhost:8000/blog/cool-blog-post used to be http://localhost:8000/blog/test-blog-post -- my understanding is that both of these should work, but right now only the newer one works. I tried Googling it and saw a lot of other people going to Prismic support with questions about this, so it might not be a very stable feature...

## Open Questions
- What do we want to do when the query fails/doesn't return anything? Should we redirect to the 404 page?
- Prismic doesn't allow us to make fields required, but there are some that we probably _do_ want to have present (ie "title", "author"). What should we do if these return empty? Should we fail loudly, or quietly in these cases?

### DIFF
--- a/src/pages/blog/{prismicBlogPost.uid}.tsx
+++ b/src/pages/blog/{prismicBlogPost.uid}.tsx
@@ -1,0 +1,120 @@
+import { graphql, PageProps } from "gatsby";
+import React from "react";
+import BlogPostTemplate from "./BlogPostTemplate";
+
+/**
+  This is the setup for our dynamically generated blog post pages.
+  All urls of the pattern "/blog/{uid}"" point here and perform the query on this page.
+  Refer to the following docs:
+  https://www.gatsbyjs.com/docs/reference/routing/file-system-route-api/
+  https://prismic.io/docs/technologies/tutorial-4-add-dynamic-pages-gatsby
+*/
+
+export const query = graphql`
+  query PrismicBlogPost($uid: String!) {
+    prismicBlogPost(uid: { eq: $uid }) {
+      data {
+        author {
+          text
+        }
+        header_image {
+          alt
+          copyright
+          url
+        }
+        title {
+          text
+        }
+        publish_date
+        body {
+          ... on PrismicBlogPostDataBodyStatsBlock {
+            id
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodyImageWithCaption {
+            primary {
+              image {
+                url
+                alt
+              }
+              caption {
+                raw
+              }
+            }
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodyCtaBlock {
+            id
+            primary {
+              button_link {
+                url
+              }
+              button_text {
+                text
+              }
+              header {
+                text
+              }
+            }
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodySeparator {
+            id
+            primary {
+              separator_image {
+                url
+              }
+            }
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodyQuoteBlock {
+            id
+            primary {
+              quote {
+                text
+              }
+            }
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodyTextBlock {
+            id
+            primary {
+              body_text {
+                html
+              }
+            }
+            slice_type
+          }
+          ... on PrismicBlogPostDataBodyFileDownloadBlock {
+            id
+            primary {
+              file {
+                url
+              }
+              file_download_header {
+                text
+              }
+            }
+            slice_type
+          }
+        }
+      }
+    }
+  }
+`;
+
+const PrismicBlogPostPage = ({
+  data,
+}: PageProps<GatsbyTypes.PrismicBlogPostQuery>) => {
+  if (!data?.prismicBlogPost?.data) return <h1>There was a problem</h1>;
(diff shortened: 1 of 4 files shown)


src/pages/blog/{prismicBlogPost.uid}.tsx
```diff
@@ -0,0 +1,120 @@
...
+const PrismicBlogPostPage = ({
+  data,
+}: PageProps<GatsbyTypes.PrismicBlogPostQuery>) => {
+  if (!data?.prismicBlogPost?.data) return <h1>There was a problem</h1>;
+  const blogData = data.prismicBlogPost.data;
+
+  return (
+    <BlogPostTemplate
+      title={blogData?.title?.text || ""}
+      author={blogData?.author?.text || ""}
```
richardxia, line 115:
```suggestion
      title={blogData?.title?.text ?? ""}
      author={blogData?.author?.text ?? ""}
```

There's a better operator to use here, the [`??` nullish coalescing operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing), which only uses the value on the right-hand side if the left-hand side is null or undefined.

richardxia (approved):
LGTM! I checked this out locally to double check the pages, and they work for me.

I did notice that I had to clear my Gatsby cache to make the pages load, which feels like this annoying thing I have to do every time we update something in the code. I think this is not our fault, and perhaps something within the Gatsby or Gatsby-Prismic codebase needs to change to better handle changes to schemas.

> This PR doesn't put all the blog content on the page, nor does it style it -- I only added enough JSX for a proof of concept. I did make the decision to break the JSX out into its own component (<BlogPostTemplate />), because the query in src/pages/blog/{prismicBlogPost.uid}.tsx is very long, so I felt one or the other should live in a different file.

Yeah, I think that's a reasonable decision. This helps separate the data querying aspects from the presentational aspects.

> What do we want to do when the query fails/doesn't return anything? Should we redirect to the 404 page?

If it returns nothing at all, I think redirecting to the 404 page makes sense.

> Prismic doesn't allow us to make fields required, but there are some that we probably do want to have present (ie "title", "author"). What should we do if these return empty? Should we fail loudly, or quietly in these cases?

Prismic posted a blog post about this, but they recommend gracefully handling fields that end up blank, mostly because it's usually better to show something. Non-critical content like the author or publish date is probably OK to omit from a page, and I feel that critical content like the title would be immediately noticeable if they were missing from the page.

src/pages/blog/{prismicBlogPost.uid}.tsx
jessica-px: I keep forgetting about that, thanks! I'll swap to using `??` here in my next branch, since I'll be touching this component a lot anyway

jessica-px:
> I did notice that I had to clear my Gatsby cache to make the pages load, which feels like this annoying thing I have to do every time we update something in the code. I think this is not our fault, and perhaps something within the Gatsby or Gatsby-Prismic codebase needs to change to better handle changes to schemas.

Should we make a script that deletes the cache before building the dev environment? And then recommend that as the "default" way to run the dev enviornment? It's a tedious enough thing that I'll probably make my own function for it anyway, but since it affects the whole team, it seems reasonable to have a reusable solution of some kind...

> If it returns nothing at all, I think redirecting to the 404 page makes sense.

> Prismic posted a blog post about this, but they recommend gracefully handling fields that end up blank, mostly because it's usually better to show something.

👍  I'll toss the redirect in with my next ticket

richardxia:
> Should we make a script that deletes the cache before building the dev environment? And then recommend that as the "default" way to run the dev enviornment? It's a tedious enough thing that I'll probably make my own function for it anyway, but since it affects the whole team, it seems reasonable to have a reusable solution of some kind...

Maybe we could measure how long it takes to start the Gatsby server between having a cache and after blowing away the cache? I think we only hit this problem when changing the Prismic schemas, and even though it's a bit painful, it might be more painful if we over-pessimistically blew out the cache before starting the server each time.

At some point, I'd also like to look into the Gatsby–Prismic plugin more to see if we can figure out why this a problem in the first place

PR #297, opened 2021-11-14
Add text changes

### LINKED ISSUE #274: Update all copy on Volunteer Page

ivanleeswe:
Fix #274

### DIFF
--- a/src/pages/volunteer/index.tsx
+++ b/src/pages/volunteer/index.tsx
@@ -120,8 +120,7 @@ export default () => {
           <>
             <p>We ask that you consider these three things before joining:</p>
             <p>
-              &bull; You are willing to volunteer at least three hours each
-              week
+              &bull; You are willing to volunteer at least three hours each week
             </p>
             <p>
               &bull; You will be kind, considerate, and ethical towards other
@@ -133,7 +132,8 @@ export default () => {
               a professional and thoughftul manner.
             </p>
             <p>
-              &bull; No matter what role you play, the essential attitude you need is a deep belief in digital equity for all people
+              &bull; No matter what role you play, the essential attitude you
+              need is a deep belief in digital equity for all people
             </p>
           </>
         }
--- a/src/pages/volunteer/index.tsx
+++ b/src/pages/volunteer/index.tsx
@@ -35,8 +35,8 @@ export default () => {
         <VolunteerSignupForm />
       </Modal>
       <ImageHeader
-        title="Volunteering"
-        subtitle="We rely on hundreds of volunteers to bridge the digital divide that our most at-risk neighbors face."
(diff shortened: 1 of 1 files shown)


src/pages/volunteer/index.tsx
```diff
@@ -93,7 +93,7 @@ export default () => {
         paragraph3={{
           title: "Research & Community Development",
           description:
-            "To better understand the specific experiences people have endured with housing insecurity, our research team works alongside Community Representatives to better understand how best to serve these challenges.",
+            "No two experiences of people facing housing insecurity is the same. We have a dedicated team working alongside individuals who have experienced homelessness learning how best to serve the diverse challenges faced by those facing housing insecurity.",
```
richardxia, line 96: It looks like paragraphs 2 and 3 are swapped compared to the new page designs, since Research & Community Development should now be the 2nd paragraph.

src/pages/volunteer/index.tsx
```diff
@@ -114,24 +114,27 @@ export default () => {
...
-        paragraph1="No matter what role you play, the essential attitude you need is a deep belief in digital equity for all people."
+        title="Volunteering at ShelterTech"
+        paragraph1=""
         paragraph2={
           <>
             <p>We ask that you consider these three things before joining:</p>
             <p>
-              &ndash; You are willing to volunteer at least three hours each
-              week.
+              &bull; You are willing to volunteer at least three hours each week
```
richardxia, line 123: It looks like this was already the case before your changes here, but it'd be better to use the markup for an actual unordered list (`<ul>`) here instead of inserting a character entity for a bullet.

src/pages/volunteer/index.tsx
```diff
@@ -114,24 +114,27 @@ export default () => {
...
-              &ndash; You will be kind, considerate, and ethical towards other
-              volunteers and the communities we serve.
+              &bull; You will be kind, considerate, and ethical towards other
+              volunteers and the communities we serve
             </p>
             <p>
-              &ndash; You understand that ShelterTech is an all-volunteer team,
+              &bull; You understand that ShelterTech is an all-volunteer team,
               and that you will try your best to always represent ShelterTech in
               a professional and thoughftul manner.
```
richardxia, line 132: Nit: It looks like you removed the trailing period from every list item except for this one. If we're going to remove the period to match the designs, then we better do it for each item.

richardxia (approved):
LGTM, and thanks for updating the copy! There's a few more changes that we need to make, but we can always make them in a followup PR.

PR #299, opened 2021-12-07
Create blog post index page

### LINKED ISSUE #298: Create Blog Post index page (Prismic integration)
https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4579%3A8185

This ticket just covers the Prismic integration of the page, which needs to handle functionality such as:

- Fetching and displaying summary information about the blog posts, sorted reverse chronologically
- Filtering by category
- Pagination

richardxia:
Closes #298.

This implements the blog post index page, complete with pagination and filtering based on topic (category). This does not style the components, which I plan to do in a follow up PR.

I've attempted to keep each commit as self-contained as possible so that they may be reviewed individually.

Let me describe some of the more interesting changes I made, along with some open questions.

## Prismic custom type changes

As briefly discussed in https://github.com/ShelterTechSF/sheltertech.org/pull/287, we need the ability to associate each Blog Post with a category. In [the Figma design](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4579%3A6027), you can choose between categories in a short menu at the top, with "All Topics" being the first option. To align with that naming, I created a new Prismic type named Blog Post Topic.

At the moment, this type only contains two values: a Name, for display purposes, and a Slug, for URL purposes. I also made a small modification to the Blog Post type to add the Topic as a field.

I'm open to renaming Topic to Category or to something else, if you have any concerns about the name "Topic".

## Dynamic pages with `gatsby-node.js`

It looks like there are multiple ways of creating dynamic pages in Gatsby. I saw that @jessica-px created the dynamic pages for the individual blog posts using the [File System Routing API](https://www.gatsbyjs.com/docs/reference/routing/file-system-route-api/), where you give the page file name a weird-looking name with curly braces in them. This unfortunately doesn't seem to work well with index/list pages, where I want the URL to contain a page number.

Both the [Gatsby](https://www.gatsbyjs.com/docs/adding-pagination/) and the [Prismic+Gatsby guides](https://prismic.io/docs/technologies/paginate-results-gatsby) recommend using the `createPage()` API in gatsby-node.js, which gives you much more flexibility over this. I used this to both achieve pagination as well as topic filtering.

It's a little bit more complex than the File System Routing API, since there are GraphQL queries both in gatsby-node.js as well as in the template page component file. The goal of the gatsby-node.js GraphQL queries are to 1) get a list of all topics, and 2) figure out how many (paginated) pages we need per topic. The template component receives the topic and page number via the page context, and it performs the much larger GraphQL query to fetch every piece of data that we want to display on the page.

You might be wondering why we don't just perform all the queries in gatsby-node.js? Gatsby has some [documentation about the specific tradeoffs here](https://www.gatsbyjs.com/docs/creating-and-modifying-pages/#trade-offs-of-querying-for-all-fields-in-the-context-object-of-gatsby-nodejs), but to summarize, there are performance implications due to Gatsby caching the page context, and it may also negatively affect hot reloading. I assume that this may also affect the Prismic Preview functionality, which I did not test.

I think the last thing I wanted to mention about dynamic pages is that the GraphQL queries all declare parameters, which is something new to me. The JavaScript APIs then allow us to construct JavaScript objects and pass them in to GraphQL queries, including heavily nested JavaScript objects, which you can see me using for the topic filter.

## Location of template files

I originally placed my page template file in `src/pages/blog/BlogIndexTemplate` to match the template for the individual blog posts. Unfortunately, this causes a Gatsby build error that shows up with our `npm run build` script. I debugged this and figured out that it was caused by the parameterized GraphQL query described above. Gatsby automatically searches all of `src/pages` for pages, and the template file looks like any other page. However, when it attempts to execute that query, it fails because the query requires parameters, which are only passed in when using the gatsby-node.js `createPage()` API.

The Gatsby documentation suggests putting template files in `src/templates`, which I assume is to avoid this issue. I moved my template file there, which indeed fixed the issue. I didn't touch the blog post template, but I think we may want to move it as well for consistency.

## Testing

In Prismic, I created Blog Post Topics for each of the topics featured in the Figma design. I updated one of the blog posts to link to one of the topics, and I actually left the other one alone, just to test the behavior when someone omits a topic. I tested the pagination by temporarily changing my `POSTS_PER_PAGE` constant to 1, mostly to avoid having to create other dummy blog posts.

I did notice that attempting to visit any of the topic indexes that don't have any posts will result in a 404. This is because my code in gatsby-node.js won't actually create the page if there are no posts under that topic. I think when we launch to prod, we can just delete the topics that don't have any posts so that the links don't show up, but for now, I'll leave them in.


src/schemas/blog_post.json
```diff
@@ -1,236 +1,246 @@
 {
-  "Main" : {
```
richardxia, line 2: I think Prismic changed the format of their JSON files or something. This diff is much, much smaller if you ignore whitespace.

gatsby-node.js
```diff
@@ -0,0 +1,78 @@
+const path = require("path");
+
+/** Create paginated blog post index pages, filtered by topic.
+ *
+ * `topic` should be a BlogPostTopic uid, or null for the case where we list all
+ * posts regardless of topic.
```
jessica-px, line 6:
```suggestion
 * posts regardless of topic.
 *
 * See: https://prismic.io/docs/technologies/paginate-results-gatsby
```

jessica-px (approved):
Thanks for the thorough PR walkthrough -- looks great! I tested it locally and everything works as described.

I'm fine with naming it `topic`, and I can move the blog post template into the templates dir in my current branch, after this merges in 👍

gatsby-node.js
jessica-px: For framework-dependent boilerplate heavy stuff like this, I've taken a liking to adding links to the docs -- although I do notice this url doesn't have a version parameter in it...hopefully it doesn't break in the future

richardxia: Good idea, let me add in the suggestion. Even if the URL breaks, we generally should be sticking close to the latest versions of Gatsby and the Gatsby-Prismic plugin, so hopefully our code will still up to date with the documentation.

PR #301, opened 2021-12-22
Add blog post page jsx

jessica-px:
## Description

This PR adds the JSX necessary to build the blog post pages, and passes the data from our Prismic query to it. As a result, blog pages are now minimally functional, but completely unstyled. Styling will be done in my next PR.

Some notes on some decisions I made/open questions:
- For each slice/component block (ie `TextBlock` or `QuoteBlock`) I made a new component in `components/blog`. These are all pretty light weight components, but I think we'll see more organizational payoff once styles have been added.
- In `BlogPostTemplate` I had to assign the `slices` prop an `any[]` type, because although Gatsby generates a type for the whole return value of our query, it doesn't generate discrete named types for the nested values -- in this case, the slices. That makes accessing the actual types very difficult. It's potentially something we could solve, but I decided not to put too much time into it for now, even though the `any`, and the accompanying disabled eslint rules, irritate me.
- I also moved `BlogPostTemplate` to `src/templates` for consistency
- I disabled `react/no-unused-prop-types` in `ImageBlock` for now, because I'm not using the `isFullWidth` prop yet. But I intend to use it for adding styles.
- I realized belatedly that I'm still not redirecting to a 404 page when the query doesn't return any data. But since that's a visual/UX thing, I think it's reasonable for that to happen in the styling PR

- I'm not sure if I should add storybook entries for these components?

## Testing

I tested this by viewing `http://localhost:8000/blog/cool-blog-post/`, which has at least one entry for every possible blog post slice. It looks like garbage, because it's unstyled, but everything renders with the correct html.


src/components/blog/ImageBlock/ImageBlock.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+
+type ImageBlockProps = {
+  caption: string;
+  url: string;
+  /* eslint-disable react/no-unused-prop-types */
```
richardxia: Nit: Not super important to fix in this PR, especially since this will go away shortly when the components are styled, but in general when disabling ESLint rules, can we scope the disabling to as narrow of a scope as possible? For example, you could use `eslint-disable-next-line` or `eslint-disable-line`. This way, we won't unintentionally disable the rule for the rest of the file.

src/templates/BlogPostTemplate/blogPostTemplate.tsx
```diff
@@ -0,0 +1,117 @@
...
+  topic,
+  title,
+  author,
+  date,
+  headerImgUrl,
+  headerImgAlt,
+  slices,
+}: BlogPostTemplateProps) => {
+  // returns either date or author, or if both are given "date - author"
+  const getDateAuthorString = (): string | null => {
```
richardxia: I'm curious as to why you chose to make this a function with no parameters that gets called in exactly one place as opposed to just assigning the value directly? The logic would be exactly the same except you would assign to a variable instead of returning. I think directly assigning a variable would be a bit more straightforward and remove a layer of indirection.

src/templates/BlogPostTemplate/blogPostTemplate.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+    if (date) {
+      const formattedDate = new Date(date).toLocaleDateString("en-US", {
+        year: "numeric",
+        month: "long",
+        day: "numeric",
+      });
+      return author ? `${formattedDate} - ${author}` : formattedDate;
+    }
+    return author || null;
```
richardxia: Independent of my question about function vs directly assigning a variable, here's a suggestion for restructuring this to make the control flow a bit more symmetric and table-like:

```js
const formattedDate = date && new Date(date).toLocaleDateString(...); // This now has type `string | null`
if (formattedDate && author) {
  return `${formattedDate} - ${author}`;
} else if (formattedDate && !author) {
  return formattedDate;
} else if (!formattedDate && author) {
  return author;
} else {
  return null;
}
```

src/components/blog/StatsBlock/StatsBlock.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+
+type StatsBlockProps = {
+  statistic: string;
+  statistic_text: string;
```
richardxia: Nit: Use camelCase for prop names. Even though the Prismic type names use snake_case, we're manually passing the prop into this component anyway, so there's no additional code required to change the name.

src/templates/BlogPostTemplate/blogPostTemplate.tsx
```diff
@@ -0,0 +1,117 @@
...
+              return (
+                <ButtonBlock
+                  header={slice.primary.header.text}
+                  url={slice.primary.button_link.url}
+                  buttonText={slice.primary.button_text.text}
+                />
+              );
+            case "separator":
+              return (
+                <img src={LogoSeparator} alt="Divider with ShelterTech logo" />
```
richardxia: It's generally not useful to add alt text to presentation-only images like this, since it's not helpful to people using screen readers. The [W3C official recommendation](https://www.w3.org/QA/Tips/altAttribute) is to set `alt=""` for images like this:

> if the image is used to create bullets in a list, a horizontal line, or other similar decoration, it is fine to have an empty alt attribute (e.g., alt=""), but it is better to use things like list-style-image in CSS

To provide a motivating example, imagine if you sent a blog post through a text-to-speech program, and between paragraphs, it suddenly spoke "Divider with ShelterTech logo."

richardxia (approved):
LGTM, aside from some mostly minor comments, and sorry for the belated review!

> In BlogPostTemplate I had to assign the slices prop an any[] type,... It's potentially something we could solve, but I decided not to put too much time into it for now ...

Yeah, this sounds fine for now. I'm curious if upgrading any of the libraries we're using would help, but we should wait until after we launch the Blog before trying that. Another idea would be to manually copy over the type definitions into `blogPostTemplate.tsx`. This is less bad than it sounds, since the types in `blogPostTemplate.tsx` really describe what the code in that file expects, and if we changed the Prismic types, we'd get a type error due to a mismatch between where the data being queried from Prismic is passed into the Template component.

> I'm not sure if I should add storybook entries for these components?

I think it'd be useful to do so, but maybe only after we actually build styles for them. Although they are extra work, I think it'll pay off in the future when we, as developers, can't just enter in dummy data into Prismic to test the appearance of the page, and being able to test style changes purely in Storybook would be very valuable.

src/templates/BlogPostTemplate/blogPostTemplate.tsx
richardxia: So I ended up implementing this myself in my local branch for styling the Blog Index page, and I ended up with this, much more concise version:

```js
const formattedDate = date && new Date(date).toLocaleDateString(...); // This now has type `string | null`
return [formattedDate, author].filter((x) => x).join(" - ");
```

The `.filter()` gets rid of any null or empty strings, and the only difference between this and your version or my older version is that this new one will return an empty string instead of null, but I think it shouldn't matter because empty strings also count as `false`.

PR #302, opened 2022-01-04
TwoParagraphBlock: Update styling for Volunteer page variant.

### LINKED ISSUE #271: Update TwoParagraphBlock

Screen Shot 2021-07-24 at 11 03 00 AM

As weird as it might sound, the new design for the Volunteer Page includes a TwoParagraphBlock with only one "paragraph", where the paragraph/subtitle on the top right is completely removed, and the image shifted up to take its place.

We should update the component to use this layout when the top right paragraph is omitted from the props.

In addition, we should add a white background version of the component.

@derekfidler, let me know if we should consider this to just be a completely different component than the TwoParagraphBlock or whether we should rename the TwoParagraphBlock to better fit its new roles.

richardxia:
Fixes #271.

The changes include:

- Allow for a white background (via theme prop)
- Making the image span both rows for the Volunteer page variant
- Eliminating the "first paragraph" `<div>` if no paragraph1 is passed in
- Removing unnecessary differences between the Volunteer page and About
  Us page variants of the components

There's still a bit of oddness due to differences in the design between
the About Us and the Volunteer pages, but this isn't adding any further
discrepancies.

Since the CSS for this component is pretty tricky and there are significant differences in styling between the About Us page and Volunteer page versions, I manually checked both pages on both laptop and mobile sized screens, just to make sure there weren't any issues. Here are a few screenshots of what this now looks like:

### About Us
Screen Shot 2022-01-03 at 9 15 19 PM

### Volunteer
Screen Shot 2022-01-03 at 9 21 25 PM

### DIFF
--- a/src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
+++ b/src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
@@ -13,9 +13,12 @@
 }
 
 .bleedBackground {
-  background: var(--color-gray-300);
   grid-column: 1 / span 2;
   grid-row: 1;
+}
+
+.bleedBackground.gray {
+  background: var(--color-gray-300);
 }
 
 .bleedMainContent {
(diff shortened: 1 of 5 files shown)


davidagustin (approved):

src/components/grid-aware/TwoParagraphBlock/TwoParagraphBlock.module.css
```diff
@@ -139,28 +121,24 @@
...
   margin-top: 15px;
 }
 
-.paragraph2WrapperAbout :not(:first-child) {
-  margin-top: 15px;
-}
-
-.image1Wrapper {
-  margin-bottom: -142px;
+@media (--laptop-and-up) {
```
davidagustin, line 136: Ahh I see, very clever here actually.

richardxia:
Thanks!

PR #303, opened 2022-01-04
Home Page: Remove black background and spacer.

richardxia:
Just a real quick change, which is I believe the last thing needed to get the site up to parity with the new Figma designs (sans Impact page). Both the spacer and the video spotlight component fortunately have props to control the background color, so I didn't need to change the components themselves, just the Home page.

Screen Shot 2022-01-03 at 9 26 04 PM

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -134,18 +134,13 @@ export default () => {
         ]}
         theme="white"
       />
-      <Spacer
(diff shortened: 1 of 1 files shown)


davidagustin (approved):
Very simple change. Looks good to me.

richardxia:
Thanks for the review!

PR #304, opened 2022-01-04
Adapt stats block for impact page

### LINKED ISSUE #294: Create new or adapt Stats Block for Impact page
https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4329%3A6859

Screen Shot 2021-10-31 at 9 52 45 AM

The new Impact page has several instances of a component that looks similar to the Stats Block that appears on the Volunteer Page. The scope of this ticket is to either adapt the existing component or to implement a brand new block to support its appearance on the Impact Page.

candywang:
Update stats block to include optional subtitle and YOY number. Closes  #294.

### DIFF
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -1,4 +1,3 @@
-
 .gray {
   background-color: var(--color-gray-300);
 }
@@ -74,15 +73,14 @@
   margin: 0;
 }
 
-.impactTitle {
+.blue {
   color: var(--color-sheltertech-blue);
 }
 
 .subtitle {
   font: var(--font-body-medium);
-  grid-column: 2 / span 10;
-  margin: 20px 0 0;
-  overflow-wrap: break-word;
+  grid-column: 2 / span 8;
+  margin-top: 20px;
 }
 
 .statCard {
@@ -113,10 +111,14 @@
   line-height: 1.3;
 }
 
-.yoy {
+.secondaryNumber {
   color: var(--color-gray-700);
   font: var(--font-body-small);
-  margin: 5px 0 10px;
+  margin-top: 5px;
+}
+
+.secondaryNumber + .statement {
+  margin-top: 15px;
 }
 
 .statement {
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -113,6 +113,12 @@
   line-height: 1.3;
 }
 
+.yoy {
+  color: var(--color-gray-700);
+  font: var(--font-body-small);
+  margin: 5px 0 10px;
+}
+
 .statement {
   font: var(--font-body-medium);
   margin-top: 5px;
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -1,3 +1,4 @@
+
 .gray {
   background-color: var(--color-gray-300);
 }
@@ -55,7 +56,7 @@
   display: flex;
   flex-direction: row;
   grid-column: 2 / span 10;
-  grid-row: 2;
+  grid-row: 3;
   margin-top: 54px;
 }
 
@@ -71,6 +72,17 @@
   font: var(--font-title);
   grid-column: 2 / span 8;
   margin: 0;
+}
+
+.impactTitle {
+  color: var(--color-sheltertech-blue);
+}
+
+.subtitle {
+  font: var(--font-body-medium);
+  grid-column: 2 / span 10;
+  margin: 20px 0 0;
+  overflow-wrap: break-word;
 }
 
 .statCard {
--- a/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
@@ -3,7 +3,7 @@ import StatsBlock from "./StatsBlock";
 
 export default {
   title: "Grid-Aware/StatsBlock",
-  subTitle: "Optional Subtitle",
+  subtitle: "Optional Subtitle",
   component: StatsBlock,
 };
 
@@ -19,22 +19,23 @@ DefaultStatsBlock.args = {
   statCards: [
     {
       number: "9,000",
-      yoy: 5,
+      secondaryNumber: "+5 YOY",
       statement:
         "At-risk community members can now access services and amenities",
     },
     {
       number: "560",
-      yoy: 5,
+      secondaryNumber: "+5 YOY",
       statement:
         "Bay Area volunteers have contributed to our cause since our founding in 2016",
     },
     {
       number: "$2M",
-      yoy: 5,
+      secondaryNumber: "+5 YOY",
       statement:
         "Worth of work provided by volunteers has been put towards our programs",
     },
   ],
   theme: "white",
+  titleColor: "blue",
 };
--- a/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
@@ -19,16 +19,19 @@ DefaultStatsBlock.args = {
   statCards: [
     {
       number: "9,000",
+      yoy: 5,
       statement:
         "At-risk community members can now access services and amenities",
     },
     {
       number: "560",
+      yoy: 5,
       statement:
         "Bay Area volunteers have contributed to our cause since our founding in 2016",
     },
     {
       number: "$2M",
+      yoy: 5,
       statement:
         "Worth of work provided by volunteers has been put towards our programs",
     },
--- a/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
@@ -3,17 +3,19 @@ import StatsBlock from "./StatsBlock";
 
 export default {
   title: "Grid-Aware/StatsBlock",
+  subTitle: "Optional Subtitle",
   component: StatsBlock,
 };
 
-const Template = ({ title, statCards }) => (
-  <StatsBlock title={title} statCards={statCards} />
+const Template = ({ title, subtitle, statCards }) => (
+  <StatsBlock title={title} subtitle={subtitle} statCards={statCards} />
 );
 
 export const DefaultStatsBlock = Template.bind({});
 
 DefaultStatsBlock.args = {
   title: "See what our volunteers have accomplished together.",
+  subtitle: "Optional subtitle.",
   statCards: [
     {
       number: "9,000",
--- a/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
@@ -7,12 +7,13 @@ export default {
   component: StatsBlock,
 };
 
-const Template = ({ title, subtitle, statCards }) => (
+const Template = ({ title, subtitle, statCards, theme, titleColor }) => (
   <StatsBlock
     title={title}
     subtitle={subtitle}
     statCards={statCards}
-    titleColor="blue"
+    theme={theme}
+    titleColor={titleColor}
   />
 );
 
--- a/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
@@ -8,7 +8,12 @@ export default {
 };
 
 const Template = ({ title, subtitle, statCards }) => (
-  <StatsBlock title={title} subtitle={subtitle} statCards={statCards} />
+  <StatsBlock
+    title={title}
+    subtitle={subtitle}
+    statCards={statCards}
+    titleColor="blue"
+  />
 );
 
 export const DefaultStatsBlock = Template.bind({});
--- a/src/components/grid-aware/StatsBlock/StatsBlock.tsx
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.tsx
@@ -53,8 +53,13 @@ const StatsBlock = ({
         </h1>
         {subtitle && <h2 className={s.subtitle}>{subtitle}</h2>}
         <div className={s.gridAreaBottom}>
-          {statCards.map(({ number, statement }) => (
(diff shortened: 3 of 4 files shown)


src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -5,23 +5,26 @@ import * as s from "./StatsBlock.module.css";
...
   number: string;
+  yoy?: number;
   statement: string;
 };
 
-const StatCard = ({ number, statement }: StatCardProps) => (
+const StatCard = ({ number, yoy, statement }: StatCardProps) => (
   <div className={s.statCard}>
     <div className={s.number}>{number}</div>
+    {yoy && <div className={s.yoy}>+{yoy} YOY</div>}
```
richardxia: Even though the only usages of this new field on the Impact page show the year-over-year change, I think it'd be better if we made this more generic and allow any arbitrary text. This would allow us to reuse the visual style of having a light gray text below the main stat without it always being a YOY metric.

Could you rename the field to something more generic like `secondaryNumber`, and change it to be a string? Also, in the Storybook stories, could you update the value to be the full string `"+5 YOY"` instead of just the number `5`?

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -32,7 +35,10 @@ const StatsBlock = ({ title, statCards, theme }: StatsBlockProps) => (
...
     <div className={s.bleedMainContent}>
       <section className={s.gridParent}>
-        <h1 className={s.title}>{title}</h1>
+        <h1 className={`${s.title} ${subtitle ? s.impactTitle : ""}`}>
+          {title}
+        </h1>
+        {subtitle && <h2 className={s.subtitle}>{subtitle}</h2>}
         <div className={s.gridAreaBottom}>
           {statCards.map(({ number, statement }) => (
             <StatCard key={statement} number={number} statement={statement} />
```
richardxia: It looks like you forgot to pass in the new `yoy` (or `secondaryNumber`) prop into the individual `StatCard` components here.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -1,3 +1,4 @@
+
```
richardxia: Nit: Please avoid inserting unnecessary blank lines in existing code.

```suggestion
```

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -73,6 +74,17 @@
   margin: 0;
 }
 
+.impactTitle {
```
richardxia: I'd like to avoid any page-specific variants of components, especially when they leak into the names of CSS styles. Even if there are differences between the components on the different pages, it's usually better to express that variability directly in the component. For example, for this component, we could name this style something more generic like `.blueTitle` or even `.blue`, which would remove any mention of "impact" from the CSS class name. This would allow us to have greater flexibility in the future if we add more pages which want the blue title variant of the StatsBlock.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -73,6 +74,17 @@
...
+  color: var(--color-sheltertech-blue);
+}
+
+.subtitle {
+  font: var(--font-body-medium);
+  grid-column: 2 / span 10;
```
richardxia: Looking at the Figma designs, I think this only spans 8 columns, not ten, since when you turn on the layout grid in Figma, you can see that the subtitles wrap at the 9th column.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -73,6 +74,17 @@
...
+  margin: 20px 0 0;
```
richardxia: When you're just setting one margin direction, could you use the individual CSS property (e.g. `margin-top`)? This way, you won't unnecessarily have to specify the margins for the other directions.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -73,6 +74,17 @@
...
+  overflow-wrap: break-word;
```
richardxia: Out of curiosity, what led you to setting this property? It looks like the only difference between this and the default setting of `normal` is that this will allow line breaks to be inserted in the middle of words. The [examples on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) only really demonstrate this on very narrow components, but I think that's unlikely to occur for the subtitle of the StatsBlock component, which is generally quite wide.

Although I agree that `break-word` is the more sensible default value for most text, my general approach with CSS is to set the minimum number of properties as possible to avoid adding extra complexity to the page.

src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
```diff
@@ -3,30 +3,35 @@ import StatsBlock from "./StatsBlock";
 
 export default {
   title: "Grid-Aware/StatsBlock",
+  subTitle: "Optional Subtitle",
```
richardxia: Nit: Could you make the "t" in "subtitle" lowercase? This would make it match everywhere else the prop name is spelled.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -101,6 +113,12 @@
   line-height: 1.3;
 }
 
+.yoy {
+  color: var(--color-gray-700);
+  font: var(--font-body-small);
+  margin: 5px 0 10px;
```
richardxia: There's a subtle little bug here due to a somewhat obscure rule in CSS regarding margins. CSS has this rule called "collapsing margins" where in certain cases, adjacent margins of adjacent elements may _overlap_ with each other instead of pushing each other apart. Here is a pretty great article with illustrations that explain how margin collapse works: https://www.joshwcomeau.com/css/rules-of-margin-collapse/

For this specific instance, the Figma designs show that when this extra YOY text is present, there should be a 15px gap between the YOY text and the other text below it (the `.statement` element). I think you placed a 10px bottom margin here because there is already a 5px top margin on the other text. However, due to the collapsing margin rule, the 10px margin ends up _overlapping_ with the 5px margin, which means that there ends up only being a 10px space between the text.

There's a few different ways you could avoid this. You can change the bottom margin on the `.yoy` style here to be the full 15px, which will overlap with the other margin and still guarantee that the other text is a full 15px away. If you fix it this way, please add a comment explaining that the margins are being collapsed, since it's difficult to notice without looking at the full HTML + CSS.

Another way to deal with this is to use an [adjacent sibling selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator) to add a custom style to the `.statement` element _only_ when it is preceded by the `.yoy` element. This would look like this:

```css
.yoy + .statement {
  margin-top: 15px;
}
```

The adjacent sibling selector only applies to the latter sibling, so this guarantees that the .statement element has a top margin of 15px _only_ when the .yoy element is just before it.

richardxia (commented):
Thanks for working on this! It looks mostly good to me, and most of my comments are just about improvements to naming and code style, so I think we can get this merged soon after you address them.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -32,7 +35,10 @@ const StatsBlock = ({ title, statCards, theme }: StatsBlockProps) => (
     />
     <div className={s.bleedMainContent}>
       <section className={s.gridParent}>
-        <h1 className={s.title}>{title}</h1>
+        <h1 className={`${s.title} ${subtitle ? s.impactTitle : ""}`}>
```
richardxia: Related to my comment about renaming the `.impactTitle` CSS class to something generic like `.blue`, could you make the conditional expression dependent on an explicit prop on the StatsBlock, similar to our `theme` props? I think it's better to be explicit here about which color we want the title to be than to implicitly decide on it based on the presence or absence of a subtitle.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
candywang: Sorry! Did not notice that. Would there be a way for the linter to check for blank lines at the beginning of a document?

src/components/grid-aware/StatsBlock/StatsBlock.module.css
candywang: I tested the subtitle with a really long word and added the `overflow-wrap` property since the text did not wrap and would leak out of the container but I can take this out since, as you mentioned, it's unlikely to occur.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
candywang: Thanks for helping me understand the margin collapse and adjacent sibling selector! 🤩

src/types.ts
```diff
@@ -1 +1,4 @@
-export type ThemeColorOption = "white" | "gray";
+type ThemeColorOption = "white" | "gray" | "blue";
```
richardxia: Currently, this type really just covers backgrounds, and I don't believe we have any instance of a component with a blue background option. Is this something you were just trying out before creating the `TitleColorOption` type? If so, and if the `blue` option isn't being used anywhere, could you remove the `blue` option from this list?

src/types.ts
```diff
@@ -1 +1,4 @@
...
+type TitleColorOption = "blue";
```
richardxia: It's a little bit strange to define a new option type like this with only one option, especially when this is not actually the actual default color that we use on most components. Could you add `black` as an option here so that it's clearer what the normal set of options are?

src/types.ts
```diff
@@ -1 +1,4 @@
...
+
+export { ThemeColorOption, TitleColorOption };
```
richardxia: Nit: I'd personally prefer to leave the `export`s on the lines that initially define the values, especially in cases like this where we're exporting every named thing in a file. In certain cases, it may be useful to keep all the exports in one place for describing what the "public" API of a file is in one place, but for files that are relatively short or that are relatively transparent, it causes more indirection for a reader to separate `export`s from the initial declarations.

I acknowledge that this is really just a matter of style and that we don't have any explicit guidance about this written down anywhere, so I'm not adamant about moving the `export`s to the declarations.

cc @jessica-px, in case she has any opinions about this.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -1,27 +1,39 @@
...
+  titleColor?: TitleColorOption;
 };
 
-const StatsBlock = ({ title, statCards, theme }: StatsBlockProps) => (
+const StatsBlock = ({
+  title,
+  subtitle,
+  statCards,
+  theme,
+  titleColor,
```
richardxia: If you add `black` to the title color options, you could even set that to be the default value here so that you don't have to check whether `titleColor` has been set later, as it will always have some value.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -32,10 +44,22 @@ const StatsBlock = ({ title, statCards, theme }: StatsBlockProps) => (
     />
     <div className={s.bleedMainContent}>
       <section className={s.gridParent}>
-        <h1 className={s.title}>{title}</h1>
+        <h1
+          className={`${s.title} ${
+            subtitle && titleColor ? s[titleColor] : ""
```
richardxia: You can remove the `subtitle` check here. Since we've made `titleColor` its own property, it would make the most sense to someone using the `StatsBlock` component that `titleColor` will control the color of the title regardless of the presence of a subtitle. Removing the unnecessary complexity makes it more direct and easier to use.

src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
```diff
@@ -3,33 +3,39 @@ import StatsBlock from "./StatsBlock";
...
 export default {
   title: "Grid-Aware/StatsBlock",
+  subtitle: "Optional Subtitle",
   component: StatsBlock,
 };
 
-const Template = ({ title, statCards }) => (
-  <StatsBlock title={title} statCards={statCards} />
+const Template = ({ title, subtitle, statCards }) => (
+  <StatsBlock title={title} subtitle={subtitle} statCards={statCards} />
```
richardxia: Since you added `titleColor` as a prop to the component and as an `arg` to the Storybook story, you'll also need to add it here. Otherwise, when you view the Storybook story, the title is still black despite the fact that the controls at the bottom say "blue".

richardxia (approved):
This mostly LGTM now, thanks for addressing the previous set of comments I had! I think this is pretty much good to go once the last set of things I pointed out are addressed, which are all pretty minor. Thanks again for working on this!

src/components/grid-aware/StatsBlock/StatsBlock.module.css
richardxia: Good call, it looks like there is a Stylelint rule for it: https://stylelint.io/user-guide/rules/list/no-empty-first-line/

I created a ticket to track this here: https://github.com/ShelterTechSF/sheltertech.org/issues/308

src/components/grid-aware/StatsBlock/StatsBlock.tsx
candywang: Do we plan to implement a class selector for `black`?

candywang: Please disregard my last question.

src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
```diff
@@ -3,33 +3,44 @@ import StatsBlock from "./StatsBlock";
...
 };
 
-const Template = ({ title, statCards }) => (
-  <StatsBlock title={title} statCards={statCards} />
+const Template = ({ title, subtitle, statCards }) => (
+  <StatsBlock
+    title={title}
+    subtitle={subtitle}
+    statCards={statCards}
+    titleColor="blue"
```
richardxia: Ah, this needs to be a prop on the `Template`, like the other props of StatsBlock. Otherwise, when you're in Storybook, attempting to change the title color in the controls section at the bottom of the page won't do anything.

Screen Shot 2022-02-01 at 7 37 23 PM

richardxia (approved):
Last comment, but this should otherwise be good to merge! Thanks again for working on this!

PR #311, opened 2022-02-08
Add meta tags for blog posts.

### LINKED ISSUE #310: Hook up page metadata for blog posts
We'll want post-specific information to appear in the page metadata, such as the blog post title, since this would be better for sharing the links on social media as well as for SEO purposes.

We already use react-helmet on the site, so this would just be adding blog-post-specific info to it.

richardxia:
Closes #310.

These are basic tags that mostly follow the guidance from
https://moz.com/blog/meta-data-templates-123

Note that this currently leaves off the `<meta name="description">` and
`<meta property="og:description">` tags. It seems that best practices for
SEO recommend writing a distinct description that's not automatically
generated from the page content itself. This is probably best done as a
separate field in Prismic, which we can do later once the page styling
is done.

It's hard to test this without actually deploying this somewhere real, where I can attempt to link to a page from an actual social media app, but I eyeballed the generated HTML tags and they look about right to me:
Screen Shot 2022-02-07 at 9 46 08 PM

Note that `twitter:site` won't show up until we wrap the whole page with `<Layout>`, since I defined it site-wide in that component, but I think that's going to happen very soon when you style the page, @jessica-px.

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -52,6 +52,7 @@ const Layout = ({ children }: LayoutProps) => {
           name="description"
           content="ShelterTech is solving the biggest technology challenges faced by those experiencing homelessness"
         />
+        <meta name="twitter:site" content="@sheltertechorg" />
         <link rel="icon" href={favicon} />
         {/* Global site tag (gtag.js) - Google Analytics */}
         <script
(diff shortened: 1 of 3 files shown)


jessica-px (approved):
Output seems alright to me 🤷‍♀️

richardxia:
Thanks! Just rebased this to fix up a few merge conflicts, and will merge it now.

PR #312, opened 2022-02-08
Style blog pages

jessica-px:
This PR styles the blog post page template as shown in [this mockup](https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4579%3A6027).

## Highlights
- I think this is my first time really doing CSS in this codebase? I'm not sure if I'm matching conventions right. At first I didn't like assigning every component the same width and margins over and over, and wanted to just shove them all into a consistent container, but then we do have some blog post components that deviate from those measurements, so I guess this method worked out.
- The mockup doesn't have a mobile version. It has a couple of notes for mobile, but aside from that I was able to mostly leave things alone and have it still look pretty good. I'm assuming design will have a chance to look at this before it goes live, but the only two shots I (knowingly?) called were:
  - adding left/right padding to text/quote/stats blocks at small screen sizes, so they didn't squish up against the edges
  - letting the CTA block be full width at small screen sizes, to match how the mockup said to handle images
- I moved `LogoSeparator` and the thing I'm calling `TitleBlock` out of `BlogPostTemplate` and into their own components
- Forgot to call it out in my git commit, but if you're wondering what changed in the `LogoSeparator.svg` diff, I shaved the whitespace off the edges so it could resize itself as expected
- Also I was able to remove the `isFullWidth` prop that I thought I was going to need for the `ImageBlock`. Reading the mockup more closely, I'd misunderstood how we were going to be handling widths.

## Screenshots
(These have been zoomed out very very far, because I wanted to show some Big Picture previews in this PR description, but if you want to actually see what it looks like, the most realistic option is to check out the branch. These pages have a lot of scrolling!

Screen Shot 2022-02-07 at 11 09 19 PM
Screen Shot 2022-02-07 at 11 09 46 PM
Screen Shot 2022-02-07 at 11 09 56 PM


src/components/blog/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,25 @@
+.bleedWrapper {
+  margin: 100px auto 0;
```
richardxia: I generally try to avoid having margins on the outside of the top-level element of a reusable component, since that often makes it more difficult to reuse in other contexts. You can always easily add margin around a component, but it's much more difficult to remove existing margin. The `auto` side margins are fine here, since they're just for centering, but the top-margin is the one that has potential to be problematic.

Although it may be controversial, I think when assembling components into a page, using a spacer component at the page-level is acceptable. We have one available at [src/components/grid-aware/Spacer/Spacer.tsx](https://github.com/ShelterTechSF/sheltertech.org/blob/3ee96124bdea6df3c9347db429ddd7c63e8b3f3a/src/components/grid-aware/Spacer/Spacer.tsx), which you should be able to drop onto the blog post template.

src/components/blog/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,25 @@
...
+  max-width: 630px;
```
richardxia: Regarding your comment:

> At first I didn't like assigning every component the same width and margins over and over, and wanted to just shove them all into a consistent container, but then we do have some blog post components that deviate from those measurements, so I guess this method worked out.

Since this number comes up in most of the components on the blog post page, we could extract this into a global CSS custom property (i.e. variable) in [src/stylesheets/global.css](https://github.com/ShelterTechSF/sheltertech.org/blob/3ee96124bdea6df3c9347db429ddd7c63e8b3f3a/src/stylesheets/global.css). This would mitigate the issue with hardcoding the same number across many separate CSS files while still granting the flexibility of not wrapping the whole page in a single fixed-width element, which, as you pointed out, would cause issues with components that need to expand past those boundaries.

src/components/blog/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,25 @@
+.bleedWrapper {
```
richardxia: "Bleed wrapper" might not be the best name for this or most of the other components on the blog page. I realized that I haven't documented the rationale behind this naming anywhere, but the term [bleed](https://en.wikipedia.org/wiki/Bleed_(printing)) here is borrowed from printing, and it specifically refers to the region of a page that is printed with color or images right up to the edge of the paper. ([Here's a website that shows a few examples](https://cwcreative.com/blog/faq-what-is-bleed).) In the regular, non-blog pages on the site, we have several components where the background bleeds all the way to the left and/or right sides of the page. For example, see the "A note from our Founder, Darcel Jackson" block on the [About Us page](https://www.sheltertech.org/about).

The blog does not really have this same kind of design, since it's primarily focused on long-form prose with a relatively narrow width. Even the wide image blocks don't "bleed" into the edges of the pages like the full-bleed backgrounds of the non-blog components.

Instead, I think we can just drop the word "bleed" from the class name and call it `.wrapper`, or, if you don't like that, something more generic like `.container`.

I can file a ticket for myself to better document this "bleed" system in our [design system docs](https://github.com/ShelterTechSF/sheltertech.org/blob/main/docs/component-design-system.md).

src/components/blog/TitleBlock/TitleBlock.module.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.topic {
+  color: var(--color-gray-700);
+  font: var(--font-subtitle);
+}
+
+.title {
+  font: var(--font-headline);
+  margin: 20px 0;
```
richardxia: I've enjoyed using the `gap`[^1] property in place of margins on the individual components since it applies generically to all grid or flexbox child elements, and it automatically handles cases when certain child elements are missing, since spacing is only placed between rendered elements. For this component, it would specifically handle the case where the topic is missing by not inserting a margin above the title, and it'd also handle the case where the title is missing by ensuring there is still a 20px gap between the top and the date/author row.

To use this, you'd need to turn the parent element (`.bleedWrapper`) into a grid or flex parent, but once you do so, you can use `row-gap: 20px` to set a 20-pixel gap between all grid rows.

Obviously, this doesn't work in cases where you may want a different about of spacing between elements, but the redlines (well, yellow bubbles) in the Figma design indicate that there should be "20px between items in the title block", so it seems logical to me to apply this generically to the wrapper element.

[^1]: Formerly named `grid-gap`, but recently renamed just `gap` because it also works on flex items now.

src/components/blog/ImageBlock/ImageBlock.module.css
```diff
@@ -0,0 +1,17 @@
+.image {
+  display: block;
```
richardxia: What's the rationale behind setting the image's display style to `block`? I'm worried that this may have some unintended consequences, since images are normally `inline` content with some [special rules](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#styling_with_css) due to it being "replaced content".

src/components/blog/ImageBlock/ImageBlock.module.css
```diff
@@ -0,0 +1,17 @@
...
+  margin: 50px auto;
```
richardxia: I think this is causing extra spacing between the image and its caption, which, according to the Figma mockup, should be just 20px. Did you mean to apply this to the nameless wrapper `<div>` that contains both the image and the caption? That said, I think my earlier comment about not having external margin on the top-level elements of reusable components applies here as well, so we may want the vertical components of this line to go away altogether.

src/components/blog/ImageBlock/ImageBlock.tsx
```diff
@@ -1,16 +1,15 @@
...
-  /* eslint-disable-next-line react/no-unused-prop-types */
-  isFullWidth?: boolean;
 };
 
 const ImageBlock = ({ url, caption }: ImageBlockProps) => (
   <div>
-    <img src={url} alt={caption} />
-    <p>{caption}</p>
+    <img className={s.image} src={url} alt={caption} />
+    <p className={s.caption}>{caption}</p>
```
richardxia: I think `<p>` isn't exactly the right tag to use here, since it should just be used for paragraph content. You could go with the [`<figcaption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) tag, which would require you to change the wrapper `<div>` into a [`<figure>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure), although I would also be fine with just changing this `<p>` into a plain `<div>`.

src/components/blog/ImageBlock/ImageBlock.stories.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import ImageBlock from "./ImageBlock";
+import sampleImage from "./stories/sampleImage.png";
+
+export default {
+  title: "Blog/ImageBlock",
+  component: ImageBlock,
+};
+
+const Template = ({ caption, url, isFullWidth }) => (
```
richardxia: I think you probably meant to remove the `isFullWidth` prop from this file, since it's no longer one of the actual component's props.

src/components/blog/TextBlock/TextBlock.module.css
```diff
@@ -0,0 +1,44 @@
+.bleedWrapper {
+  margin: auto;
+  max-width: 630px;
+}
+
+.richText h1,
+h2,
+h3,
+h4,
+h5 {
```
richardxia:
```suggestion
.richText h1,
.richText h2,
.richText h3,
.richText h4,
.richText h5 {
```

You need to prefix each of these with `.richText`, since the comma has lower precedence than all other operators in CSS selectors. Without the additional `.richText` prefixes, this actually ends up applying to _all_ `<h2>`, `<h3>`, etc. elements, even those outside of the `.richText` area.

Same comment goes for all the other selectors in this file that don't have `.richText` before the tag selector.

src/components/blog/TextBlock/TextBlock.module.css
```diff
@@ -0,0 +1,44 @@
...
+  font: var(--font-subtitle);
+  margin: 50px 0 20px;
```
richardxia, line 14: I noticed there's a discrepancy between the Figma design and the yellow bubble comment to the side, where the bubble says 20px but the actual design seems to have 40px of spacing below headings. I've added a comment in the Figma design asking Derek to clarify, but I'm leaving a comment here in the PR just so that I don't forget.

src/components/blog/TextBlock/TextBlock.module.css
```diff
@@ -0,0 +1,44 @@
...
+}
+
+.richText p,
+ul,
+ol {
+  font: var(--font-body-medium);
+  margin-top: 20px;
```
richardxia: The Figma design says 10px, but the design appears to actually have 26px (or 1em at this font size) of spacing between paragraphs and lists. I've asked Derek to clarify, but I'm once again leaving a comment here just as a reminder.

src/components/blog/TextBlock/TextBlock.stories.jsx
```diff
@@ -0,0 +1,61 @@
...
+        {
+          start: 29,
+          end: 38,
+          type: "strong",
+        },
+      ],
+    },
+    {
+      type: "o-list-item",
+      text: "This list item has italicized text",
```
richardxia, line 41: This is not a problem on this line, but there wasn't a better place for me to leave this comment. It looks like we're missing the correct font for italicized text, but it's a bit subtle because we're accidentally picking up a thinner style of the same italic font from the legacy site.

When rendered, the italic text looks thinner than the surrounding non-italic text:
Screen Shot 2022-02-13 at 5 32 25 PM
 
The reason this is happening is because we have two files on the site that load fonts:

- The modern site's CSS: https://github.com/ShelterTechSF/sheltertech.org/blob/3ee96124bdea6df3c9347db429ddd7c63e8b3f3a/src/stylesheets/global.css#L5
- The legacy site's CSS: https://github.com/ShelterTechSF/sheltertech.org/blob/3ee96124bdea6df3c9347db429ddd7c63e8b3f3a/src/legacy/stylesheets/utils/_typography.scss#L9

Both are still being loaded on every page, since we're simultaneously serving legacy and modern pages. However, the legacy site loads the 300 weight version of Lato in both regular and italics (that's what the "i" in "300i" means), while the modern site does not load an italic version at all. Our normal font on the modern site is 400 weight, and the lack of an italic version of this means that the browser ends up falling back to the closest italic font available, which happens to be the 300 weight version from the legacy site.

If it's not too much trouble, could you alter the modern site's CSS file to load in the italics version of the 400 weight font?

src/components/blog/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,23 @@
+.bleedWrapper {
+  margin: 50px auto;
+  max-width: 630px;
+}
+
+.statistic {
+  color: var(--color-sheltertech-blue);
+  font-family: var(--font-family-merriweather);
+  font-size: 72px;
+  line-height: 92px;
```
richardxia: It's usually much [better to set line-height to a unitless value](https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/). Here, it should actually be exactly `1.3`, which when multiplied with the font-size of 72px would give you exactly the 93.6px that you see in the Figma design.

src/components/blog/QuoteBlock/QuoteBlock.tsx
```diff
@@ -7,7 +8,7 @@ type QuoteBlockProps = {
 
 const QuoteBlock = ({ quote, attributee }: QuoteBlockProps) => {
   const quoteContent = attributee ? `${quote} - ${attributee}` : quote;
-  return <p>{quoteContent}</p>;
+  return <p className={s.quote}>{quoteContent}</p>;
```
richardxia: Nit: Use the [`<blockquote>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote) tag for block quotes. Though, in order to be fully spec-compliant, you would also need to split the attributee out of the element into a separate element, since the HTML spec says that the only contents of the blockquote should be the quotation itself and not any other reference or citation information. Here's an article that goes into a lot of detail into it and comes up with an approach with `<figure>` that is both semantically correct and easy to style: https://css-tricks.com/quoting-in-html-quotations-citations-and-blockquotes/#aa-hey-what-about-the-figure-element

src/components/blog/ButtonBlock/ButtonBlock.module.css
```diff
@@ -0,0 +1,14 @@
...
+  display: flex;
+  flex-direction: column;
+  margin: 50px auto;
+  max-width: 630px;
+  padding: 50px 15px;
+}
+
+.header {
+  font: var(--font-subtitle);
+  margin: 0 0 20px;
```
richardxia: I think if we're only really trying to set one direction of margin, then the individual margin properties like `margin-bottom` would be clearer.

But for this, you could actually just set `row-gap: 20px` in the `.container` class, which would also take care of the spacing and usually do the correct thing if a row were missing.

richardxia (commented):
Awesome work on this! It's pretty cool seeing this live on my computer! Overall, everything looks great, and most of my comments are either about factoring out general patterns or low-level details about fixing up small discrepancies in the spacing in various places.

> The mockup doesn't have a mobile version. It has a couple of notes for mobile, but aside from that I was able to mostly leave things alone and have it still look pretty good. I'm assuming design will have a chance to look at this before it goes live, but the only two shots I (knowingly?) called were:
> - adding left/right padding to text/quote/stats blocks at small screen sizes, so they didn't squish up against the edges
> - letting the CTA block be full width at small screen sizes, to match how the mockup said to handle images

I think these look mostly good to me. The only thing that looks a bit off to me is the internal padding of the CTA block, which we may want to set to 40px to avoid having the CTA text getting too close to the edge of the screen:

Screen Shot 2022-02-13 at 6 13 04 PM

@derekfidler, if you have a chance, feel free to clone and check out this branch and manually navigate to the blog page (http://localhost:8000/blog) to see what it looks like on mobile and let us know if there's anything that you think should be changed.

---

Other than that, this all looks great, and I'm excited to have this merge in soon!

src/components/blog/TextBlock/TextBlock.module.css
richardxia: From Derek's response in https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5?node-id=4579:8133#151959166, it looks like this should be 16px, so let's go with that.

src/components/blog/TextBlock/TextBlock.module.css
richardxia: From Derek's response in https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5?node-id=4579:8137#151958348, looks like the margins are a bit complicated here, and there was a bit of confusion over whether the numbers listed in the Figma document included margin-collapse, but he's illustrated in detail to the right of the comment what the exact spacing should be in several scenarios.

src/components/blog/ImageBlock/ImageBlock.module.css
jessica-px: Setting the margin to `auto` wouldn't center the image unless I set `display` to `block`, but since `block` is concerning here I just replaced it with `flex` and `align-items`

src/components/blog/QuoteBlock/QuoteBlock.tsx
jessica-px: I ended up using the `<q>` element instead, because `<blockquote>` isn't inline, so there was no good way for me to get the attributee on the same line.

jessica-px:
@richardxia The biggest changes in this batch of commits are:

1. I replaced margins between blocks with Spacer components, and I added a helper function in the template to wrap the blocks in Spacers dynamically. Potentially another route could involve a `flex` container and `gap`, which would save us from ever having to worry about consecutive Spacers, but by the time it occurred to me I'd already finished doing it this way 🤷‍♀️ But I think one advantage of doing it this way is might be the more explicitly dynamic spacing we get from using Spacers. With `gap` I think we would need to set a baseline amount and add margins to individual elements to account for variances (eg. the `LogoSeparator` would need an extra 30px whitespace on top/bottom to play nice with the default 50px spacing)...

2. I believe I've got the margins worked out for the rich text elements now -- thanks for analyzing the mockups and catching so many little details! Note: Since I wrote this code around the idea that every other block type is wrapped in separators, the text block itself doesn't have any kind of "outer margin". One side effect of this is that if you have two in a row (which there is no point in doing), there wouldn't be any spacing between them. This was true of my last batch of commits as well, but I just wanted to point it since it's a known weak point in this implementation. Again, a container with `gap` would resolve that, but handling the different margin sizes would be weirder...

src/stylesheets/global.css
```diff
@@ -2,7 +2,7 @@
 
 @import '~normalize.css';
 @import url('https://fonts.googleapis.com/css2?family=Merriweather:wght@300&display=swap');
-@import url('https://fonts.googleapis.com/css2?family=Lato:wght@400;700&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Lato:wght@400i,400;700&display=swap');
```
richardxia: This doesn't appear to be working for me, and, disappointingly, it seems to fail silently.

I manually generated a new URL by going to [Google Fonts](https://fonts.google.com/specimen/Lato?query=lato#standard-styles) and manually selecting the variants we want, and I got the following URL, which seems to work correctly when I modify this line locally:

```suggestion
@import url('https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400;0,700;1,400&display=swap');
```

I think I may have incorrectly suggested the wrong manual edit to you previously, since I now see that the `@import` statements from the legacy CSS files use `/css`, but this file uses `/css2`, which I'm assuming is Google's v2 of the API.

After I made the change locally with this URL, the italicized text finally looks like it has the same thickness as the regular text around it.

src/stylesheets/global.css
```diff
@@ -69,6 +69,9 @@
     minmax(40px, 1fr)
     minmax(0, 1290px)
     minmax(40px, 1fr);
+
+  /* Blog Pages */
+  --block-block-width: 630px;
```
richardxia: Did you mean to write `--blog-block-width`?

src/templates/BlogPostTemplate/blogPostTemplate.tsx
```diff
@@ -38,69 +39,98 @@ const BlogPostTemplate = ({
...
+      case "separator":
+        return <LogoSeparator />;
+      default:
+        return null;
+    }
+  };
+
   return (
     <Layout>
+      <Spacer heightDesktop="80px" heightMobile="80px" />
```
richardxia: I know we don't have any mobile designs, but I think this spacing is probably too much for mobile. When doing responsive design, it's common to not using linear scaling for different screen sizes, that is, when going from desktop to mobile, you usually shrink larger font sizes by a greater factor than smaller font sizes. Our site is no exception, and I found the following scaling factors:

- The headline font used for page titles is 50px on desktop and 30px on mobile, which is a 40% shrinkage (or 0.6 ratio)
- The regular paragraph font is 20px on desktop and 16px on mobile, which is a 20% shrinkage (or 0.8 ratio)

My suggestion, then, would be to shrink 80px spacers to 50px spacers on mobile (0.625 ratio), and the 50px spacers to 30px (0.6 ratio).

I think we may also need to change the top and bottom margins on some of the individual elements like h1 tags, p tags, lists, etc., but you might be able to avoid using media queries by defining the margin in `em` units, which scale with the font-size.

I'm happy to defer these and other mobile-only spacing tweaks to a separate PR, since what we have right now looks fine enough to be shippable.

src/templates/BlogPostTemplate/blogPostTemplate.tsx
```diff
@@ -38,69 +39,98 @@ const BlogPostTemplate = ({
...
+  const getBlockSpacerSize = (sliceType: string): string => {
+    switch (sliceType) {
+      case "separator":
+        return "80px";
+      case "text_block":
+        return "0";
+      default:
+        return "50px";
+    }
+  };
```
richardxia: I see where you're going with this, as well as your acknowledgement that this has a few shortcomings, but I think this could all be both simpler and more correct adding that 30px spacer to both ends of the separator which you had mentioned in your comment.

I understand your point about being more explicit with the type-dependent spacer sizes, but I think in the general case we'll need the "set a baseline amount and add margins" approach anyway, since that's closer to how spacing "works" on the page from a design perspective. Part of the issue is hidden by the fact that the example in the Figma design only ever interleaves non-text-block components between text blocks, but this is not a necessary restriction with real blog posts, which may want to do things like put a stats block right after an image block. In your current implementation, I believe these scenarios would have double the expected spacing.

The "spacer sandwich" approach you have here doesn't actually give you enough flexibility to control the spacing in this case, since the correct amount of spacing is a function of both the previous and next slice element. To correctly handle a separator adjacent to any non-text-block component, you'd need to know both types in order to insert the correct spacing. (Note that this is why collapsing margins exist in the first place in CSS, since there was a common need for "set the margin to be the max of this element's bottom margin and the next element's top margin".)

In any case, given the current set of components in the Figma design, the logo separator is the only component that has larger spacing requirements, and we can obviate the issue by padding the separator block with 30px spacers on both ends.

To be more specific, I think you can still keep the spacers all here in this page-level component, and just put the 30px spacers around the separator within the `getBlockComponent()` method. If you do that, then it would allow you to unconditionally put a single 50px spacer directly inside the `slice.map()` in the main body, since you no longer need to select a spacing amount based on the type of the block. With a single spacer per block, I think you'd want to put it above the block being mapped over, and just remove the bottom spacer from the header image. Alternatively, you could switch to `gap` if you'd prefer.

This would have several benefits, including:

- Simplifying this code by making the spacing constant regardless of the block type. One fewer functional call means less indirection in the code, which helps readability.
- Avoiding the double spacers in the generated HTML, which was a bit noisy when I was inspecting the HTML in the developer console.
- Fixing a few of the edge cases, including the case where there are two consecutive text blocks like you had mentioned, but also the case of any consecutive non-text blocks in a row having an extra spacing.

Does this seem reasonable to you, and does this seem reasonable to address before merging this PR?

src/components/blog/StatsBlock/StatsBlock.module.css
```diff
@@ -0,0 +1,23 @@
+.wrapper {
+  margin: 0 auto;
+  max-width: var(--block-block-width);
+}
```
richardxia: Super nit, but I noticed that there's actually supposed to be a 5px margin between the statistic and its accompanying text. Could you add that in here?

richardxia (approved):
Thanks for the last set of updates. I have a few more mostly minor comments, along with a much larger one about spacing. Let me know what you think about it. If you want to get this PR merged without addressing that, just to get all the other changes in, I totally understand and think we can address it in a followup PR.

Thanks again for working on this! Seeing all this live in my local Git checkout is pretty awesome!

src/components/blog/ImageBlock/ImageBlock.module.css
richardxia: Ah, I see. Yeah, I think having the parent be a flex container makes sense. I'm in general a bit wary of directly messing with `img` elements, so this looks like a good alternative to me.

src/components/blog/QuoteBlock/QuoteBlock.tsx
richardxia: Sounds good. You probably could have just set the display style to `inline`, but I think this is fine, too.

src/stylesheets/global.css
jessica-px: Weird, it was definitely working for me 🤔 This works too though!

src/stylesheets/global.css
jessica-px: Hahahaha it's funny how completely invisible that typo was to me, thanks for noticing!

src/templates/BlogPostTemplate/blogPostTemplate.tsx
jessica-px: I went ahead and adjusted the spacer sizes, since that was fast, but I agree that the other things don't need to block this PR

src/templates/BlogPostTemplate/blogPostTemplate.tsx
jessica-px: Refactored!

richardxia:
Thanks for making all the changes!

PR #314, opened 2022-03-01
Implement BarStatCard component and refactor

### LINKED ISSUE #295: Create Bar Stats Block component
https://www.figma.com/file/kKaWvJvyzhNS9hD21v7Ei5/ShelterTech.org-2020?node-id=4329%3A6859

Screen Shot 2021-10-31 at 9 52 59 AM

The Impact Page has a special component with bars that represent some metrics. These bars do not need to be exactly proportional to the numbers, but can be relatively sized and colored.

candywang:
I've refactored the `StatsBlock` component to include a `BarStatCard` component and made  the properties `statBars` and `statCards` optional.


src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
+  barStatCards?: BarStatProps[];
+  statCards?: StatCardProps[];
   theme: ThemeColorOption;
   titleColor?: TitleColorOption;
 };
 
 const StatsBlock = ({
   title,
   subtitle,
+  barStatCards,
```
richardxia: Although it's tempting reuse the existing StatsBlock component and add another prop to it for the bar stats cards, I would actually suggest against doing that. It's a bit counterintuitive, since we kind of trained to minimize unnecessary duplication, but I think there are a few drawbacks to putting both of these in the same component:

- The shape of the data within the two different kinds of stat cards is very different, and since they are the most prominent part of the component, it's almost not worth sharing the outer wrapper bits
- The props are now a bit more complicated, and it's now possible to end up with two "invalid" states: 1) when both `barStatCards` and `statCards` are undefined and 2) when both `barStatCards` and `statCards` are defined. Having them be in separate components makes the props much simpler because you only need one prop and you can make it required.
- You no longer have separate "names" for the two different variants, which may make it slightly more difficult to communicate to others, since the variants aren't explicitly named. "We need to make a change to the Stats Block" "Which Stats Block, the one with the numbers or the one with the bars?"
- None of our other components allow for this much variation in their props.

I don't think it's unreasonable to do what you've done here, but I think this is a case where some duplication of the wrapper code is acceptable, since it keeps the components more crisp and single purpose. Let me know what you think, and if you agree with my reasoning, then let's split the Bar Stats into an entirely separate component.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
+      <div className={s.secondaryNumber}>{legend.subtitle}</div>
+    </div>
+  );
+};
+
 type StatsBlockProps = {
   title: string;
-  subtitle?: string;
-  statCards: StatCardProps[];
+  subtitle?: string | React.ReactNode;
```
richardxia: You may want to revert this if you decide to split out the bar stats, but I just wanted to point out that `ReactNode` already includes `string`, since `ReactNode` is basically meant to represent anything that can appear in a JSX (HTML) region, and `string`s can appear in JSX. This means that you could just replace this with `React.ReactNode` and it'll have the exact same type.

package-lock.json
```diff
@@ -12146,9 +12146,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001239",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001239.tgz",
-      "integrity": "sha512-cyBkXJDMeI4wthy8xJ2FvDU6+0dtcZSJW3voUF8+e9f1bBeuvyZfc3PNbkOETyhbR+dGCPzn9E7MA3iwzusOhQ=="
+      "version": "1.0.30001312",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001312.tgz",
+      "integrity": "sha512-Wiz1Psk2MEK0pX3rUzWaunLTZzqS2JYZFzNKqAiJGiuxIjRPLgV6+VDPOg6lQOUxmDwhTlh198JsTTi8Hzw6aQ=="
```
richardxia: Since you didn't make any changes to the set of installed packages, could you revert the change to package-lock.json? Let's only change this file when we intend to make changes to the set of installed packages.

src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
```diff
@@ -7,11 +7,11 @@ export default {
   component: StatsBlock,
 };
 
-const Template = ({ title, subtitle, statCards, theme, titleColor }) => (
+const Template = ({ title, subtitle, barStatCards, theme, titleColor }) => (
```
richardxia: You'll want to revert all these changes if you split out the bar starts into a separate component, but for future reference, if we ever did want to add a completely different variant to an existing component, we should just define a new "story" instead of changing the existing one. This way, all the stories for a given component can be viewed in Storybook.

Here's an example from the Storybook docs on defining multiple stories: https://storybook.js.org/docs/react/writing-stories/introduction#using-args

src/components/grid-aware/StatsBlock/StatsBlock.stories.jsx
```diff
@@ -20,28 +20,57 @@ const Template = ({ title, subtitle, statCards, theme, titleColor }) => (
 export const DefaultStatsBlock = Template.bind({});
 
 DefaultStatsBlock.args = {
-  title: "See what our volunteers have accomplished together.",
-  subtitle: "Optional subtitle.",
-  statCards: [
+  title: "ShelterConnect",
+  subtitle: (
+    <div>
```
richardxia: A `<div>` is a structural element, so we should avoid creating them unless we are intentionally creating a level structure. I would say it's probably a code smell to have a `<div>` that doesn't contain a class name.

For cases like this, you can just create a React [Fragment](https://reactjs.org/docs/fragments.html) with `<>` and `</>`, which allows you to have a list of items without having to put them under a real parent element. `<>` and `</>` are special shorthand syntax for `<React.Fragment>` and `</React.Fragment>`, in case you're not familiar with it..

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
+  legend: { title: string; subtitle: string };
+  position: string;
+};
+
+const BarStatCard = ({ axes, legend, position }: BarStatProps) => {
+  const isVertical = position === "vertical";
+  const statNumbers = axes.map((axis) => axis.y);
+  const maxNumber = Math.max(...statNumbers);
+  const fillPixels = statNumbers.map(
+    (number) => `${Math.round((number / maxNumber) * 100)}px`
```
richardxia: Sorry, I think I wasn't clear in my ticket, but I don't think these bars are actually meant to be scaled proportionally to the actual numbers. They're only meant to be viewed relatively or figuratively, so that you can see which one is the largest, but not necessarily proportional to the actual values.

We can just have the lengths of the bars be fixed at three static sizes, shortest, middle, and longest. This would simplify the implementation a lot, and it would also make the component more flexible in the cases where the actual "stat" might not be represented as a pure (JavaScript) number, for example, "700m" "800m" "1.2km".

Sorry for asking you to significantly rework this, but I think it'll at least be a lot more straightforward with fixed sizes. You may, however, want to add a prop that allows you to control whether the bars are in increasing or decreasing order, as a complement to the horizontal/vertical orientation prop you already have.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
 
+type AxisProp = {
+  x: string | number;
+  y: number;
+};
+
+type BarStatProps = {
+  axes: AxisProp[];
+  legend: { title: string; subtitle: string };
+  position: string;
```
richardxia: I think "orientation" might be a better name for this. "Position" makes me think of "where this should be placed", such as "top", "bottom", "bottom-left". I think "orientation" is a more common word for whether something is horizontal or vertical.

Could you also set the type to `"horizontal" | "vertical"` instead of `string`? This would give is much tighter type checking, since all other string values would be invalid.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
   </div>
 );
 
+type AxisProp = {
+  x: string | number;
+  y: number;
```
richardxia: x and y might not be the best names here because in 2D graphs, the overwhelming convention is for x to be the horizontal dimension and y to be the vertical dimension. In your code, it looks like y is the "stat" that affects the bar size and "x" is the "label" that describes what the bar represents but not its size.

What if you named this type `Bar`, and gave it "label" and "statistic" properties? I'm not sure if "label" and "statistic" are the best names, so please feel free to come up with other ideas, but I think they are at least better than "x" and "y", especially when "x" and "y" might describe the opposite axes that people might expect from a 2D coordinate system.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
+
+  return (
+    <div className={s.card}>
+      <div className={`${s.barStat} ${s[position]}`}>
+        {axes.map(({ x, y }, i) => {
+          const isYMaxNumber = y === maxNumber;
+          const yAxis = y.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+
+          return (
+            <div key={`${x}-${y}`} className={`${s.barItem} ${s[position]}`}>
```
richardxia: React keys are meant to be some kind of identifier that is "sticky," even if the data might change due to other interactions on the page. It's not a huge deal on our app, since this is a static website, but I think using just `x` (or the `label` under my suggested name) would be a more appropriate key, since it's closer to being an "identifier" for that particular bar.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -132,3 +132,65 @@
     line-height: 1.3;
   }
 }
+
+.barStat {
```
richardxia: Maybe this could be renamed to something like "barsContainer"? I misinterpreted "stat" to mean "statistic", as in a single statistic as opposed to multiple statics, and therefore I thought this was just one bar and its associated label and value rather than an element that contained all of the bars.

src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -132,3 +132,65 @@
...
+
+.barStat.vertical {
+  flex-direction: row;
+  gap: 26px;
+}
+
+.barStat.horizontal {
+  flex-direction: column;
+  gap: 24px;
+}
```
richardxia: This is really clever and slick!

However, I'm seeing an issue with how the bars are aligned in the horizontal variant:
Screen Shot 2022-03-02 at 9 43 17 PM

It's going to be tricky to align the bars with your two-level flexbox approach here, since the widths of the labels is not known statically. You _might_ be able to pull this off with CSS grid, which should allow the width of the widest label determine the width of all other labels in the column, but it'll require restructuring your HTML and CSS a bit. On the plus side, if we stop precisely computing the lengths of the bars and just stick with some fixed lengths, I think it'll simplify this by a lot.

If you try out the grid approach, one piece of advice I have is to treat the bar and the statistic number as a single column/row. This is because the numbers are _not_ aligned with each other, and they should appear right after the bar. In other words, you'll want a 3x2 (or 2x3) grid, not a 3x3 grid, where the label is one row/column and the bar + number combined is the other row/column.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
   </div>
 );
 
+type AxisProp = {
+  x: string | number;
+  y: number;
+};
+
+type BarStatProps = {
+  axes: AxisProp[];
```
richardxia: If you go with my suggestion to rename `AxisProp` to `Bar`, then this prop can also be renamed to `bars`.

src/components/grid-aware/StatsBlock/StatsBlock.tsx
```diff
@@ -19,17 +19,69 @@ const StatCard = ({ number, secondaryNumber, statement }: StatCardProps) => (
...
+  legend: { title: string; subtitle: string };
```
richardxia: I think this is not quite a "legend", since it doesn't really map symbols, colors, or other graphical hints to their meanings in the graph. I can't really think of a good name for the combination of both this title and subtitle, so it might be better to just remove the "legend" part of it and just make "title" and "subtitle" be direct properties of `BarStatProps`.

richardxia (commented):
Good work on this! Although I'm suggesting that we make some significant changes, I think the baseline of what you have here is great, and what is already visible in the Storybook story is very close to what we want!

Let me know if you have any questions about my comments or if you prefer stick with the way you approached some of this stuff. I'm not 100% set on some of my suggestions, since I think some of them could go either way, and I just wanted to offer some ideas.

candywang:
@richardxia you were right about the keys warnings in PR #317. I've fixed the warnings.

src/components/grid-aware/StatsBarBlock/StatsBarBlock.module.css
```diff
@@ -0,0 +1,218 @@
...
+
+.blue {
+  color: var(--color-sheltertech-blue);
+}
+
+.subtitle {
+  font: var(--font-body-medium);
+  grid-column: 2 / span 8;
+  margin-top: 20px;
+}
```
richardxia, line 86: I was looking carefully at the spacing in the developer console, and it looks like `h2` elements have a default `margin-bottom` of about 16px, which you may want to explicitly set to 0 to exactly match the design.

src/components/grid-aware/StatsBarBlock/StatsBarBlock.module.css
```diff
@@ -0,0 +1,218 @@
...
+    padding: 50px 0 38px;
+  }
+}
+
+.gridAreaBottom {
+  display: flex;
+  flex-direction: row;
+  grid-column: 2 / span 10;
+  grid-row: 3;
+  margin-top: 54px;
```
richardxia: I'm seeing a margin of 50px in the Figma design:

Screen Shot 2022-03-16 at 11 15 37 PM

Could you change this to 50px?

```suggestion
  margin-top: 50px;
```

src/components/grid-aware/StatsBarBlock/StatsBarBlock.module.css
```diff
@@ -0,0 +1,218 @@
...
+  font: var(--font-body-medium);
+  margin-top: 5px;
+}
+
+.barsContainer {
+  display: grid;
+}
+
+.barsContainer.vertical {
+  column-gap: 26px;
```
richardxia: Screen Shot 2022-03-16 at 11 17 53 PM

It's subtle, but I think there's actually a column gap of 20px here. The trick is that the year label at the bottom is actually the widest part of the column, not the bar nor the stat number, and the spacing between the labels is 20px.

I also noticed that when I tweaked this number to 20px, it becomes a lot more obvious that the bars are not properly center-aligned to the year label. You'll probably also need to change the alignment (or justification, because I always mix them up) to make sure the bars are centered over the labels.

richardxia (approved):
Nice, this is really great work! I've left some final spacing and CSS comments, but I think they're all relatively straightforward.

I have two other requested changes:

1. Currently `index.ts` is empty. We'll want to export the main component from that file, just like all our other components, or else it'll be a bit more annoying to import the component onto the actual page when we implement it.
2. I noticed that the horizontally-oriented bar stat cards are a bit too high up:
Screen Shot 2022-03-16 at 11 22 28 PM

Notice how the titles and subtitles don't line up. If you look at the Figma design, they should all line up at the bottom. Since the vertically-oriented cards are taller, this means we really want to bottom-align all the cards so that the horizontal cards line up with the vertical ones. Could you change the styles so that the cards are all bottom-aligned? I think you can do this with one of the alignment/justification grid properties on the containing grid element.

Feel free to merge this in once you address these final comments, and thanks again for all the hard work!

src/components/grid-aware/StatsBarBlock/StatsBarBlock.tsx
```diff
@@ -0,0 +1,101 @@
...
+  titleColor = "black",
+}: StatsBarBlockProps) => (
+  <div
+    className={`${s.bleedWrapper} 
+      ${s[theme]}`}
+  >
+    <div
+      className={`${s.bleedBackground} 
+        ${s[theme]}`}
+    />
```
richardxia: (Sorry, forgot to actually attach this to my review)

Since these will all fit on a single line, could you make it so that each of these `<div>`s take up only a single line? I think it'll be more readable that way.

```suggestion
  <div className={`${s.bleedWrapper} ${s[theme]}`}>
    <div className={`${s.bleedBackground} ${s[theme]}`}/>
```

PR #317, opened 2022-03-10
Upgrade caniuse-lite package to latest version

candywang:
I upgraded the outdated **caniuse-lite** package to the latest version which will fix the failed test in PR #314.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12146,9 +12146,9 @@
       }
     },
     "caniuse-lite": {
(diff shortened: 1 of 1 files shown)


richardxia (approved):
This LGTM, but I don't think this is actually causing the CI failure in #314, which seems to be related to a React keys warning. I think we can still merge this in, since it shouldn't be harmful to update the caniuse database.

PR #322, opened 2022-03-20
Merge prismic branch into main

richardxia:
As we had discussed in today's meeting, we feel pretty comfortable merging the current `prismic` branch into the `main` branch, since the only outstanding tasks affect workflows, but not content. It might be nice to take a quick look over these changes, just in case there's anything we want to remove, but otherwise, I'll hit the merge button on this in a few days.


richardxia:
Using special admin powers to merge this with a merge commit, since I do want to preserve the individual merges into this branch.

PR #323, opened 2022-03-22
Add preview for blog posts

### LINKED ISSUE #315: Add Prismic previews for blog posts
Although we added integration with Prismic previews in https://github.com/ShelterTechSF/sheltertech.org/pull/280, there is a bit of extra work that needs to be done for each previewable page.

We did not enable previews with the initial development of the Blog Post page, so this ticket is meant to track that task.

richardxia:
Resolves #315.

@jessica-px had done most of the work for Prismic previews in https://github.com/ShelterTechSF/sheltertech.org/pull/280, and this is just doing the remaining plumbing work to enable them for the Blog Posts.

In addition to adding the usual Prismic Preview boilerplate to the Blog Post page component, this adds two more things:

- It adds a `/preview` route to the app. This corresponds corresponds with the preview route that we configure within the Prismic app so that previewing a document from the Prismic app will eventually take you to the correct page in our site
- It adds the special "unpublished" Prismic boilerplate to the 404 page, which is necessary for previewing unpublished documents. Since the documents haven't been published, there will not have been a page created for them, so you can't preview them on their page. The special Prismic boilerplate on the 404 page will essentially dynamically construct the page from the 404 page itself, which allows you to preview that content even if the page doesn't exist yet.

One more note about that second bullet: The 404 page previewing is a bit weird in development mode, since Gatsby hijacks the normal 404 page. You can still preview unpublished content in development mode, but you have to click this button on the 404 page:
Screen Shot 2022-03-21 at 9 00 09 AM

(The styling is messed up on the Gatsby development 404 page because we have a global CSS reset stylesheet from the legacy site that annihilates the styles for everything, including buttons, which end up looking like regular text.)

Presumably in production this won't be an issue, but I won't be 100% sure until we actually deploy to at least staging.

I wrote a FAQ for this repo to explain the weird previewing behavior in development.


richardxia:
Just double checked that this still works. Going to merge this now and test one more time in production just to double check it works on the production version of the site through the preview link in Prismic.

richardxia:
How frustrating. It wasn't working in production until I added a trailing slash to the `/preview/` URL, which we apparently need because we're hosting the site on AWS S3. Thankfully, enough people have it this exact issue that Prismic even put this in [their official docs](https://prismic.io/docs/technologies/previews-gatsby#1.-enable-previews-in-your-repository).

But it's now working in prod, so everything's good.

Issue #324, opened 2022-03-28
Update StatsBlock component


candywang:
The large stat number in the StatsBlock component has a font of merriweather, 72px, and 300. We can create a `--font-stat` typography in `global.css` or change the font styling in `StatsBlock.module.css`. @richardxia , which is preferred?

In StatCard, the `secondaryNumber` prop type should be a string or a number.

richardxia:
Nice catch! I think we had hit something similar when we first built out this component for the Volunteer page, where at the time the Figma design had a totally different font than anything else on the site. It looks like after we created the new Impact page designs, we also updated the Volunteer page as well to use Merriweather at 300 weight, so you're right that we should update the styles to match.

Since this style only appears in one component, I think it's fine to continue to define the style within the component's CSS, but I wouldn't be opposed to defining it in global.css if you wanted to.

PR #325, opened 2022-03-28
Implement Impact page

candywang:
I've implemented the Impact page using the current Figma design/content, minus the subtitle link and statCards - secondaryNumber. I will update the content once we get approval for the final design/content. I've also opened up a new issue #324 to fix the font for the large stat numbers and the statCards - secondaryNumber prop type in the StatBlock component.


src/pages/impact/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+        },
+        {
+          number: "13.4k",
+          secondaryNumber: 6,
+          statement: "Annual users",
+        },
+        {
+          number: "111",
+          statement: "New resources",
+        },
```
richardxia: I just noticed that the vertical alignment is a bit off on this card, since it doesn't have the secondary number. In the Figma design, it looks like everything should be bottom-aligned, similar to the stats bar block. Do you think you could make that change to the component? It could even be in the same PR as when you update the font styles for https://github.com/ShelterTechSF/sheltertech.org/issues/324.

src/pages/impact/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+          number: "$79k",
+          secondaryNumber: 6,
+          statement: "Annual funding",
+        },
+        {
+          number: "$2m",
+          secondaryNumber: 6,
+          statement: "in in-kind contributions",
+        },
+      ]}
```
richardxia: It looks like the horizontal alignment of these cards is a bit off when there are only two cards. In the Figma design, it looks like the cards should be in the exact same places as when there are three cards, so we probably need to get rid of some of styling that causes the cards to expand. Could you update that as well?

richardxia (commented):
Nice job putting this together! This mostly looks good for now, until we get the final content. I did notice a few minor styling issues with the original stats block, so I think it'd be worth spending the time to fix them, especially since we're blocked on content at the moment.

src/components/grid-aware/TextHeader/TextHeader.module.css
```diff
@@ -13,17 +12,25 @@
...
+
+  .bleedWrapper.hasBottomPadding {
+    padding-bottom: 40px;
+  }
 }
 
 .bleedMainContent {
   grid-column: 2;
   grid-row: 1;
+  width: 100%;
```
richardxia: I took a look at this offline, and I figured out why we had to add this to get things to line up properly. The reason is that the `.bleedWrapper` sets `justify-items: center` for some reason, so then if the bleed content doesn't fill up 100% of the space, it ends up getting centered. I don't think we need `justify-items: center` for the `.bleedWrapper`, so let's get rid of it and also get rid of this `width: 100%`.

src/pages/impact/index.tsx
```diff
@@ -0,0 +1,187 @@
...
+    <StatsBlock
+      title="Hosted Partners"
+      subtitle="ShelterTech is proud to provide the data on the SF Service Guide platform for other organizations to use on their own websites. We have worked closely with SF Families and Link-SF to provide a custom experience catered to their users’ needs, leveraging our data curation pipeline to keep that information accurate and up to date."
+      statCards={[
+        {
+          number: <img src={linkSFLogo} alt="Link-SF Logo." />,
+        },
+        {
+          number: <img src={sfFamiliesLogo} alt="SF Families Logo." />,
+        },
```
richardxia: Nit: Could you swap the two photos? The paragraph above mentions them in the opposite order, and it's a little bit distracting to see the logos in a different order.

richardxia (approved):
This looks great! I figured out the alignment issue with the text header, so let's fix that before merging this in, but otherwise, this LGTM!

I just wanted to jot down a few notes for some of the hacks we added here so that we have a record of them:

- We added the `hasBottomPadding` prop and CSS class to the TextHeader, since the padding on this component and the Stat Block components were adding too much space. We need to revisit spacing in general in the app, since right now things are sometimes too rigid and make it difficult to control spacing at the page level
- We're currently abusing the Stat Block component by putting the image logos in there, even though they are meant to hold stat numbers. We'll eventually want to put real stats here, but this was the most expedient thing to do, and it actually looks pretty good when rendered.

richardxia (approved):
LGTM!

PR #329, opened 2022-04-20
Update stats block component

### LINKED ISSUE #324: Update StatsBlock component

candywang:
These changes fix and close #324.

### DIFF
--- a/src/components/grid-aware/StatsBlock/StatsBlock.module.css
+++ b/src/components/grid-aware/StatsBlock/StatsBlock.module.css
@@ -105,7 +105,7 @@
 
 .number {
   color: var(--color-sheltertech-blue);
-  font: 300 72px/1.3 merriweather;
+  font: 300 72px/1.3 var(--font-family-merriweather);
 }
 
 .secondaryNumber {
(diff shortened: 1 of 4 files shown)


src/components/grid-aware/StatsBlock/StatsBlock.module.css
```diff
@@ -105,10 +105,7 @@
 
 .number {
   color: var(--color-sheltertech-blue);
-  font-family: var(--font-family-lato);
-  font-size: 72px;
-  font-weight: bold;
-  line-height: 1.3;
+  font: 300 72px/1.3 merriweather;
```
richardxia: Could you use the variable that we have for Merriweather? This way, we have the fallback fonts set in case of this font doesn't load for some reason.

```suggestion
  font: 300 72px/1.3 var(--font-family-merriweather);
```

richardxia (approved):
LGTM, thanks for pulling out these smaller changes into their own PR!

PR #339, opened 2022-11-08
Delete legacy files.

### LINKED ISSUE #337: Remove legacy pages
When we first launched the redesigned website, we left several of the old pages around, since the new site didn't have equivalent versions of those pages. Although we still have not created equivalents for some of those old pages, they are so out of date now that it would be better to remove them for now.

The full set of pages we should remove are:

- /annual-report-2018/
- /events/
- /get-involved/
- /givingtuesday/
- /team/

In addition to removing the source files associated with these pages in `src/pages/`, we should remove anything in `src/legacy/` that is no longer necessary. We might need to be a bit careful about removing some of the legacy CSS code, since some of them may introduce global CSS rules that affect pages on the new site, so if necessary, we can keep them around and deal with them in a separate issue.

ivanleeswe:
#337 

Removed files related to annual report 2018, events, get involved, giving tuesday, and team.


richardxia (approved):
Sorry for taking such a long time to review this, but this LGTM!

I think there's probably more files we can delete from the `legacy/` directory, but let's get these changes in first.

PR #364, opened 2023-06-16
Change SF Families logo to their new Our 415 logo.

richardxia:
SF Families rebranded as Our 415, and this is their updated logo.

This also makes a few small tweaks to the PartnersAndSponsorsBlock so that image heights are capped and properly centered. We missed that styling previously because all of our previous images were the exact same dimensions (they had already been resized in Figma), but our component should handle images of any size without disturbing the overall layout.


richardxia:
Going to merge this in so we can get the site deployed with the updated logos, but feel free to leave any post-merge feedback if you'd like.

Issue #365, opened 2023-08-10
Upgrade packages


richardxia:
Most of our package dependencies are pretty out of date, with most probably being roughly 2–3 years old. This actually makes it impossible to build the site on an Apple Silicon MacBook, since the versions of some of the packages we depend on don't support darwin/arm64.

For this ticket, I am proposing that we do the following major dependency upgrades, updating other dependencies as needed:

- Node.js 14 -> 18
- Gatsby 3 -> 5
- React 16 -> 18

Once these are all upgraded, it may be easier to bump most remaining packages to their latest versions.

This upgrade looks like it'll be tricky, since Gatsby's [v3 -> v4](https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v3-to-v4/) and [v4 -> v5](https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v4-to-v5/) migration guides contain a nontrivial number of things, and most importantly, Gatsby v5 requires Node.js 18.

Node.js upgrades are also a little bit tricky because it will also force us to bump certain packages that depend on native modules, since they require native code (e.g. C/C++) to be compiled against specific Node.js ABI versions, and package maintainers usually only upload precompiled binaries for specific Node.js ABI versions (the ones that are still supported!).

I think it'll be less tricky to do this in stages, so I'm going to propose that we first target Gatsby v4 and Node.js 16, and then we can tackle Gatsby 5 and Node.js 18.

PR #366, opened 2023-08-25
Upgrade to Gatsby 4 and Node.js 16

richardxia:
Refs #365, though doesn't get us all the way just yet.

This upgrade is partially motivated by the fact that you can't get the local development environment running on an Apple Silicon MacBook, due to dependencies that are too old and don't support the Arm CPUs. As described in #365, it's a bit too much to try to upgrade everything at once, so this is an incremental step that bumps Gatsby from v3 to v4 and Node.js from v14 to v16. Doing so required bumping a significant other number of dependencies, which is why I wanted to do this as an incremental step.

I have more details about specific workarounds in my individual commit messages, so please look at them if you're curious.

One thing that I will point out is that we are now getting a lot of warning messages. It doesn't seem to impact the behavior of the site, but most of them are things that we may want to fix before further upgrading packages, since a lot of them are deprecation warnings. I don't want to do them right now, though, since it was already a lot of work to get to this point.

To test this locally, please make sure you switch to Node 16 and then rerun the normal site setup (see the [README](https://github.com/ShelterTechSF/sheltertech.org/blob/61048d9753052a90cfada3adf8b654928106dd03/docs/getting-started.md)). If you use nvm (the Node version manager), you can automatically switch to the right version of Node.js by running `nvm use` in the repo, since I updated the `.nvmrc` file in this PR.


schroerbrian:
I got stuck on installing the `node-sass` package. I tried a number of things, including installing Python via Home Brew and installing the latest node-gyp globally. When I remove the `node-sass` from the package.json, I'm able to run `npm i` successfully. 

I'm running Monterey version 12.5.1. My machine is using the Apple M1 Pro chip.

richardxia:
> I got stuck on installing the `node-sass` package. I tried a number of things, including installing Python via Home Brew and installing the latest node-gyp globally. When I remove the `node-sass` from the package.json, I'm able to run `npm i` successfully.
> 
> I'm running Monterey version 12.5.1. My machine is using the Apple M1 Pro chip.

Thanks for trying this out! Which versions of `node` and `npm` do you have installed? I've got an M2 MBP running macOS Ventura 13.5, and this branch seems to work for me.

You might need to wipe your `node_modules` directory, since there may be leftover stuff from older package versions or Node.js versions.

In any case, I might just merge this PR in as is without removing `node-sass` as an explicit dependency, since it's no longer supported and should be replaced with the new `sass` package anyway. I think upgrading more of our packages will hopefully alleviate whichever issue you're hitting now, and it might not be worth fine tuning it until we get mostly caught up with the latest versions.

schroerbrian:
> > I got stuck on installing the `node-sass` package. I tried a number of things, including installing Python via Home Brew and installing the latest node-gyp globally. When I remove the `node-sass` from the package.json, I'm able to run `npm i` successfully.
> > I'm running Monterey version 12.5.1. My machine is using the Apple M1 Pro chip.
> 
> Thanks for trying this out! Which versions of `node` and `npm` do you have installed? I've got an M2 MBP running macOS Ventura 13.5, and this branch seems to work for me.
> 
> You might need to wipe your `node_modules` directory, since there may be leftover stuff from older package versions or Node.js versions.
> 
> In any case, I might just merge this PR in as is without removing `node-sass` as an explicit dependency, since it's no longer supported and should be replaced with the new `sass` package anyway. I think upgrading more of our packages will hopefully alleviate whichever issue you're hitting now, and it might not be worth fine tuning it until we get mostly caught up with the latest versions.

Go for it! Let me know if you want me to check it out again. I'm running the following NPM/Node versions:
```
➜  sheltertech.org git:(365-upgrade-to-gatsby-4-nodejs-16) ✗ npm -v
8.19.4
➜  sheltertech.org git:(365-upgrade-to-gatsby-4-nodejs-16) ✗ node -v
v16.20.2
```

richardxia:
Weird, that exactly matches what it should be according to this PR. In any case, let me merge this in for now and work on further package upgrades.

PR #367, opened 2023-09-13
Fix Storybook build after Gatsby v4 upgrade.

richardxia:
The Storybook builds were failing for several reasons after performing the [Gatsby v4 upgrade](https://github.com/ShelterTechSF/sheltertech.org/pull/366). Several were due to package hoisting issues, where Storybook wanted to import a package at the root level, but the Gatsby upgrade caused those packages not to exist at the root level. These were fixed by declaring an explicit dependency on webpack and postcss-loader, which forces them to be hoisted. These can hopefully be removed once Gatsby and Storybook's dependencies on webpack and postcss-loader share common versions, which would allow npm to naturally hoist the packages.

There was one additional change required to the Storybook webpack configuration, which was to exclude a conditional import in the Gatsby codebase that is only meant for React 18. Since we're still on React 16, the import will fail if Storybook's webpack attempts to execute it, but we can configure it to ignore that package altogether. The official Gatsby docs on the Storybook integration made this change in https://github.com/gatsbyjs/gatsby/pull/37222, so we also pull in those changes into our configuration.

Finally, this adds the Storybook build to CI so that we catch these kinds of issues before merging. This was only caught after merging because we have a post-merge GitHub action for building and deploying the Storybook site, but we don't actually run it beforehand.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,3 +32,6 @@ jobs:
 
       - name: Run tests
         run: npm run test
+
+      - name: Check build-storybook
+        run: npm run build-storybook
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,5 +1,6 @@
 const path = require("path");
 const postcssCustomMedia = require("postcss-custom-media");
+const React = require("react");
 
 module.exports = {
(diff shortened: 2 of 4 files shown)


schroerbrian (approved):
LGTM!

PR #368, opened 2023-09-13
Upgrade to storybook 7.4.1.

richardxia:
Refs #365.

This was mostly turnkey, since Storybook provides a command for automatically upgrading and even updating configuration files.

One manual change that was necessary was to update the Storybook webpack configuration, which was described in terms of patching the existing webpack configuration. The patched rule changed from array index 0 to 2, so we had to update the indexes in our webpack configuration. This matches the changes described in
https://github.com/storybookjs/storybook/issues/22065#issuecomment-1646273441

The biggest benefit of making this change is that it removes webpack 4 from our dependencies list. Although we had configured Storybook 6 to use webpack 5, some of Storybook's dependencies still pulled in webpack 4. Now that we've upgraded to Storybook 7, every framework we use now uses webpack 5.


schroerbrian (approved):
👍

PR #369, opened 2023-09-18
Replace gatsby-plugin-typegen with built-in Gatsby typegen.

richardxia:
Refs #365.

Gatsby 4.15 has a built-in feature for generating TypeScript type stubs for GraphQL queries within an app. gatsby-plugin-typegen is no longer maintained, and it's currently blocking some upgrades of other packages because it declares a conflicting NPM peer dependency on graphql.

This commit removes gatsby-plugin-typegen, and it enables the new built-in type stub generation as described in
https://www.gatsbyjs.com/docs/how-to/local-development/graphql-typegen/. Since the new types are not 100% compatible with the previous ones, this commit also includes changes for migrating to the new types.

One issue with the new types is that they are very precise, which actually causes some code to be unnecessarily verbose. In particular, in all instances of inline fragments (the `... on TypeName`) expressions, the generated types infer a pessimistic fallback type of `{}` for the case where an object doesn't match any of the explicit type cases. This makes it impossible to option chaining syntax (`?.`), since TypeScript errs on the side of reporting an error when not all objects in a union declare a property.

For a concrete example, suppose we have the GraphQL query:

```graphql
{
  foo {
    ... on Bar {
      bar
    }
  }
}
```

and suppose we have the inferred TypeScript type:

```typescript
{
  foo: { bar: string } | {} | null
}
```

The following expression contains a type error:

```typescript
foo?.bar
```

This is because `foo` could be `{}`, and the `{}` type in TypeScript could be any object with any properties, _not_ the empty object with no properties.

We can work around this issue by performing an explicit key check:

```typescript
"bar" in foo ? foo.bar : undefined
```

This gets pretty verbose when we have heavily nested objects, but it's not clear if we have any other option for working around this, barring turning off automatic type generation for GraphQL queries. For now, we employ the workaround, but if this proves to be too onerous, we should just disable the GraphQL type stub generation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ jspm_packages/
 # gatsby files
 .cache/
 public
-src/__generated__/
+src/gatsby-types.d.ts
 
 # Storybook files
 /storybook-static
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -67,6 +67,12 @@ module.exports = {
         repositoryName: "sheltertech",
       },
     },
-    "gatsby-plugin-typegen",
   ],
+  graphqlTypegen: {
+    // Needed for CI so that the type stubs are available for type checking.
+    // They are otherwise only generated when running a development server, but
+    // it's difficult to start and stop the development server in a CI
+    // environment (hard to know when to send a Ctrl-C to the process in CI).
+    generateOnBuild: true,
+  },
 };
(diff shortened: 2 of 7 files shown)


schroerbrian (approved):
LGTM! Thanks for the detailed explanation on Gatsby's type checking. I think that could prove to be quite onerous indeed if other files will need to also have that verbosity.

richardxia:
Yeah, I hope later versions of Gatsby and the Gatsby-Prismic plugin we're using have improved and simpler types, since I think some of the types are overly pessimistic. I think it might also be the case that a newer version of TypeScript has better inference for some of the specific cases we're hitting. If not, then we can always disable the plugin.

PR #372, opened 2023-10-04
Upgrade to Node.js 18 and React 18

richardxia:
Refs #365.

This is an intermediate PR to bump us from Node.js 16 to Node.js 18 and React 16 to React 18. This should be the last intermediate step before being able to bump Gatsby to 5, which is the end goal of #365.


.github/workflows/deploy-staging.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/setup-node@v2-beta
+      - uses: actions/setup-node@v3
         with:
-          node-version: '16'
+          node-version-file: .nvmrc
```
schroerbrian, line 16: this is very useful!

schroerbrian (approved):
LGTM!

PR #378, opened 2023-10-20
Add UCSF Acute Care Innovation Center Logo.

richardxia:
This logo is being added to both the Partners and Sponsors block as well as to the Impact Page, under the Hosted Partners section.

Since we wanted to add a hyperlink to their website, this also includes some code changes to those components to allow for optional hyperlinks.

Partners and Sponsors block on the home page:

Screen Shot 2023-10-19 at 6 45 11 PM

Impact Page:

Screen Shot 2023-10-19 at 6 45 01 PM

BTW, @schroerbrian, since the raw image of their logo that we got was huge (like over 5000 px wide), I ended up resizing them to something a lot smaller, to avoid forcing visitors to download huge images that are ultimately rendered at much smaller sizes. This is actually something that [Gatsby's supposed to handle out of the box](https://www.gatsbyjs.com/plugins/gatsby-plugin-image), but we're kind of using Gatsby wrong, so it isn't easy without doing some larger refactoring.

I added two different resized versions, since the Partners and Sponsors block has much smaller images than the Impact page, since I targeted approximately a 2x resolution of the rendered image (for Retina displays). Feel free to grab them for the SF Service Guide site if you want, src/pages/impact/ucsf-acute-care-innovation-center-logo.png is the larger one that's probably better for the SF Service Guide pages.


src/stylesheets/global.css
```diff
@@ -111,6 +111,7 @@ iframe,
 img,
 svg,
 video {
+  max-width: 100%;
```
richardxia, line 114: Oh, I forgot that I should probably briefly explain this. In https://github.com/ShelterTechSF/sheltertech.org/pull/370, I got rid of the legacy style sheets, but had copy over a few of the reset styles manually over to our new site's stylesheet. This one was actually originally in the old reset styles, but I had missed copying it over because it didn't look like any of the images at the time needed it. Since I'm trying to add images at 2x resolution so that they look nicer on high density displays, like Apple Retina displays, we really need this so that images don't grow wider than their parent element.

schroerbrian (approved):
LGTM, thanks for resizing those images!

src/components/grid-aware/PartnersAndSponsorsBlock/PartnersAndSponsorsBlock.tsx
```diff
@@ -2,11 +2,33 @@ import * as React from "react";
 
 import * as s from "./PartnersAndSponsorsBlock.module.css";
 
-type LogoPropType = Readonly<{
+export type LogoPropType = Readonly<{
+  /// The URL of the image
   url: string;
+  /// Alternative text for the image
   alt: string;
+  /// An link for a URL to go to when clicking on the logo
```
richardxia: I just realized that I made a typo here, so I'm going to fix this up right before merging. Thanks for reviewing!

```suggestion
  /// A link for a URL to go to when clicking on the logo
```

PR #381, opened 2024-01-10
Update README.md to fix a small typo

zen-gineer:
Changed "To get the code running for the first _team_," 
To "To get the code running for the first **time**,"


richardxia:
Sorry for the belated review, but thanks for fixing this!

PR #384, opened 2024-02-22
Use new Candid (formerly GuideStar) seal.

richardxia:
This is also a direct link to an image on their website, so it should automatically update when our status changes in the future, which will avoid having to make a code change to update this.

Also, this bumps up the image size slightly, from 60px to 80px. The text was difficult to read at 60px, even with the old GuideStar seal, so we want to increase it a bit. We still want it to be smaller than our own logo in the footer, so 80px is a reasonable middle ground.

**Before**
Screenshot 2024-02-21 at 8 26 04 PM

**After**

Screenshot 2024-02-21 at 8 25 32 PM

### DIFF
--- a/src/components/grid-aware/Footer/NavigationBlock/NavigationBlock.module.css
+++ b/src/components/grid-aware/Footer/NavigationBlock/NavigationBlock.module.css
@@ -61,8 +61,8 @@
 }
 
 .sealImage {
-  height: 60px;
-  width: 60px;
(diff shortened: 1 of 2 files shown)


richardxia:
I think this is a relatively minor change, and since we haven't had too much activity in this repo, I'm going to merge this and deploy it without waiting for a review. But I definitely welcome any post-review comments you have, if you have any.

